PR #25, opened 2025-07-18
Add "getting started" to the top-level readme

rcoh:
📬 *Issue #, if available:*

✍️ *Description of changes:* Update the top level readme so that there is a glanceable summary of how it works


jlizen (approved):
Good incremental improvement.

Only concern is, do we want to increasingly refer to destinations as 'queues' rather than the new 'sink' verbiage? We had intentionally renamed out of respect for `FlushImmediately`, but now our crate name alludes to queues anyway...?

I could go either way, just raising the question.

PR #35, opened 2025-07-22
Attempt to make close by value ref

arielb1:
📬 *Issue #, if available:*

✍️ *Description of changes:*

### DIFF
--- a/metrique-macro/src/lib.rs
+++ b/metrique-macro/src/lib.rs
@@ -121,9 +121,9 @@ pub fn metrics(attr: TokenStream, input: proc_macro::TokenStream) -> proc_macro:
 }
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


metrique/examples/timers.rs
```diff
@@ -49,7 +49,7 @@ type DefaultTimestampFormat = EpochMicros;
 
 // By implementing the `ValueFormatter` trait, you can also define your own entirely custom formats.
 
-#[metrics(prefix = "subevent_", subfield)]
```
rcoh, line 52: this has to be owned because `TimestampOnClose` mutates?

rcoh: we don't need to do it here, but we should add a UI test for `subfield_owned` when you need `subfield` (and ideally have a nice error)

PR #38, opened 2025-07-27
Add 'flex' to allow custom keys

rcoh:
📬 *Issue #, if available:*

✍️ *Description of changes:* This adds `Flex` which allows producing custom keys from metric values more easily


jlizen (approved):
This is a great feature! Thanks for cutting. 

As a further step, would it be worth opening an issue to have support for a type of flex field that is structured as a Vec that desugars to distinct metrics/properties on flush? That gets people everything but the ability to conveniently modify already-written fields in place, but it is a lot cheaper than a HashMap / covers a lot of the use cases that I've seen.

PR #46, opened 2025-07-30
doc: glossary

arielb1:
📬 *Issue #, if available:*

✍️ *Description of changes:*


README.md
```diff
@@ -74,6 +74,59 @@ fn initialize_metrics(service_log_dir: PathBuf) -> AttachHandle {
 
 You can either attach it to a global destination or thread the queue to the location you construct your metrics object directly. Currently, only formatters for [Amazon EMF] are provided, but more may be added in the future.
 
+## Glossary
```
rcoh, line 77: I was imagining this was a slightly higher level listing of terms, specifically ordered by the ones that are new to people:

- `Entry`: "an entry is the unit of work that is flushed at the end of a request or other event. An entry is typically defined by something using the `#[metric]` attribute
- `Sink`: A sink is where entries are flushed to. Builders typically configure a global sink with ...
- Guard: The standard usage pattern with an entry is to drop it use `append_on_drop` (link to example) that flushes it to a `Sink`
- (now include the fields you have below like dimension, datapoint etc.)
- `slot`: ...

metrique/README.md
```diff
@@ -4,9 +4,15 @@ metrique is a crate to emit unit-of-work metrics
...
 ## Getting Started (Applications)
 
-Most metrics your application records will be "unit of work" metrics. These are typically tied to the request/response scope. You declare a struct the represents the metrics you plan to capture over the course of the request and annotate it with `#[metrics]`
+Most metrics your application records will be "unit of work" metrics. These are typically tied to the request/response scope. You declare a struct the represents the metrics you plan to capture over the course of the request and annotate it with `#[metrics]`.
 
-This library exposes a wrapper `<MetricName>Guard` type (a type alias to [`AppendAndCloseOnDrop`]) that implicitly appends to a global queue when the struct is dropped. This wrapper in turn exposes other APIs to further tune behavior.
+The `#[metrics]` macro will make your struct implement the [`CloseEntry`] trait, which represents an entry that
+can be "closed" into an [`InflectableEntry`].
+
+The macro will also expose a wrapper `<MetricName>Guard` type (a type alias to [`AppendAndCloseOnDrop`]) that implicitly appends to a global queue when the struct is dropped. This wrapper in turn exposes other APIs to further tune behavior.
```
rcoh, line 12: this seems too low level for the getting started? I suspect many people will use library and never know what an `InflectableEntry` is. The key thing to capture in these docs is that instead of having some sort of global counters like you might expect from a metric library, you emit individual entries.

metrique/README.md
```diff
@@ -4,9 +4,15 @@ metrique is a crate to emit unit-of-work metrics
...
+
+The best way to emit the [`InflectableEntry`] is by defining a global entry sink using the [`metrique_writer::sink::global_entry_sink`] macro, which defines a global variable that provides a global rendezvous point where metrics can be written into. Then, you need to attach to that global entry sink an actual entry sink that will actually write your metrics somewhere - the example below will write the metrics to an [`tracing_appender::rolling::RollingFileAppender]
```
rcoh, line 14: the key here is that there is the `global` destination which must be separately wired to a destination. I think if we said something like:

```
Callers can write to the global queue with `::queue()` and wire the queue to a destination with `::attach_to_stream()`.

The queue must be wired to a destination before metrics can be sent to the queue
```

metrique/README.md
```diff
@@ -2,11 +2,26 @@ metrique is a crate to emit unit-of-work metrics
...
+The log entries being structured means that you can easily use problem-specific aggregations to track down the cause of issues, rather than only observing the symptoms.
+
+[Amazon CloudWatch]: https://docs.aws.amazon.com/AmazonCloudWatch
+
 ## Getting Started (Applications)
 
-Most metrics your application records will be "unit of work" metrics. These are typically tied to the request/response scope. You declare a struct the represents the metrics you plan to capture over the course of the request and annotate it with `#[metrics]`
+Most metrics your application records will be "unit of work" metrics. In a classic HTTP server, these are typically tied to the request/response scope.
+
+You declare a struct the represents the metrics you plan to capture over the course of the request and annotate it with `#[metrics]`. That will make them implement the [`CloseEntry`] trait, which represents an entry that can be "closed" into an [`InflectableEntry`]. The macro will also expose a wrapper `<MetricName>Guard` type (a type alias to [`AppendAndCloseOnDrop`]) that implicitly appends to a sink when the struct is dropped.
```
rcoh, line 15:
```suggestion
You declare a struct that represents the metrics you plan to capture over the course of the request and annotate it with `#[metrics]`. That makes it possible to write it to a `Sink`. Rather than writing to the sink directly, you typically use `append_on_drop(sink)` to obtain a guard that will automatically write to the sink when dropped.
```

metrique/README.md
```diff
@@ -2,11 +2,26 @@ metrique is a crate to emit unit-of-work metrics
...
+
+The best way to emit the entry is by emitting it to a global entry sink, defined by using the [`metrique_writer::sink::global_entry_sink`] macro. That will create a global rendezvous point - you can attach a destination by using [`attach`] or [`attach_to_stream`], and then write to it by using the [`sink`] method (you must attach a destination before calling [`sink`], otherwise you will encounter a panic!).
```
rcoh, line 17:
```suggestion
The simplest way to emit the entry is by emitting it to a global entry sink, defined by using the [`metrique_writer::sink::global_entry_sink`] macro. That will create a global rendezvous point - you can attach a destination by using [`attach`] or [`attach_to_stream`], and then write to it by using the [`sink`] method (you must attach a destination before calling [`sink`], otherwise you will encounter a panic!).

Typically applications have a single sink. One sink can handle different `#[metric]` structs.
```

metrique/README.md
```diff
@@ -53,6 +70,8 @@ async fn count_ducks() {
 
 
 fn initialize_metrics(service_log_dir: PathBuf) -> AttachHandle {
+    // attach a BackgroundQueue EntrySink to our global entry sink,
```
rcoh, line 73:
```suggestion
    // attach a EMF-formatted rolling file appender to `ServiceMetrics`
```

metrique/README.md
```diff
@@ -66,6 +85,8 @@ fn initialize_metrics(service_log_dir: PathBuf) -> AttachHandle {
 
 #[tokio::main]
 async fn main() {
+    // not strictly needed, but metrique will emit tracing errors
+    // in some condition, and it's best to be able to see them.
```
rcoh, line 89:
```suggestion
    // when entries are invalid and it's best to be able to see them.
```

Issue #42, opened 2025-07-28
Feature Request: First-class support for applying dimensions to all fields in a metrics struct


slack2450:
Currently, when emitting metrics with dimensions, each individual metric field must be wrapped with WithDimension<T>. For structs with many metrics, this becomes cumbersome and makes the code messy, especially for common dimensions like service context.

```rs
// Current approach - each field needs WithDimension wrapper
#[metrics(rename_all = "PascalCase")]
struct RequestMetrics {
    request_count: WithDimension<Counter>,
    processing_time: WithDimension<Timer>, 
    error_count: WithDimension<Counter>,
    // ...
}
```

Please could I request first-class support for applying dimensions to all fields in a metrics struct. I'm unsure how I'd like the API to look potentially like this:

```rs
let metrics = RequestMetrics { /* ... */ }
    .with_dimensions([
        ("Service", "UserService"),
        ("Environment", "production")
    ]);
```

This would be perfect for when we have a group of metrics that relate to the same operation, e.g Timings and their correlated request id.

I discussed this offline with @rcoh who thought this would be a good idea and provided [this](https://github.com/awslabs/metrique/compare/dimensions-all-fields-example?expand=1) workaround in the meantime.

rcoh:
This will go out in the next release

PR #47, opened 2025-07-30
feat: allow WithDimensions and ForceFlag support for entries

arielb1:
📬 *Issue #, if available:*

Helps with #42 

✍️ *Description of changes:*

This allows wrapping entries with `ForceFlag` (and wrappers, e.g. `NoMetric` etc.) or `WithGlobalDimensions`, as in the test:
```rust
    #[metrics(flatten)]
    sub: HighStorageResolution<WithGlobalDimensions<MySubfield, 1>>,
```

previously only values could be wrapped

### DIFF
--- a/metrique-core/src/close_value_impls.rs
+++ b/metrique-core/src/close_value_impls.rs
@@ -4,14 +4,16 @@
 //! All default implementations of CloseValue, grouped for clarity
 
 use core::time::Duration;
+use std::marker::PhantomData;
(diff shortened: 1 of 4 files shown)


rcoh (commented):
discussed offline: instead we will make `WithDimensions` work for entries & remove the deny list functionality. If you want to have a field _not_ have a dimension applied, you can put it in a different struct under the root

rcoh (approved):

PR #85, opened 2025-08-14
Null entry sink

### LINKED ISSUE #77: Add docs / clean support for a "DropEverythingSink" to be used in tests.
Since:
1. calling `ServiceMetrics::sink` will panic if a sink isn't attached
2. you can't attach a sink multiple times

Properly setting up a dev-null sink globally across your tests is mildly annoying. We should capture this pattern as something in `test-util` so that callers can do something like:

```rust
#[test]
fn my_test_that_uses_metrics() {
  // automatically deals with having it be a global lazy static, not attaching multiple, etc.
  let _guard = test_util::dev_null_metrics(MySink);
}
```

arielb1:
📬 *Issue #, if available:*

Fixes #77 

✍️ *Description of changes:*


metrique-writer/src/sink/mod.rs
```diff
@@ -116,28 +117,58 @@ impl<E> VecEntrySink<E> {
     }
 }
 
+/// An [EntrySink] that drops all entries.
+///
+/// Useful for testing, or when you want to ignore entries.
+#[derive(Copy, Clone, Default)]
+pub struct NullEntrySink {
```
rcoh, line 124: `DevNullSink`?

metrique-writer/src/sink/mod.rs
```diff
@@ -116,28 +117,58 @@ impl<E> VecEntrySink<E> {
...
+    // have a private field
+    _marker: PhantomData<()>,
```
rcoh, line 126: or `#[non_exhaustive]` to avoid all the boilerplate of having a field

rcoh (commented):
lgtm. `NullEntry` doesn't totally capture it, might be hard to search for too. Doesn't really matter though.

rcoh (approved):

PR #89, opened 2025-08-24
fix: fix imports via metrique to work without a metrique-writer in Ca…

arielb1:
…rgo.toml

also fix metrique/test-util to depend on metrique-metricsrs/test-util

📬 *Issue #, if available:*

✍️ *Description of changes:*

Add a version of the `Entry` macro that uses imports from `metrique`, to allow you to write code like this:

```toml
[dependencies]
metrique = "0.1"
```

```rust
use metrique::writer::Entry;

#[derive(Entry)]
struct Globals {
    foo: String,
}
```

Previously, you had to have a `metrique-writer = "0.1"` line in your `Cargo.toml` as well.

Also, it seems that rstest is better developed than test_case. I don't think there is an urgent reason to migrate, but I think people might be using `metrique` as an example and I'll like to lead them to a pit of success

### DIFF
--- a/metrique/src/lib.rs
+++ b/metrique/src/lib.rs
@@ -417,12 +417,17 @@ pub mod writer {
     pub use metrique_writer_macro::MetriqueEntry as Entry;
 
     pub use metrique_writer::AttachGlobalEntrySinkExt;
-
-    #[cfg(feature = "test-util")]
-    pub use metrique_writer::test_util;
+    pub use metrique_writer::{AttachGlobalEntrySink, EntryIoStreamExt, FormatExt};
     pub use metrique_writer::{entry, format, sample, sink, stream, value};
 
+    #[cfg(feature = "test-util")]
+    #[doc(hidden)] // prefer the metrique::test_util re-export
+    pub use metrique_writer::test_util;
+
+    #[doc(hidden)] // prefer the metrique::unit re-export
+    pub use metrique_writer::unit;
+
+    // used by macros
     #[doc(hidden)]
     pub use metrique_writer::core;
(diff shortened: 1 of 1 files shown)


metrique/src/lib.rs
```diff
@@ -403,4 +403,26 @@ pub use metrique_metricsrs as metrics_rs;
...
+
+    pub use metrique_writer::AttachGlobalEntrySinkExt;
+
+    #[cfg(feature = "test-util")]
+    pub use metrique_writer::test_util;
+    pub use metrique_writer::{entry, format, sample, sink, stream, value};
+
+    #[doc(hidden)]
+    pub use metrique_writer::core;
+    pub use metrique_writer::{AttachGlobalEntrySink, EntryIoStreamExt, FormatExt, unit};
```
rcoh, line 427: `unit` should be `doc(hidden) in this position since its already ex-exported from metrique.

metrique/src/lib.rs
```diff
@@ -403,4 +403,26 @@ pub use metrique_metricsrs as metrics_rs;
...
+    pub use metrique_writer::{MetricFlags, MetricValue, Observation, Value, ValueWriter};
+    pub use metrique_writer::{ValidationError, ValidationErrorBuilder};
+
+    // Use the variant of the macro that has `metrique::` prefixes.
+    pub use metrique_writer_macro::MetriqueEntry as Entry;
+
+    pub use metrique_writer::AttachGlobalEntrySinkExt;
+
+    #[cfg(feature = "test-util")]
+    pub use metrique_writer::test_util;
```
rcoh, line 422: we re-export this above already at the root. make this one doc hidden?

rcoh (approved):

PR #98, opened 2025-09-10
chore(docs): clarify that you can also bring your own format with the Format trait

jlizen:
📬 *Issue #, if available:*

✍️ *Description of changes:*

Currently our docs read like the library only offers EMF, or else you're out of luck until new formats are added. In fact you can bring your own format if you implement the `Format` trait. It's not trivial work, and this PR doesn't add a full-form guide, but users can at least reference the Emf implementation of `Format`.

For now, I just added that breadcrumb to places where we say 'only emf is supported for now, but more formats might be supported in the future'. I also tweaked our link to point to our `metrique`-re-export of the metrique-writer types since that keeps people from needing to jump around as much (they probably DO want to immediately go to the EMF implementation).

Context: I was having a conversation with a prospective user that was interested in bringing their own minimalist, non-standard format. If they are willing to do the legwork to implement the `Format` trait, `metrique` would support that just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 A set of crates for collecting and exporting metrics, especially unit-of-work metrics.
 
 This currently supports exporting metrics in [Amazon EMF] format to CloudWatch.
-More formats might be supported in future versions.
-
+More formats might be supported in future versions. You can also configure a custom
+format using the [`Format`] trait.
 
 ## Getting Started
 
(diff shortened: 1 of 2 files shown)


metrique-writer/README.md
```diff
@@ -229,4 +231,4 @@ queue.append_any(metric);
 ```
 
 [emf-docs]: htps://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html
-[`Format`]: https://docs.rs/metrique-writer/0.1/metrique_writer/format/trait.Format.html
+[`Format`]: https://docs.rs/metrique-writer/0.1/metrique_writer/format/trait.Format.html
```
jlizen: no idea what changed here... whitespace?

README.md
```diff
@@ -127,6 +127,9 @@ fn initialize_metrics(service_log_dir: PathBuf) -> AttachHandle {
 > See [`metrique-writer`](metrique-writer) for more information about queues and destinations.
 
 You can either attach it to a global destination or thread the queue to the location you construct your metrics object directly. Currently, only formatters for [Amazon EMF] are provided, but more may be added in the future.
+You can also implement a custom
```
arielb1: Should you insert a paragraph break here? Or these should be on the same line?

arielb1 (approved):

PR #112, opened 2025-10-17
Add aggregation RFC

rcoh:
📬 *Issue #, if available:* 

✍️ *Description of changes:* Adds design docs for aggregation support

[Public API docs](https://github.com/awslabs/metrique/blob/aggregation-rfc/docs/aggregated.md)

[Internals docs](https://github.com/awslabs/metrique/blob/aggregation-rfc/docs/aggregated-internals.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ enum Operation {
 
 #[metrics]
 struct RequestMetrics {
-    operation: Operation, /* you can use `operation: &'static str` if you prefer */
+    operation: Operation, // you can use `operation: &'static str` if you prefer
     #[metrics(timestamp)]
     timestamp: Timestamp,
     number_of_ducks: usize,


docs/aggregated.md
```diff
@@ -0,0 +1,423 @@
...
+    #[metrics(aggregate = Counter)]
+    processing_errors: u64,
+}
+
+async fn setup_queue_processor() {
+    // Wrap your normal sink with aggregation
+    let base_sink = ServiceMetrics::sink();
+    let aggregating_sink = AggregatingEntrySink::with_config(
+        base_sink,
+        AggregateConfig {
```
jlizen: Minor: It was pretty disorienting to see these two properties with no real explanation. I see later down:
```
let config = AggregateConfig {
    max_entries: 1000,           // Flush when 1000 unique keys accumulated
    raw_events_per_second: 5.0,  // Sample 5 raw events per second for debugging
};
```

Might be worth hoisting those comments up?

docs/aggregated.md
```diff
@@ -0,0 +1,423 @@
...
+
+let config = AggregateConfig {
+    max_entries: 1000,           // Flush when 1000 unique keys accumulated
+    raw_events_per_second: 5.0,  // Sample 5 raw events per second for debugging
+};
+
+let sink = AggregatingEntrySink::with_config(base_sink, config);
+```
+
+## Sampling Integration
```
jlizen: Will this be optional? Will it support the congressional sampling?

docs/aggregated.md
```diff
@@ -0,0 +1,423 @@
...
+        // Metrics automatically aggregated and emitted when dropped
+    }
+
+    // Periodically flushes aggregated results + sampled raw events
+}
+```
+
+**Output**: Multiple aggregated entries like `ItemType: "email", Priority: 1, ItemsProcessed: 1247, ProcessingTime: [histogram], ProcessingErrors: 23`
+
+## Aggregation Strategies
```
jlizen: Is there a way for customers to define custom aggregation strategies if they implement `AggregateValue<T>` rather than use `aggregate` field-level macro?

What would that look like? Is that worth mentioning in the docs?

I guess they could also implement a custom entry-level aggregation, but that would involve going off the rails of the macro entirely so probably not worth focusing on...?

docs/aggregated.md
```diff
@@ -0,0 +1,423 @@
...
+        }
+
+        // Metrics automatically aggregated and emitted when dropped
+    }
+
+    // Periodically flushes aggregated results + sampled raw events
+}
+```
+
+**Output**: Multiple aggregated entries like `ItemType: "email", Priority: 1, ItemsProcessed: 1247, ProcessingTime: [histogram], ProcessingErrors: 23`
```
jlizen: What do the sampled events look like? How do they avoid metric pollution? Is the guarantee that, if they are sampled, they just won't be also aggregated? It's worth stating as such if so.

PR #117, opened 2025-10-31
MultiFlex

rcoh:
📬 *Issue #, if available:*

✍️ *Description of changes:*
Produces `devices.0.size` etc.
```rust
#[metrics]
struct MyDevices {
    // devices.0.size etc.
    #[metrics(flatten, prefix = "devices")]
    devices: MultiFlex<Device>,
}

#[metrics]
struct Device {
    id: String,
    size: usize,
}

impl FlexItem for Device {
    fn prefix_item(&self, idx: usize) -> std::borrow::Cow<'static, str> {
        return Cow::Owned(format!(".{idx}."));
    }
}
```

The implementation does way more allocations than should be needed, but I didn't find any super easy ways to fix them with the current lifetime issues.

### DIFF
--- a/metrique/examples/multi-flex.rs
+++ b/metrique/examples/multi-flex.rs
@@ -1,0 +1,83 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! This example demonstrates using MultiFlex to emit metrics for dynamic lists of items.
+//!
+//! MultiFlex is useful when you have a variable number of similar items to track metrics for,
+//! such as multiple database connections, API endpoints, or processing stages.
+
+use metrique::{
+    emf::Emf,
+    multi_flex::{FlexItem, MultiFlex},
+    unit_of_work::metrics,
+    writer::{AttachGlobalEntrySinkExt, FormatExt, GlobalEntrySink, sink::global_entry_sink},
+};
+use std::borrow::Cow;
+
+global_entry_sink! { ServiceMetrics }
+
+#[metrics]
+struct RequestMetrics {
+    request_id: String,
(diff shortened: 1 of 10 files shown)


metrique/examples/multi-flex.rs
```diff
@@ -0,0 +1,110 @@
...
+#[metrics]
+#[derive(Debug, Clone)]
+struct EndpointCall {
+    name: String,
+    response_time_ms: u64,
+    status_code: u16,
+}
+
+impl FlexItem for EndpointCall {
+    fn prefix_item(&self, idx: usize) -> Cow<'static, str> {
```
jlizen: Seems like this could be a macro annotation, if it is common enough?

jlizen (commented):
It's not clear to me how much value is derived from having metrics with eg `endpoint.0.success`, since any sensible alarming or metrics needs deterministic naming. I guess it's nice for deep dive debugging.

What would be more useful would be having the ability to associate a 'key' field with a dynamic value that is hoisted into the metric name, rather than naming via indexing.

metrique/src/multi_flex.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    fn sample_group(&self) -> impl Iterator<Item = SampleGroupElement> {
+        // sample groups are ignored inside of multi-groups
+        vec![].into_iter()
+    }
+}
+
+struct DynPrefix<'b, T> {
```
arielb1: (you can hold `idx` and `item` within the `DynPrefix` and then you will not need the `SmartString`. You'll still need to allocate a `String` for every item.

arielb1 (changes requested):
Have comments.

Maybe worth having a comment that this will do an allocation for every metric and therefore is slightly slow?

PR #122, opened 2025-11-09
feat: implement `#[metrics(explicit_prefix)]` which is not inflected

### LINKED ISSUE #121: #metrics prefix & rename_all drop characters
```
#[metrics(rename_all = "PascalCase" prefix = "api_")]
struct MyMetrics {
  some_thing
}
```
Results in a metric name `ApiSomeThing`.   This is true for any character
one might use instead of `_`, such as `.`, `/`, or `:`.  Multiple symbols
are collapsed.

In converting to snake case, you get `api-some-thing`.  If you don't have
a delimiter, ("api" instead of "api_") the you get `apisome-thing`.  So the
delimiter is load bearing today.

In our metrics a colon is often used as a scoping mechanism, so we would
like `Api:SomeThing`, but we can't get that using `prefix` & `rename_all`,
we need to get there with `#[metric(name="...")]` for each field.

Since rust fields start as snake case, the request is to constrain the symbols
which act as word boundaries to `_` and pass through other symbols. All
the documentation examples use `_`.

By example:
```
#[metrics(rename_all = "PascalCase" prefix = "Api:_")]
struct MyMetrics {
  some_thing
}
```
Would result in
 * Api:_some_thing (orig)
 * api:_some_thing (snake)
 * api:-some-thing (kabob)
 * Api:SomeThing (pascal)

This depends on programmers to make good choices which work for
their consumers, but this is source code.

Thanks for the consideration.

arielb1:
This is a breaking change since we intentionally forbid prefixes that contain non-standard non-alphanumeric characters, but these normally did not do what you want.

Fixes #121

📬 *Issue #, if available:*

✍️ *Description of changes:*

### DIFF
--- a/metrique-macro/src/lib.rs
+++ b/metrique-macro/src/lib.rs
@@ -76,11 +76,12 @@ use crate::inflect::{metric_name, name_contains_dot, name_contains_uninflectable
 ///
 ///    ```rust
 ///    # use metrique::unit_of_work::metrics;
+///    # use std::time::Duration;
 ///    #[metrics(subfield)]
 ///    struct Subfield {
-///        field_val: u32, // inflected
-///        #[metrics(name="SomeName")] // not inflected (since `name` is not inflected), prefixed
-///        some_name: u32,
+///        request_latency: Duration, // inflected
+///        #[metrics(name="NDucks")] // not inflected (since `name` is not inflected), prefixed
+///        number_of_ducks: u32,
 ///    }
 ///
 ///    #[metrics(rename_all = "kebab-case")]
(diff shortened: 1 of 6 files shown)


metrique/tests/ui/fail/bad_field_attrs.rs
```diff
@@ -34,11 +34,27 @@ struct MetricsWithInvalidUnit {
...
 struct SubMetric {
     a: usize,
 }
 
+#[metrics(prefix = "foo", explicit_prefix = "foo")]
+struct PrefixAndExplicit {
+    a: usize,
+}
+
+#[metrics(prefix = "foo@")]
```
rcoh: just for the purposes of integration testing, we should probably use `:` that is the field most commonly used in the wild right now

metrique/tests/renames.rs
```diff
@@ -56,3 +65,18 @@ fn metrics_renames_work() {
     // For sub metrics which _don't_ set a rename, they get rename transitively from the pattern
     assert_eq!(entry.metrics["sub-field-a"], 4);
 }
+
+#[test]
+fn explicit_prefix_metrics() {
+    let metrics = ExplicitPrefixedMetrics {
+        a: 100,
+        local_rename: "abcd",
+    };
```
rcoh: some additional tests that are probably useful, at least to capture the current behavior
```
#[metrics(explicit_prefix = "API@")]
struct Outer {
    #[metrics(flatten, prefix = "inner_")]  // Mix of explicit + inflectable
    inner: Inner,
}
```

```
#[metrics(explicit_prefix = "API@")]
struct Outer {
    #[metrics(flatten, explicit_prefix = "Inner:")]
    inner: Inner,  // Does this become "API@Inner:field"?
}
```

metrique/tests/ui/fail/bad_metrics_value.stderr
```diff
@@ -114,13 +150,13 @@ error[E0277]: `FormattedValue<'_, u32, Foo, _>` is not a metric value
            and $N others
 
 error[E0277]: CloseValue is not implemented for &String
```
rcoh: I thought we improved this?

metrique/tests/renames.rs
```diff
@@ -26,6 +26,15 @@ struct PrefixedMetrics {
     local_rename: &'static str,
 }
 
+#[metrics(rename_all = "snake_case", explicit_prefix = "prefix@")]
+struct ExplicitPrefixedMetrics {
+    a: usize,
+
+    // local renames take precence, this field is just called `name`
+    #[metrics(name = "name")]
```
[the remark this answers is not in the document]
arielb1: *external* prefixes apply to such a field, internal prefixes don't

metrique/tests/renames.rs
```diff
@@ -56,3 +65,18 @@ fn metrics_renames_work() {
     // For sub metrics which _don't_ set a rename, they get rename transitively from the pattern
     assert_eq!(entry.metrics["sub-field-a"], 4);
 }
+
+#[test]
+fn explicit_prefix_metrics() {
```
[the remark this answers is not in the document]
arielb1: done

metrique-macro/src/lib.rs
```diff
@@ -471,12 +485,52 @@ struct MetricsFieldAttrs {
...
+}
+
+impl Prefix {
+    fn from_inflectable_and_explicit(
+        inflectable: &Option<SpannedKv<String>>,
+        explicit: &Option<SpannedKv<String>>,
+    ) -> darling::Result<Option<SpannedValue<Self>>> {
+        match (inflectable, explicit) {
+            (Some(prefix), None) => {
+                if name_contains_uninflectables(&prefix.value) {
```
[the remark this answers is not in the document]
arielb1: done

metrique-macro/src/lib.rs
```diff
@@ -59,13 +65,130 @@ use crate::inflect::metric_name;
...
+///    ```rust
+///    # use metrique::unit_of_work::metrics;
+///    #[metrics(subfield)]
+///    struct Subfield {
+///        field_val: u32 // inflected
+///     }
+///
+///     #[metrics(prefix = "Foo-" /* prefix gets inflected */, rename_all = "kebab-case")]
+///     struct Base {
+///         // prefix does not propagate to subfield
```
rcoh:
```suggestion
///         // prefix does not propagate to subfield. Use `prefix = "Foo-"` to propagate
```

metrique-macro/src/lib.rs
```diff
@@ -471,12 +612,73 @@ struct MetricsFieldAttrs {
...
+    fn inflected_prefix_message(c: char, is_warning: bool) -> String {
+        let warning_text = if is_warning {
+            " `.` is currently allowed in an inflected prefix, but will be disallowed in a future version."
+        } else {
+            ""
+        };
+        format!(
+            "Found {c:?} in inflected prefix. `prefix` inflection treats all non-alphanumeric \
+                        characters identically. Use `-` or `_` for an inflectable separator, or use `exact_prefix` for \
+                        a separator that will not be inflected.{warning_text}"
```
rcoh: "Or use `exact_prefix` to preserve special characters in the metric name"?

I worry that people don't necessarily know exactly what is meant by "inflected"

metrique-macro/src/snapshots/metrique_macro__tests__field_prefix_warning_dot.snap
```diff
@@ -0,0 +1,179 @@
...
+    #[deprecated(
+        note = "these fields will become private in a future release. To introspect an entry, use `metrique::writer::test_util::test_entry`"
+    )]
+    #[doc(hidden)]
+    operation: <&'static str as metrique::CloseValue>::Closed,
+}
+const _: () = {
+    const _: () = {
+        #[deprecated(
+            note = "Found '.' in inflected prefix. `prefix` inflection treats all non-alphanumeric characters identically. Use `-` or `_` for an inflectable separator, or use `exact_prefix` for a separator that will not be inflected. `.` is currently allowed in an inflected prefix, but will be disallowed in a future version."
```
rcoh: I think overall we could make a more obvious error message here

rcoh (approved):

PR #129, opened 2025-12-04
feat: show useful error information on validation error with no tracing

### LINKED ISSUE #128: Make errors more obvious when configuration is invalid
Frequently, when builders first get started, they often have a slightly misconfigured metric struct. For example, they might be missing a field required by their formatter.

Right now, this fails semi-silently (It is only visible if you have a tracing subscriber attached).

In debug builds only, we should consider a louder failure (or other mechanism to ease this common onboarding stumbling block)

arielb1:
Fixes #128

This feels a little "over-engineered" but I can't find a simpler way that will be format-generic, and it does not cause that much collateral impact.

📬 *Issue #, if available:*

✍️ *Description of changes:*

### DIFF
--- a/metrique-writer-core/src/config.rs
+++ b/metrique-writer-core/src/config.rs
@@ -26,11 +26,11 @@ impl AllowSplitEntries {
 impl EntryConfig for AllowSplitEntries {}
 
 /// This config is used for basic error messages. It allows generating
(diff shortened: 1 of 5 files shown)


rcoh (commented):
overall lgtm, just a couple of small naming things

PR #143, opened 2025-12-23
feat: Implement `Histogram`

rcoh:
📬 *Issue #, if available:* #112

✍️ *Description of changes:* Starts a new `metrique-aggregation` crate by adding a `Histogram` type

### Why a new crate?
Eventually, I `metrique` will re-export this crate. However, I suspect we will require a few releases with breaking changes to get the full API perfect. This allows us to do that without too much downstream impact.

### What's added here

The new crate has two types `Histogram` and AtomicHistogram`:
```rust
#[derive(Default)]
struct TestMetrics {
    latency: Histogram<Duration>,
    #[metrics(unit = Byte)]
    size: Histogram<u32>,

    // for thread safe, use AtomicHistogram
    atomics: AtomicHistogram<usize>,

    // other strategies are available, e.g. SortAndMerge preserves all
    // data points
    all_values: Histogram<SortAndMerge>,
}
```

Histogram respects units and dimensions.

### DIFF
--- a/metrique-aggregation/src/histogram.rs
+++ b/metrique-aggregation/src/histogram.rs
@@ -177,8 +177,10 @@ impl<T, S: AggregationStrategy> Histogram<T, S> {
                         Observation::Unsigned(v) => self.0.record(v as f64),
                         Observation::Floating(v) => self.0.record(v),
                         Observation::Repeated { total, occurrences } => {
-                            let avg = total / occurrences as f64;
-                            self.0.record_many(avg, occurrences);
+                            if occurrences > 0 {
+                                let avg = total / occurrences as f64;
+                                self.0.record_many(avg, occurrences);
+                            }
                         }
                         _ => {}
                     }
@@ -259,8 +261,10 @@ impl<T, S: SharedAggregationStrategy> SharedHistogram<T, S> {
                         Observation::Unsigned(v) => self.0.record(v as f64),
                         Observation::Floating(v) => self.0.record(v),
                         Observation::Repeated { total, occurrences } => {
-                            let avg = total / occurrences as f64;
-                            self.0.record_many(avg, occurrences);
+                            if occurrences > 0 {
+                                let avg = total / occurrences as f64;
+                                self.0.record_many(avg, occurrences);
+                            }
                         }
                         _ => {}
                     }
--- a/metrique-aggregation/tests/histogram.rs
+++ b/metrique-aggregation/tests/histogram.rs
@@ -7,8 +7,8 @@ use metrique_aggregation::histogram::{
     Histogram, SharedAggregationStrategy, SharedHistogram, SortAndMerge,
 };
 use metrique_writer::Observation;
-use metrique_writer::unit::{Byte, Millisecond};
-use metrique_writer::value::WithDimensions;
+use metrique_writer::unit::{Byte, Millisecond, UnitTag};
+use metrique_writer::value::{MetricFlags, MetricValue, ValueWriter, WithDimensions};
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 use rstest::rstest;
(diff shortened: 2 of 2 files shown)


metrique-aggregation/src/histogram.rs
```diff
@@ -0,0 +1,439 @@
...
+                _unit: metrique_writer::Unit,
+                _dimensions: impl IntoIterator<Item = (&'a str, &'a str)>,
+                _flags: MetricFlags<'_>,
+            ) {
+                for obs in distribution {
+                    match obs {
+                        Observation::Unsigned(v) => self.0.record(v as f64),
+                        Observation::Floating(v) => self.0.record(v),
+                        Observation::Repeated { total, occurrences } => {
+                            let avg = total / occurrences as f64;
```
arielb1: We should have some sort of record_many there - occurrences can be large.

It is fine if the sorted aggregation strategy does not have a count field, but the exponential aggregation strategy should be O(1) in the number of occurrences.

metrique-aggregation/src/histogram.rs
```diff
@@ -0,0 +1,439 @@
...
+            strategy,
+            _value: PhantomData,
+        }
+    }
+
+    /// Add a value to the histogram.
+    ///
+    /// The value is converted to observations using the metric value's implementation,
+    /// then recorded in the aggregation strategy.
+    pub fn add_value(&mut self, value: T)
```
arielb1: Is it possible to do some unit management here? (Tho we want to add both Duration and f64, which means we can't condition on MetricValue being anything in particular).

Or maybe unit management does not make sense in an "aggregated" context, and there should just be a comment (or an assertion?) that you should be consistent with units.

metrique-aggregation/src/histogram.rs
```diff
@@ -0,0 +1,439 @@
...
+    /// Drain all observations and return them as a vector.
+    ///
+    /// This resets the strategy's internal state.
+    fn drain(&mut self) -> Vec<Observation>;
+}
+
+/// Thread-safe strategy for aggregating observations in a histogram.
+///
+/// Like [`AggregationStrategy`] but allows recording values through a shared reference.
+pub trait AtomicAggregationStrategy {
```
arielb1: Maybe this should be "SharedAggregationStrategy" rather than "Atomic"? You can have a shared non-atomic aggregated strategy (using `RefCell` or the like) using this trait

arielb1 (changes requested):
see comments

metrique-aggregation/src/histogram.rs
```diff
@@ -0,0 +1,444 @@
...
+
+impl AggregationStrategy for ExponentialAggregationStrategy {
+    fn record(&mut self, value: f64) {
+        self.inner.add(value as u64, 1).ok();
+    }
+
+    fn drain(&mut self) -> Vec<Observation> {
+        let snapshot = std::mem::replace(
+            &mut self.inner,
+            histogram::Histogram::with_config(&histogram::Config::new(4, 32).unwrap()),
```
jasgin: `.expect("known good configuration")`?

metrique-aggregation/src/histogram.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+}
+
+impl<const N: usize> AggregationStrategy for SortAndMerge<N> {
+    fn record(&mut self, value: f64) {
+        self.values.push(value);
+    }
+
+    fn drain(&mut self) -> Vec<Observation> {
+        self.values.sort_by(|a, b| a.partial_cmp(b).unwrap());
```
jasgin: Is it possible to end up in a case where comparison is impossible (e.g. NaN) and panics?

arielb1: probably so - you should be using a nan-safe function

metrique-aggregation/src/histogram.rs
rcoh: fixed in https://github.com/awslabs/metrique/pull/143/commits/f383d9661d134e159086ac3d874e26135894c552

There is a separate issue to work on which is that the Durations are only going to store millisecond precision. I have some ideas to address that but I'll address that in a follow up

metrique-aggregation/src/histogram.rs
arielb1: you should handle occurrences == 0

metrique-aggregation/src/histogram.rs
```diff
@@ -0,0 +1,482 @@
...
+/// Uses exponential bucketing with configurable precision. Default configuration
+/// uses 4-bit mantissa precision (16 buckets per order of magnitude, ~6.25% error).
+pub struct ExponentialAggregationStrategy {
+    inner: histogram::Histogram,
+}
+
+impl ExponentialAggregationStrategy {
+    /// Create a new exponential aggregation strategy with default configuration.
+    pub fn new() -> Self {
+        let config = histogram::Config::new(4, 32).expect("known good configuration");
```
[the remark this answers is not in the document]
rcoh: fixed

PR #148, opened 2025-12-26
fix: Add scaling to ExponentialAggregationStrategy

rcoh:
📬 *Issue #, if available:* N/A

✍️ *Description of changes:* Scale entries inside of ExponentialHistogram by 1024x

If the internal `histogram` crate only supports whole numbers. Since `Duration`s are represented as Milliseconds, if we don't do any scaling, we lose the ability to have buckets that capture microsecond precision.

This currently isn't configurable although we can always make it configurable if that becomes needed in the future.

I also added the `test_metric` helper.

### DIFF
--- a/metrique-aggregation/tests/histogram.rs
+++ b/metrique-aggregation/tests/histogram.rs
@@ -7,6 +7,7 @@ use metrique_aggregation::histogram::{
     Histogram, SharedAggregationStrategy, SharedHistogram, SortAndMerge,
 };
 use metrique_writer::Observation;
+use metrique_writer::test_util::test_metric;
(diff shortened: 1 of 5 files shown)


metrique-aggregation/tests/histogram.rs
```diff
@@ -547,3 +560,27 @@ fn test_shared_histogram_ignores_zero_occurrences() {
     histogram.add_value(ZeroOccurrences);
     // Should not panic, just ignore the invalid observation
 }
+
+#[test]
+fn test_histogram_microsecond_accuracy() {
```
jlizen: Thanks for the detailed test!

jlizen (approved):

Issue #151, opened 2025-12-29
[metrics macro] Add support for non-value Enums


jlizen:
Currently our `#[metrics]` macro only supports Value enums (`#[metrics(value(string))]`). That converts each variant to a string.

I want macro support for instead having variants that contain their own entry structs, that I flatten down.

My use case is that I want to have a parent metrics struct that contains one of several operations. 

```
#[metrics]
struct MyMetric {
    success: bool,
    #[metrics(flatten)]
    operation: OperationMetric
}

#[metrics(subfield_owned)]
enum OperationMetric {
    OperationA(AMetric),
    OperationB(BMetric),
}

#[metrics(subfield_owned, prefix = "a_")]
struct AMetric {
   success:bool,
   frobs: usize,
   property: String,
}

#[metrics(subfield_owned, prefix = "b_")]
struct BMetric {
   success:bool,
   frongles: usize,
   property: String,
}
```

jlizen:
One possible bikeshed would be, it would be nice to be able to still have the value field as well, since that commonly will be desired when nesting variants.

IE: 
```
#[metrics]
struct MyMetric {
    success: bool,
    #[metrics(flatten, value)
    operation: OperationMetric
}

#[metrics(subfield_owned)]
enum OperationMetric {
    A(AMetric),
    B(BMetric),
}

..
```

Would produce: 
```
{
   success: bool,
   a_success: bool,
   operation: "a"
}
```

jlizen:
Seems like one sort of concise way to accomplish this today is via [EnumMapEntry](https://docs.rs/metrique/0.1.11/metrique/writer/entry/struct.EnumMapEntry.html)?

Still fighting with a custom `Entry` impl that doesn't need to know about all the fields...

PR #156, opened 2025-12-31
Add support for entry enums

jlizen:
📬 *Issue #, if available:*

closes: https://github.com/awslabs/metrique/issues/151

✍️ *Description of changes:*

Adding enum entry support. It does what you would expect and supports either variants with tuples (which must be `flatten`, `flatten_entry`, or `ignore`) or variants with structs (fields behave like normal struct fields).

Probably the best introductory view would be [the new macro doc section](https://github.com/jlizen/metrique/blob/6c04b0cfac7eac45edbcb9ce279d8046399a6c47/metrique-macro/src/lib.rs#L246).

There is also a new `metrique/examples/enums.rs`.

### DIFF
--- a/metrique-macro/src/lib.rs
+++ b/metrique-macro/src/lib.rs
@@ -601,12 +601,10 @@ impl RawRootAttributes {
                 MetricMode::RootEntry | MetricMode::Subfield | MetricMode::SubfieldOwned => {
                     Ok(tag.into_inner().into())
                 }
-                MetricMode::Value | MetricMode::ValueString => {
-                    Err(darling::Error::custom(
-                        "value and value(string) do not support tag",
-                    )
-                    .with_span(&tag.span()))
-                }
+                MetricMode::Value | MetricMode::ValueString => Err(darling::Error::custom(
+                    "value and value(string) do not support tag",
+                )
+                .with_span(&tag.span())),
             })
             .transpose()?;
 


README.md
```diff
@@ -70,7 +70,10 @@ enum Operation {
 
 #[metrics]
 struct RequestMetrics {
-    operation: Operation, // you can use `operation: &'static str` if you prefer
+    operation: Operation,
+    request_id: String,
```
jlizen: I added a few fields that I think make the example a bit more realistic / covers things people get wrong a lot (bool as 0/1). But does this make it too long? Open to adding more white space, or slimming it back now.

Same goes for the one at top of `metrique/README.md`

metrique/examples/enums.rs
```diff
@@ -0,0 +1,100 @@
...
+    operation: OperationType,
+
+    #[metrics(flatten)]
+    details: OperationMetrics,
+
+    success: bool,
+}
+
+fn main() {
+    let sink = VecEntrySink::new();
```
rcoh: why did you use a VecEntrySink instead of just using the inspector since you are running it through `test_test_entry` anyway?

Alternatively, since the point of this is not queue related, you could use the new `test_metric` API that would allow you to do:

```rust
let entry = test_metric(RequestMetrics {
            operation: OperationType::Read,
            details: OperationMetrics::Read(ReadMetrics {
                bytes_read: 1024,
                cache_hit: true,
            }),
            success: true,
        }
)
```

rcoh (commented):
I haven't gotten into the internals much yet. WDYT about merging all your refactorings as a separate PR to avoid conflicts then working on the enum stuff?

README.md
```diff
@@ -79,6 +82,24 @@ struct RequestMetrics {
 }
 ```
 
+Enums can also be used as metrics entries, with different fields per variant:
```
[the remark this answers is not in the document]
jlizen: The fields are flushed the same as it they were in structs? I can add in a sample output though, just adds a bit more verbosity.

metrique/examples/enums.rs
jlizen: No good reason, I'll change it

README.md
arielb1: OK

metrique-macro/src/entry_impl.rs
```diff
@@ -46,226 +22,455 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
+    use crate::{MetricsField, MetricsFieldKind, value_impl::format_value};
+
+    pub(crate) fn generate_struct_entry_impl(
+        entry_name: &Ident,
+        fields: &[MetricsField],
+        root_attrs: &RootAttributes,
+    ) -> Ts2 {
+        let writes = generate_write_statements(fields, root_attrs);
+        let sample_groups = generate_sample_group_statements(fields, root_attrs);
+        // we generate one entry impl for each namestyle. This will then allow the parent to
```
arielb1: This comment is wrong, should be removed

metrique-macro/src/entry_impl.rs
```diff
@@ -46,226 +22,455 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
+        // Split the iterators in half and recursively build the tree
+        let mid = iterators.len() / 2;
+        let left = make_binary_tree_chain(iterators[..mid].to_vec());
+        let right = make_binary_tree_chain(iterators[mid..].to_vec());
+
+        quote! { #left.chain(#right) }
     }
+}
+
+pub(crate) mod enum_impl {
```
arielb1: These should probably be their own files?

metrique-macro/src/entry_impl.rs
```diff
@@ -46,226 +22,455 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
+        entry_name: &Ident,
+        variants: &[MetricsVariant],
+        root_attrs: &RootAttributes,
+    ) -> Ts2 {
+        let write_arms = variants.iter().map(|variant| {
+            let variant_ident = &variant.ident;
+
+            match &variant.data {
+                None => {
+                    // Defensive guard: unit variants rejected earlier during parsing
```
arielb1: If it's not reachable, use an `unreachable!`

metrique-macro/src/entry_impl.rs
```diff
@@ -46,226 +22,455 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
+                        #entry_name::#variant_ident => {
+                            compile_error!("unit variants are not allowed in entry enums; use #[metrics(value(string))] for unit-only enums");
+                        }
+                    )
+                }
+                Some(VariantData::Tuple { kind, .. }) => {
+                    match kind {
+                        MetricsFieldKind::Flatten { span, prefix } => {
+                            // Container-level rename_all and prefix do NOT apply to flattened fields
+                            // See: https://github.com/awslabs/metrique/issues/160
```
arielb1: rename_all *does* apply to flattened fields?

metrique-macro/src/entry_impl.rs
```diff
@@ -46,226 +22,455 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
+}
+
+pub(crate) mod enum_impl {
+    use super::*;
+    use crate::{
+        MetricsFieldKind,
+        enums::{MetricsVariant, VariantData},
+    };
+
+    pub(crate) fn generate_enum_entry_impl(
```
arielb1: shouldn't the code for an enum variant be the same as the code for a struct, matching the normal Rust structure where a struct is an enum with 1 variant?

metrique-macro/src/enums.rs
```diff
@@ -0,0 +1,420 @@
...
+                ));
+            }
+
+            let field = &fields.unnamed[0];
+            let raw_attrs = RawMetricsFieldAttrs::from_field(field)?;
+            let attrs = raw_attrs.validate()?;
+
+            match &attrs.kind {
+                MetricsFieldKind::Flatten { .. } | MetricsFieldKind::FlattenEntry(_) => {}
+                _ => {
```
arielb1: Exhaustive match?

metrique-macro/src/enums.rs
```diff
@@ -0,0 +1,420 @@
...
+}
+
+fn parse_variant_data(fields: &syn::Fields) -> Result<Option<VariantData>> {
+    match fields {
+        syn::Fields::Unit => Ok(None),
+        syn::Fields::Unnamed(fields) => {
+            if fields.unnamed.len() != 1 {
+                return Err(syn::Error::new_spanned(
+                    fields,
+                    "tuple variants must have exactly one field",
```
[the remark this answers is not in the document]
jlizen: fair, also i was thinking about this, and there isn't really a reason to block multiple `#[flatten]` tuple fields anyway

jlizen:
#162 landed which will simplify this PR. it also fixes the trybuild mismatches (the issue was, 1.89 is source of truth).

More feedback offline from @arielb1 :
> btw, maybe we should merge the enum-variant and struct code
> since currently it seems duplicated
> take a look, either find a good reason why the code should be duplicated or de-dupe it
> most derives work like this

metrique-macro/src/entry_impl.rs
jlizen: Yeah, I started off with it that way, then thought better of it. I'll go back to my original plan, which is just making them a submodule of `enums.rs` and `structs.rs`

metrique/examples/enums.rs
```diff
@@ -0,0 +1,100 @@
...
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    #[metrics(sample_group)]
+    operation: OperationType,
+
+    #[metrics(flatten)]
+    details: OperationMetrics,
+
+    success: bool,
+}
```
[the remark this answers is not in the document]
jlizen: I've been thinking more about this and have concluded that for structs:
`#[metrics(name = "foo")] struct Foo {}` sounds cute because of symmetry but is not actually all that useful. 

It pretty much would only be useful if you want to have eg `#[metrics(flatten_entry, sample_group)] plugin_metrics: Box<dyn InflectableEntry + SampleGroup>`. Which seems a bit premature to optimize for as a use case.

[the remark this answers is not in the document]
jlizen: And then, for field-level `#[metrics(flatten)] my_enum: MyEnum`... well, you could have `#[metrics(flatten, tag = "foo")] my_enum: MyEnum`, but now you would need to hoist the `SampleGroup` implementation from `MyEnum` to `MyEnumEntry` if you want to reuse its machinery, which seems confusing. And forcing callers into using `#[metrics(flatten_entry)]` is stinky.

It only really makes sense to interact with the `SampleGroup` impl on a subfield for value enums - which we already have, no `tag` needed. You can annotate them with `#[metrics(sample_group)]` already as well. That's their whole point.

So for entry enums clearest semantics are pretty much what you suggested... container-level
`#[metrics(tag(name="foo"))]`, which defers to the existing `SampleGroup` impl for `MyEnum` (not `MyEnumEntry`).

And then, if you want container-level `#[metrics(tag(name="foo", sample_group)]`, that just chains the field into the rest of the fields it `MyEnumEntry`'s `sample_group()` codegen. And that's a lot easier to accomplish when we are already in the container's scope compared to the field-level case.

jlizen:
Pushed support for propagating sample groups from variant contents. Added a snap demonstrating. It generates a helper iterator struct to avoid boxing.

metrique-macro/src/lib.rs
```diff
@@ -43,11 +39,14 @@ use crate::inflect::{
 /// | `prefix` | String | Adds a prefix to all field names (prefix gets inflected) | `#[metrics(prefix = "api_")]` |
 /// | `exact_prefix` | String | Adds a prefix to all field names without inflection | `#[metrics(exact_prefix = "API_")]` |
 /// | `emf::dimension_sets` | Array | Defines dimension sets for CloudWatch metrics | `#[metrics(emf::dimension_sets = [["Status", "Operation"]])]` |
-/// | `sample_group` | Flag | On `#[metrics(value)]`, forwards `sample_group` to the inner field | `#[metrics(value, sample_group)]` |
+/// | `tag` | Nested | On entry enums, adds a tag field with the variant name. Tag value respects `rename_all` and variant `name`, but not `prefix`. | |
+/// | - `name` | String | Name of the tag field (not affected by `prefix` or `rename_all`) | `#[metrics(tag(name = "operation"))]` |
```
jlizen: It's definitely a bit awkward that you can't prefix eg a flattened `success` tag to `operation_1_success`. But, it felt too confusing to have `name` correspond to an inflectable name.

I was thinking we could just add `tag(base_name="..")`? Or we could do `tag(name="..", inflectable)` ? I can cut a backlog issue if anybody else thinks that seems useful.

metrique-macro/src/structs.rs
```diff
@@ -1,55 +1,21 @@
...
 };
 
 use crate::{
-    MetricMode, MetricsField, MetricsFieldKind, RawMetricsFieldAttrs, RootAttributes, clean_attrs,
-    entry_impl, generate_close_value_impls, generate_on_drop_wrapper, value_impl,
+    MetricMode, MetricsField, MetricsFieldKind, RootAttributes, clean_attrs, entry_impl,
+    generate_close_value_impls, generate_on_drop_wrapper, parse_metric_fields, value_impl,
 };
 
 fn parse_struct_fields(
```
arielb1: If the function just calls `parse_metric_fields`, it doesn't need to exist.

metrique-macro/src/snapshots/metrique_macro__tests__entry_enum_tag.snap
```diff
@@ -0,0 +1,326 @@
...
+    #[deprecated(
+        note = "these fields will become private in a future release. To introspect an entry, use `metrique::writer::test_util::test_entry`"
+    )]
+    #[doc(hidden)]
+    Read {
+        #[deprecated(
+            note = "these fields will become private in a future release. To introspect an entry, use `metrique::writer::test_util::test_entry`"
+        )]
+        #[doc(hidden)]
+        operation: ::std::borrow::Cow<'static, str>,
```
arielb1: Why do you need this field? Doesn't it have a fixed value?

metrique-macro/src/entry_impl/enum_impl.rs
```diff
@@ -0,0 +1,312 @@
...
+                    .iter()
+                    .filter_map(|field| collect_field_sample_group(field, root_attrs, |f| quote!(#f)))
+                    .unzip();
+
+                if let Some(tag_sg) = tag_sample_group {
+                    sample_groups.insert(0, tag_sg);
+                }
+
+                let pattern = match (tag_field.as_ref(), used_fields.is_empty()) {
+                    (Some(tag), true) => quote!(#entry_name::#variant_ident { #tag, .. }),
```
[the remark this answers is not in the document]
jlizen: Sure, though it'll need an extra flag since we need the `..` for non-sample-group fields to avoid lint warnings.

jlizen:
Refactored to address feedback. Notably:
- respect inflection with `tag(name="..")`, add `tag(name_exact="..")`
- don't store any tag field in the enum entry impl, instead just statically set it in the `write()` impl
- lots of other small refactors, thanks for the close read arielby@.

I started to poke at refactoring our organization more to  have a unified "generate entry impl" entrypoint for structs/enums. But also, our usage is woven together in lots of kind of awkward ways. I think we should tackle the refactor separately once this PR + aggregation PR have landed. I cut #167  with some ideas there.

AGENTS.md
```diff
@@ -3,6 +3,9 @@
 ## Testing
 - Use `cargo +1.89 nextest run` to run all tests in this workspace
 - If there are mismatches in trybuild or insta snapshots, share the diff for user approval before accepting them
+- For test utilities:
+  - Use `test_metric()` for short, focused examples (e.g., doc examples, simple assertions)
```
rcoh: should include a short snippet and location for each

metrique-macro/src/structs.rs
```diff
@@ -172,7 +134,7 @@ fn generate_entry_struct(
     let body = wrap_fields_into_struct_decl(has_named_fields, config.into_iter().chain(fields));
     Ok(quote!(
         #[doc(hidden)]
-        pub struct #name #body
+        #vis struct #name #body
```
rcoh: ah, I see. I wonder when this regressed...

In any case, I'd like to put this change in its own release we could yank in a non-breaking way if needed.

metrique/README.md
```diff
@@ -176,6 +177,51 @@ For more complex examples, see the [examples folder].
...
+```rust
+use metrique::unit_of_work::metrics;
+
+// generally entry enums will be used as subfields,
+// though they can also be root containers
+#[metrics(tag(name = "operation"), subfield)]
+enum Operation {
+    MeetDogs { dogs_met: usize },
+    FindGoose { goose_found: bool },
+    CountCats(#[metrics(flatten)] CatMetrics),
```
rcoh: would appreciate an example here or elsewhere showing how to combine this with `Instrumented`

metrique/examples/enums.rs
```diff
@@ -0,0 +1,119 @@
...
+    #[metrics(flatten)]
+    details: OperationMetrics,
+
+    success: bool,
+    request_id: String,
+}
+
+fn main() {
+    // Get operation
+    let get_entry = test_metric(RequestMetrics {
```
rcoh: consider making this _slightly_ more real to demonstrate patterns for actually producing the enum in a real-er code structure

metrique/tests/enum_basic.rs
```diff
@@ -0,0 +1,160 @@
...
+// Tuple variant with field-level prefix
+#[metrics]
+#[derive(Clone)]
+struct PrefixedNested {
+    metric: u32,
+}
+
+#[metrics]
+enum TuplePrefixEnum {
+    WithPrefix(#[metrics(flatten, prefix = "nested_")] PrefixedNested),
```
rcoh: should we emit a warning if you have `flatten`, a root prefix, but no inner prefix? I feel like its a little surprising. The caller could set `prefix = ""` to explicitly override it?

metrique/tests/ui/fail/bad_metrics_value.stderr
```diff
@@ -1,11 +1,12 @@
-error: Only structs are supported for entries
- --> tests/ui/fail/bad_metrics_value.rs:4:1
+error: enums must have at least one variant
```
rcoh: any good reason?

rcoh (commented):
no blockers

README.md
```diff
@@ -59,21 +59,18 @@ However, if your goal is to emit structured events that produce metrics with as
...
-#[metrics(value(string))]
-enum Operation {
-     CountDucks,
-}
-
 #[metrics]
 struct RequestMetrics {
-    operation: Operation, // you can use `operation: &'static str` if you prefer
+    request_id: String,
+    response_code: &'static str,
```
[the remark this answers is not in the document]
jlizen: It's probably how i would store it, since [http::StatusCode](https://docs.rs/http/latest/http/status/struct.StatusCode.html#method.as_str) offers it.

I guess you could make a custom formatter or we could add a blanket implementation, but it seems annoying for little benefit.

metrique/tests/enum_basic.rs
jlizen: (or might dogpile this one with it)

metrique-macro/src/entry_impl.rs
```diff
@@ -46,64 +25,193 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
                 });
             }
             MetricsFieldKind::Ignore(_) => {
                 continue;
             }
             MetricsFieldKind::Field { format, .. } => {
-                let (extra, name) = make_inflect_metric_name(root_attrs, field);
-                let value = format_value(format, field_span, quote! { &self.#field_ident });
+                let (extra, name) =
+                    make_inflect(&ns, &field.ident.to_string(), field_span, |style| {
```
arielb1: We do have `make_inflect_metric_name`? Can you use it?

metrique-macro/src/entry_impl.rs
```diff
@@ -46,64 +27,172 @@ fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
...
     }
+}
+
+/// Generate 4 ConstStr structs (one per naming style) and build an Inflect namespace type.
+/// The `name_fn` callback computes the string value for each style.
+/// Returns (extra_code, inflected_type).
+fn make_inflect(
+    ns: &Ts2,
+    inflect_method: syn::Ident,
+    base_name: &str,
```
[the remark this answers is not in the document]
jlizen: Ah, I recall the issue - `name_fn()` includes the prefix, which might be an exact_prefix, and therefore might break the identifier. So would need an extra bit of sanitization.

It also just looked a bit weirder. But anyway, it is scoped, so I am fine to cut to `name_fn(PascalCase)` and sanitize.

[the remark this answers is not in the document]
jlizen: (not to mention, lets us get rid of the `#[allow(non_camel_case)]`

README.md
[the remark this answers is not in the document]
jlizen: Exact the lifetime was wrong, you would have to use canonical-reason to get a `&'static str`, or else have closevalue on http::StatusCode (probably too opinionated for us to impil). I just stripped it out of the example. The only thing I think really matters to add is `bool`.

metrique/tests/ui/fail/bad_metrics_value.stderr
jlizen: It was annoying due to not being able to exhaustively match against an empty enum, need slightly different codegen. I don't mind adding it in though.

jlizen: Actually, i think we're better off leaving it out, it requires tweaking a lot of match statements and is basically entirely useless.

jlizen: That said i should have left the old test in too rather than modifying...

jlizen:
New changes up:

Refactors:
- allow unit enums in entry enums (useful alongside tag field)
- no longer pass around `base_name` when generating ConstStr names, instead use a sanitized/PascalCase version of the name_fn, rip out the defunct `allow(non_camel_case_types)` lint. This changes the codegenned type names in some cases, but it's still reasonably human-readable.
- make the enums.rs example a bit more realistic to show how you might construct an enum differently (though the examples will get a lot better once we can use #165 to have useful data-containing success vs error variants)
- show usage of entry enum in library-provided metrics examples
- move the tag attribute parsing that doesn't need input context into `RawRootAttribute::validate()` to give better span scoping on errors / simplify `generate_metrics` parsing

fixes:
- - restore the hardcoded pub visibility (and fix all my resulting lint warnings :/)
   -  it forces nested things to be pub even with private outers to avoid a pub-in-private warning, unfortunately (pre-existing issue)
   - I think it's probably a bad idea to programatically override that lint... we can clean up any examples forced to specify inner pub once we fix #163 
- entry enums should not prefix the variant name for its `From<T> for &str` impl
- enums properly reject `#[metrics(value)]` again

Issue #155, opened 2025-12-31
Feature Request: Aggregate Service Event Observations by Metric Dimension Set


LiAlexJ:
**Summary**

As an application owner, I would like a feature that aggregates multiple service event observations with the same metric dimension set into a single observation.

**Use Case**

When multiple events are observed at the same timestamp with identical dimension sets (e.g., AvailabilityZone and Id), they should be automatically aggregated rather than emitted as separate observations.

**Example**
Input: Multiple Events

Events observed at epoch 1767153321.144 for the API BatchPutData:

```
[
  {
    "AvailabilityZone": "us-east-2a",
    "Id": "d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f",
    "Failures": 300,
    "Success": 0
  },
  {
    "AvailabilityZone": "us-east-2a",
    "Id": "d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f",
    "Failures": 123,
    "Success": 0
  },
  {
    "AvailabilityZone": "us-east-2b",
    "Id": "02b8d786247046e98667cce4ca5af74e206f1793bf56f7a518ceb7fd",
    "Failures": 523,
    "Success": 0
  },
  {
    "AvailabilityZone": "us-east-2a",
    "Id": "d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f",
    "Failures": 0,
    "Success": 1500
  },
  {
    "AvailabilityZone": "us-east-2a",
    "Id": "d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f",
    "Failures": 0,
    "Success": 1300
  },
  {
    "AvailabilityZone": "us-east-2b",
    "Id": "02b8d786247046e98667cce4ca5af74e206f1793bf56f7a518ceb7fd",
    "Failures": 0,
    "Success": 4000
  }
]
```

These observations should be aggregated into the following 

```
[
  {
    "Operation": "BatchPutData",
    "AvailabilityZone": "us-east-2a",
    "Id": "d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f",
    "StartTime": 1767153321.144,
    "Counters": {
      "Failures": 423,
      "Success": 2800
    }
  },
  {
    "Operation": "BatchPutData",
    "AvailabilityZone": "us-east-2b",
[...]

PR #158, opened 2025-12-31
feat: Add support for `#[aggregate]`

rcoh:
📬 *Issue #, if available:*

✍️ *Description of changes:* This implements the `#[aggregate]` proc macro and associated traits.

```rust
#[aggregate]
#[metrics]
struct BackendCall {
    #[aggregate(strategy = Sum)]
    requests_made: u64,

    #[aggregate(strategy = Histogram<Duration, SortAndMerge>)]
    latency: Duration,

    #[aggregate(strategy = Sum)]
    errors: u64,

    #[aggregate(strategy = KeepLast)]
    error_message: Option<String>,
}

#[metrics(rename_all = "PascalCase", emf::dimension_sets = [["QueryId"]])]
struct DistributedQuery {
    query_id: String,
    #[metrics(flatten)]
    backend_calls: Aggregate<BackendCall>,
}
```

When you attach the `#[aggregate]` attribute macro, it generates the `AggregateStrategy` trait implementation for `T::Closed`:

```rust
pub trait AggregateStrategy: 'static {
    /// The source type being aggregated
    type Source: Merge;
    /// The key extraction strategy
    type Key: Key<Self::Source>;
}
```

### DIFF
--- a/metrique-macro/src/lib.rs
+++ b/metrique-macro/src/lib.rs
@@ -490,8 +490,7 @@ pub fn metrics(attr: TokenStream, input: proc_macro::TokenStream) -> proc_macro:
 /// ## Built-in Strategies
 ///
 /// - **`Sum`** - Sums numeric values together
-/// - **`Histogram<T>`** - Collects values into a distribution.
-///      Histogram has a second generic that can control how values are stored. See the `Histogram` docs for more info.
+/// - **`Histogram<T>`** - Collects values into a distribution. Histogram has a second generic that can control how values are stored. See the `Histogram` docs for more info.
 /// - **`KeepLast`** - Keeps the most recent value
 ///
 /// ```
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,9 +190,9 @@ dependencies = [
 
 [[package]]
 name = "clap_lex"
-version = "0.7.6"
+version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1d728cc89cf3aee9ff92b05e62b19ee65a02b5702cff7d5a377e32c6ae29d8d"
+checksum = "c3e64b0cc0439b12df2fa678eae89a1c56a529fd067a9115f7827f1fffd22b32"
 
 [[package]]
 name = "condtype"
@@ -546,9 +546,9 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "histogram"
-version = "0.11.3"
+version = "0.11.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95aebe0dec9a429e3207e5e34d97f2a7d1064d5ee6d8ed13ce0a26456de000ae"
+checksum = "e7e05f6bebaf5e4ae9ffac23c348518ceda3775c3b5edabdd95c3c0abee0c040"
 dependencies = [
  "thiserror 1.0.69",
 ]
@@ -1148,9 +1148,9 @@ dependencies = [
 
 [[package]]
 name = "rand_core"
-version = "0.9.3"
+version = "0.9.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "99d9a13982dcf210057a8a78572b2217b667c3beacbf3a0d8b454f6f82837d38"
+checksum = "76afc826de14238e6e8c374ddcc1fa19e374fd8dd986b0d2af0d02377261d83c"
 dependencies = [
  "getrandom",
 ]
@@ -1548,30 +1548,30 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.44"
+version = "0.3.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e7d9e3bb61134e77bde20dd4825b97c010155709965fedf0f49bb138e52a9d"
+checksum = "f9e442fc33d7fdb45aa9bfeb312c095964abdf596f7567261062b2a7107aaabd"
 dependencies = [
  "deranged",
  "itoa",
  "num-conv",
  "powerfmt",
- "serde",
+ "serde_core",
  "time-core",
  "time-macros",
 ]
 
 [[package]]
 name = "time-core"
-version = "0.1.6"
+version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40868e7c1d2f0b8d73e4a8c7f0ff63af4f6d19be117e90bd73eb1d62cf831c6b"
+checksum = "8b36ee98fd31ec7426d599183e8fe26932a8dc1fb76ddb6214d05493377d34ca"
 
 [[package]]
 name = "time-macros"
-version = "0.2.24"
+version = "0.2.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "30cfb0125f12d9c277f35663a0a33f8c30190f4e4574868a330595412d34ebf3"
+checksum = "71e552d1249bf61ac2a52db88179fd0673def1e1ad8243a00d9ec9ed71fee3dd"
 dependencies = [
  "num-conv",
  "time-core",
@@ -2146,6 +2146,6 @@ dependencies = [
 
 [[package]]
 name = "zmij"
-version = "1.0.12"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fc5a66a20078bf1251bde995aa2fdcc4b800c70b5d92dd2c62abc5c60f679f8"
+checksum = "bd8f3f50b848df28f887acb68e41201b5aea6bc8a8dacc00fb40635ff9a72fea"


metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
-## Aggregation strategies
+### Field-level aggregation
 
-By default, histograms use `ExponentialAggregationStrategy`. To use a different strategy, specify it as the second type parameter:
+Individual fields use aggregation strategies that implement `AggregateValue<T>`:
 
-```rust
-use metrique_aggregation::histogram::{Histogram, SortAndMerge};
-use std::time::Duration;
+- **`Counter`** - Sums values together
```
arielb1: Could you call it `Sum`?

metrique-aggregation/src/keyed_sink.rs
```diff
@@ -0,0 +1,115 @@
...
+    marker::PhantomData,
+    sync::{
+        Arc,
+        mpsc::{Sender, channel},
+    },
+    thread,
+    time::Duration,
+};
+
+/// Helper that "Roots" an inflectable entry (temporary copy from metrique, needs to move to core)
```
arielb1: Why can't metrique-aggregation depend on metrique?

I think it should either depend on metrique (if it's a "user"), or be a part of metrique.

Eventually we should move the traits to `metrique-core` (or to a `metrique-aggregation-core` crate if you'll prefer).

metrique-aggregation/src/traits.rs
```diff
@@ -0,0 +1,183 @@
...
+pub trait AggregateValue<T> {
+    /// The accumulated type (often same as T, but can differ for histograms).
+    type Aggregated;
+
+    /// Aggregate a value into the accumulator.
+    fn add_value(accum: &mut Self::Aggregated, value: T);
+}
+
+/// Trait for implementing AggregateValue when references are possible
+pub trait AggregateValueRef<T>: AggregateValue<T> {
```
arielb1: Should this trait be an alias like `CloseValueRef`?

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,511 @@
...
+        .clone()
+        .ok_or_else(|| Error::new(field.span(), "aggregate only supports named fields"))?;
+
+    let mut strategy = None;
+    let mut is_key = false;
+
+    for attr in &field.attrs {
+        if attr.path().is_ident("aggregate") {
+            attr.parse_nested_meta(|meta| {
+                if meta.path.is_ident("strategy") {
```
arielb1: Could you parse the attributes automatically like metrique does?

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,511 @@
...
+
+                fn key(source: &Self::Source) -> Self::Key {
+                    #[allow(deprecated)]
+                    #key_expr
+                }
+            }
+        })
+    } else {
+        Ok(quote! {
+            impl metrique_aggregation::__macro_plumbing::MergeOnDropExt for #original_name {}
```
arielb1: Non-standard for an ext trait to be implemented manually. I feel like a macro-generated inherent impl is better, since you avoid having to implement the trait.

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+use metrique_writer::unit::{Millisecond, Byte};
 use std::time::Duration;
 
-#[metrics(rename_all = "PascalCase")]
-struct QueryMetrics {
-    query_id: String,
-    
+#[aggregate]
+#[metrics]
+struct ApiCall {
```
jlizen: Minor: maybe rename to something like `ApiCallAttempt` and use it as an opportunity to demonstrate `last one wins` via a `success: bool`, `status_code: &'static str`, and `error_msg: Option<String>` field (along with preserving latency, response size, etc).

Fine to defer if that is too verbose for this, but i think that example might belong somewhere.

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
 
-By default, histograms use `ExponentialAggregationStrategy`. To use a different strategy, specify it as the second type parameter:
+Individual fields use aggregation strategies that implement `AggregateValue<T>`:
 
-```rust
-use metrique_aggregation::histogram::{Histogram, SortAndMerge};
-use std::time::Duration;
+- **`Sum`** - Sums values together
+- **`Histogram<T>`** - Collects values into a distribution
+- **`LastValueWins`** - Keeps the most recent value
```
jlizen: How about `FirstValueWins`? Fine if that is scope creep but might be worth a backlog for improvements in mind?

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+    request_id: String,
+}
 ```
 
-Available strategies:
+### Thread-safe aggregation with `MutexAggregator<T>`
+
+Use `MutexAggregator<T>` when you need to aggregate from multiple threads or use `merge_and_close_on_drop`:
+
+```rust,ignore
```
jlizen: same feedback on ignore

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+metrics.api_calls.add(ApiCall {
+    latency: Duration::from_millis(45),
+    response_size: 1024,
+});
+metrics.api_calls.add(ApiCall {
+    latency: Duration::from_millis(67),
+    response_size: 2048,
+});
+
+// When metrics drops, emits a single entry with aggregated values
```
jlizen: would be pretty nice to give a sense of what the output might look like, especially the histogram

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+    response_size: 50,
+}.merge_and_close_on_drop(&metrics.api_calls);
+
+// Modify the call before it's closed and merged
+call.latency.stop();
+call.response_size = 75;
+// Automatically closed and merged when guard drops
+```
+
+### Keyed aggregation with `KeyedAggregationSink`
```
jlizen: A big piece that is missing here is the "so what" with the keys. It looks like here, the way this would be used would be, emit a dimensionset based on endpoint so that you can access the latency value per key separately?

I think we should spell something like that out, even if we don't go as far as giving a real emf dimensionset in the example.

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+    endpoint: "GetItem".to_string(),
+    latency: Duration::from_millis(10),
+});
+```
+
+## Histogram strategies
+
+Histograms support different bucketing strategies:
+
+- **`ExponentialAggregationStrategy`** (default) - Exponential bucketing with ~6.25% error
```
jlizen: Future improvement here would presumably be a knob to adjust precision? (which we wait for anybody to ask for?)

metrique-aggregation/src/traits.rs
```diff
@@ -0,0 +1,183 @@
...
+    type Aggregated: Send + CloseEntry;
+
+    /// Aggregation Key. For structs with no key, you typically use `()`
+    type Key: Send + Hash + Eq;
+
+    /// Merge a given entry into the Aggregate
+    fn merge_entry(accum: &mut Self::Aggregated, entry: Self::Source);
+
+    /// Create a new, empty, aggregated entry for a given key
+    fn new_aggregated(key: &Self::Key) -> Self::Aggregated;
```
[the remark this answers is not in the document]
rcoh: yeah I am messing around with options here. I think I will have something like `Key<'a>` and then a way to convert it to `Key<'static>`

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+- **`SortAndMerge`** - Stores all observations exactly for perfect precision
+- **`AtomicExponentialAggregationStrategy`** - Thread-safe exponential bucketing for `SharedHistogram`
+
+```rust,ignore
```
jlizen: this example is to show strategy selection? I'd suggest putting it in a minimal #[aggregate] entry rather than this stub that wouldn't compile (and remove the ignore)

jlizen (commented):
Leaving some preliminary feedback on docs before digging into impl

metrique-macro/src/aggregate.rs
rcoh: there is a bug (?) in darling where it can't parse stuff like `x = Histogram<Duration, Blah>` 

since the macro is still simple I didn't try that hard

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    #[metrics(flatten)]
+    api_calls: Aggregate<ApiCall>,
+    request_id: String,
+}
 ```
 
-Available strategies:
+### Thread-safe aggregation with `MutexAggregator<T>`
```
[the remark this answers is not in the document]
rcoh: I was imagining we probably have one that is queue based and one that is mutex based and the trade off is pick:

1. mutex
2. storage (a channel that gets read on close)
3. having an additional thread

And that callers would have a better sense about what they wanted

metrique-aggregation/README.md
rcoh: yeah exactly.

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+
+// Modify the call before it's closed and merged
+call.latency.stop();
+call.response_size = 75;
+// Automatically closed and merged when guard drops
+```
+
+### Keyed aggregation with `KeyedAggregationSink`
+
+Use `KeyedAggregationSink` to aggregate by key with time-based flushing:
```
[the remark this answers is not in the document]
rcoh: what do you mean by 2 keys? Oh dual aggregation? that isn't possible right now because I haven't wired up `AggregateEntryRef` into the sink

metrique-aggregation/src/lib.rs
```diff
@@ -5,3 +5,16 @@
...
 pub mod histogram;
+pub mod keyed_sink;
+pub mod sink;
+pub mod traits;
+pub mod value;
+
+#[doc(hidden)]
+pub mod __macro_plumbing {
+    pub use crate::sink::MergeOnDropExt;
+    pub use crate::traits::{AggregateEntry, AggregateEntryRef, AggregateValue};
```
jlizen: I guess our plan is to experiment with these more before moving into `*-core`?

metrique-aggregation/src/lib.rs
```diff
@@ -5,3 +5,16 @@
...
+    pub use crate::value::IfYouSeeThisUseAggregateOwned;
```
jlizen: Meta: This is a nice pattern, is there a way to do similar for `subfield_owned`? I recently added an unconditional tweak to the missing CloseValue diagnostic, but it's not ideal.

metrique-aggregation/src/sink.rs
```diff
@@ -0,0 +1,204 @@
...
+pub trait AggregateSink<T: AggregateEntry> {
+    /// Merge a given item into the sink
+    fn merge(&self, entry: T::Source);
+}
+
+/// Sink that aggregates a single type of entry backed by a mutex
+///
+/// Compared to [`Aggregate`], this type allows appending with `&T` so it supports
+/// using [`MergeOnDropExt::merge_on_drop`]
+pub struct MutexAggregator<T: AggregateEntry> {
```
jlizen: Same feedback on `Mutex*` naming -> something about implementation

metrique-aggregation/src/sink.rs
```diff
@@ -0,0 +1,204 @@
...
+    aggregator: Arc<Mutex<Option<T::Aggregated>>>,
```
jlizen: Is there a case where people would want to use this with &self even without the arc? For instance, with task-local concurrency? I think probably yes... Should we make the caller wrap it with their own arc instead?

metrique-aggregation/README.md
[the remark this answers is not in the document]
arielb1: A key that is a tuple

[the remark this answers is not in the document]
jlizen: I meant two distinct keys. Yes you could accomplish that as a tuple for now. But I'm fine to sit tight on this until we have `AggregateEntryRef`. 

I mention since that seems like a requirement for: https://github.com/awslabs/metrique/issues/155

jlizen (commented):
Overall impl looks solid to me, nothing hugely blocking, though I agree with a lot of ariel's feedback

metrique-aggregation/src/traits.rs
rcoh: trait is actually unused, deleted.

metrique-aggregation/src/keyed_sink.rs
rcoh: ah actually...it needs to. It uses the `#[metrics]` macro in its expansion. So a caller would need it. Probably reasonable to depend on it

metrique-aggregation/README.md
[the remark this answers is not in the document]
jlizen: Agree that capturing that split is good.

How about:
- `SharedAggregator` (currently backed by a mutex)
- `BufferedAggregator` (backed by eg a channel, not yet implemented)
- `WorkerAggregator` (backed by a thread/task, not yet implemented)

metrique-aggregation/src/lib.rs
jlizen: Opened: https://github.com/awslabs/metrique/issues/164

metrique-aggregation/AGENTS.md
```diff
@@ -0,0 +1,2 @@
+# Working on the macro
+The macro is define in ../metrique-macro
```
jlizen: nit: defined

Would it be worth mentioning where the insta and try_assert tests are and how to run them, also? (As well as general integ tests?)

metrique-aggregation/Cargo.toml
```diff
@@ -7,19 +7,30 @@ license = "Apache-2.0"
...
-metrique = { path = "../metrique", features = ["test-util"] }
+metrique = { path = "../metrique", features = ["test-util", "emf"] }
 divan = "0.1"
 rand.workspace = true
 rand_chacha.workspace = true
 assert2 = "0.3"
 rstest = "0.23"
+trybuild.workspace = true
+rustversion.workspace = true
+tokio = { version = "1", features = ["full"] }
```
jlizen: nit: could also be workspace dependency

metrique-aggregation/README.md
```diff
@@ -1,74 +1,168 @@
...
+struct ApiCall {
+    // Aggregate latency with a histogram
+    #[aggregate(strategy = Histogram<Duration>)]
     #[metrics(unit = Millisecond)]
-    backend_latency: Histogram<Duration>,
+    latency: Duration,
+    
+    #[aggregate(strategy = Sum)]
+    #[metrics(unit = Byte)]
+    response_size: usize,
```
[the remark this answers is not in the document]
rcoh: did this in some of the actual examples

metrique-aggregation/README.md
```diff
@@ -1,74 +1,394 @@
...
 
-Use histograms when you have many observations of the same metric within a single unit of work. For example:
+- **High-frequency, low-level events**: TLS handshakes, storage operations, or other infrastructure-level metrics
+- **Fan-out operations**: A single unit of work spawns multiple sub-operations you want to aggregate
+- **Background processing**: Queue workers that generate one metric per processed item at an extremely high rate
 
-- A distributed query that fans out to multiple backend services
-- Processing a batch of items where you want to track per-item latency
-- Any operation that generates multiple measurements you want to aggregate
+Typically, aggregation should only be used when sampling is not sufficient, however, metrique can help you achieve both: Since metrique has access to the raw entries during aggregation, you can emit a sampled set of full events (e.g. with congressional sampling to ensure all errors are precisely recorded) to one sink while aggregating data to another sink.
```
jlizen: > Typically, aggregation should only be used when sampling is not sufficient

This is a bit eyebrow raising. Perhaps:

> Standalone aggregation should be relied on with caution. It strips useful per-entry metadata, which can help you understand _why_ metrics look a certain way. But, `metrique` allows you to have the best of both worlds, by both aggregating and sampling the same entry. Since metrique has access...

metrique-aggregation/README.md
```diff
@@ -1,74 +1,394 @@
...
+
+- **[`Sum`]** - Sums values together (use for counts, totals)
+- **[`Histogram<T>`]** - Collects values into a distribution (use for latency, sizes)
+- **[`KeepLast`]** - Keeps the most recent value (use for gauges, current state)
+
+### Entry-Level Aggregation
+
+The [`aggregate`] macro generates implementations that define how complete entries are combined. It creates the merge logic, key extraction, and aggregation strategy for your type.
+
+### Keys
```
jlizen: Is it worth mentioning that keys are only relevant for entry-level aggregation? That's true right?

metrique-aggregation/README.md
```diff
@@ -1,74 +1,394 @@
...
+[`KeyedAggregator`].
+
+```rust
+use metrique::unit_of_work::metrics;
+use metrique_aggregation::{aggregate, histogram::Histogram};
+use std::time::Duration;
+
+#[aggregate]
+#[metrics]
+struct ApiCall {
```
jlizen: What is the behavior if there are non-aggregated fields present. For example, some other String field?

Are they just ignored? Is that worth mentioning in the example? (`// ignored because it has no #[aggregate(key)] statement`

metrique-aggregation/examples/sink_level.rs
```diff
@@ -0,0 +1,179 @@
...
+            "Processed item #{}: type={}, priority={}, time={}ms, result={}",
+            total_processed,
+            item.item_type,
+            item.priority,
+            processing_time.as_millis(),
+            if result.is_ok() { "OK" } else { "ERROR" }
+        );
+
+        // Send metrics to the aggregating sink
+        QueueItem {
```
jlizen: Is there a reason we are showing the manual `close_and_merge` instead of `MergeOnDrop`? I would have assumed that would be the standard usage.

metrique-aggregation/examples/embedded.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        println!(
+            "  {} - {}ms - {}",
+            shard,
+            latency.as_millis(),
+            if result.is_ok() { "OK" } else { "ERROR" }
+        );
+
+        // Insert each backend call into the aggregator
+        metrics.backend_calls.insert(BackendCall {
```
jlizen: You could also use `MergeOnDrop` here right? Would it be worth including that example usage too?

metrique-aggregation/README.md
```diff
@@ -1,74 +1,395 @@
...
+// Can be shared across threads
+let h = histogram.clone();
+std::thread::spawn(move || {
+    h.add_value(42);
+});
+# }
 ```
 
-Available strategies:
+See the `histogram` example for more usage patterns.
```
jlizen: This example has a shared histogram but doesn't seem to do anything with it yet?

metrique-aggregation/README.md
```diff
@@ -1,74 +1,395 @@
...
+
+Individual fields use aggregation strategies that implement [`AggregateValue<T>`]:
+
+- **[`Sum`]** - Sums values together (use for counts, totals)
+- **[`Histogram<T>`]** - Collects values into a distribution (use for latency, sizes)
+- **[`KeepLast`]** - Keeps the most recent value (use for gauges, current state)
+
+### Entry-Level Aggregation
+
+The [`aggregate`] macro generates implementations that define how complete entries are combined. It creates the merge logic, key extraction, and aggregation strategy for your type.
```
jlizen: Is it worth mentioning that you can use this both with sink-level aggregates, and also as subfields on individual entries?

metrique-aggregation/README.md
```diff
@@ -1,74 +1,395 @@
...
+- **[`MergeRef`]** - Like [`Merge`], but merges by reference (enables [`SplitSink`] to send to multiple destinations)
+- **[`Key`]** - Extracts grouping keys from entries to determine which entries should be merged
+- **[`AggregateStrategy`]** - Ties together the source type, merge behavior, and key extraction
+- **[`AggregateSink<T>`]** - Destination that accepts and aggregates entries
+
+The [`aggregate`] macro generates implementations of these traits for your type. For most use cases, you don't need to implement these manually - the macro handles it.
+
+## When NOT to Use Aggregation
+
+- **Request/response services**: Use sampling instead
```
jlizen: I'm not sure I agree with this, I think "use both" is a fine recommendation

metrique-aggregation/README.md
```diff
@@ -1,74 +1,395 @@
...
+- **Low-frequency events**: Aggregation overhead isn't worth it
+- **Need individual event details**: Aggregation loses individual event context
+- **Simple counting**: Basic counters don't need aggregation complexity
+
+## Examples
```
jlizen: I might move this higher up

metrique-aggregation/src/aggregator.rs
```diff
@@ -0,0 +1,255 @@
+//! Aggregation structures for collecting and merging entries
+
+use metrique_core::CloseValue;
+use std::marker::PhantomData;
+
+use metrique_writer::BoxEntrySink;
```
jlizen: available via `metrique::writer`

metrique-aggregation/examples/split.rs
```diff
@@ -0,0 +1,173 @@
+//! Example: Split Aggregation Pattern
+//!
+//! This example demonstrates using `SplitSink` to send the same data to multiple
+//! destinations. We aggregate metrics for precise counts while also emitting raw
+//! events for debugging and tracing.
+
+use metrique::test_util::test_entry_sink;
+use metrique::unit_of_work::metrics;
+use metrique_aggregation::histogram::{Histogram, SortAndMerge};
```
jlizen: Any plans to re-export `metrique::agreggation::*`? (with eg `aggregation` feature)

metrique-macro/src/lib.rs
```diff
@@ -403,6 +404,177 @@ pub fn metrics(attr: TokenStream, input: proc_macro::TokenStream) -> proc_macro:
...
+///
+/// ## Raw Mode
+///
+/// Use `#[aggregate(raw)]` to aggregate on the raw struct before closing. In raw mode:
+/// - Aggregation strategies receive the raw field type before `CloseValue` is applied
+/// - Use this if the base struct is not also a metric
+///
+/// # Aggregation Keys
+///
+/// Fields marked with `#[aggregate(key)]` define the aggregation key. Observations with different
```
jlizen: Same question, what is the relationship of keys to field-level aggregates (where the struct is a subfield)?

jlizen: It might be good to throw a compiler error if using entries with key as a subfield, just to avoid confusing semantics... not sure, maybe that's too annoying if you want to use both in embedded + sink-level aggregates...

metrique-macro/src/lib.rs
```diff
@@ -403,6 +404,177 @@ pub fn metrics(attr: TokenStream, input: proc_macro::TokenStream) -> proc_macro:
...
+///     latency: Duration,
+///
+///     #[aggregate(strategy = Sum)]
+///     bytes_sent: usize,
+/// }
+/// ```
+///
+/// # Generated Types
+///
+/// For a struct named `MyMetrics`, the macro generates:
```
jlizen: > ... the `#[aggregate]` macro generates...

It's a bit confusing since right before this we show both `#[aggregate]` and `#[metrics]` on the same block

jlizen (commented):
More doc feedback

metrique-aggregation/src/sink/mutex.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use std::sync::{Arc, Mutex};
+
+use metrique_core::CloseValue;
+
+use crate::traits::{AggregateSink, RootSink};
+
+/// Sink that aggregates a single type of entry backed by a mutex
+///
+/// Compared to [`crate::aggregator::Aggregate`], this type allows appending with `&T` so it supports
```
arielb1: You mean, with `&Self` (&T sounds like `&Self::Source`)?

Also, add an example.

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+        let strategy = f.strategy.as_ref().unwrap();
+        let source_ty = &f.ty;
+        let value_ty = if entry_mode {
+            quote! { <#source_ty as metrique::CloseValue>::Closed }
+        } else {
+            quote! { #source_ty }
+        };
+        quote! {
+            #(#metrics_attrs)*
+            #name: <#strategy as metrique_aggregation::__macro_plumbing::AggregateValue<#value_ty>>::Aggregated
```
arielb1: I think it should be `::metrique_aggregation` (and in all other places) for proper resolution [of course, in a truely-ideal world, Rust would have "proc macro, code" pairs, but in today's world `::crate` is the best you can do].

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+            .ok_or_else(|| Error::new(field.span(), "aggregate only supports named fields"))?;
+
+        let mut strategy = None;
+        let mut is_key = false;
+        let mut use_clone = false;
+
+        for attr in &field.attrs {
+            if attr.path().is_ident("aggregate") {
+                attr.parse_nested_meta(|meta| {
+                    if meta.path.is_ident("strategy") {
```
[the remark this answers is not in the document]
arielb1: Or garbage meta items of various sorts

metrique-aggregation/src/traits.rs
```diff
@@ -0,0 +1,371 @@
...
+    fn from_source(source: &Source) -> Self::Key<'_>;
+    /// Convert borrowed key to static lifetime
+    fn static_key<'a>(key: &Self::Key<'a>) -> Self::Key<'static>;
+
+    /// Compare the static and reference versions of the key for equality.
+    ///
+    /// Why not use Equivalent or another trait?
+    /// - Equivalent has a blanket impl that prevents it from being used
+    /// - You can implement this by doing `owned == borrowed` on your concrete type, however the bounds seem hard to express
+    fn static_key_matches<'a>(owned: &Self::Key<'static>, borrowed: &Self::Key<'a>) -> bool;
```
arielb1: Maybe you can use `type Key<'a>: ... + PartialEq<Self::Key<'static>>`? Or will that cause errors?

metrique-aggregation/README.md
arielb1: You can use `KeyedAggregator` within an entry. Tho it will only work if you enable split entry mode / a split entry supporting format. Worth noting.

metrique-macro/Cargo.toml
```diff
@@ -22,6 +22,7 @@ darling = { workspace = true }
 insta = { workspace = true }
 prettyplease = { workspace = true }
 metrique = { path = "../metrique", features = ["test-util"] }
+metrique-aggregation = { path = "../metrique-aggregation" }
```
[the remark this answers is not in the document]
arielb1: metrique-macro is not supposed to be used directly by customers anyway.

metrique-aggregation/examples/embedded.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use metrique::DefaultSink;
+use metrique::emf::Emf;
+use metrique::unit_of_work::metrics;
+use metrique::writer::{FormatExt, sink::FlushImmediatelyBuilder};
+use metrique_aggregation::aggregator::Aggregate;
+use metrique_aggregation::histogram::Histogram;
+use metrique_aggregation::value::{KeepLast, Sum};
+use metrique_aggregation::{aggregate, histogram::SortAndMerge};
+use metrique_writer::BoxEntrySink;
```
jlizen: available via `metrique::writer` (also Millisecond)

metrique-aggregation/examples/split.rs
```diff
@@ -0,0 +1,173 @@
...
+use metrique_aggregation::sink::{RawSink, SplitSink};
+use metrique_aggregation::{KeyedAggregator, WorkerSink, aggregate};
+use metrique_writer::unit::Millisecond;
```
jlizen: available via `metrique::writer`

metrique-writer/Cargo.toml
```diff
@@ -29,6 +29,7 @@ metrics-util_020 = { workspace = true, optional = true }
 metrique-writer-core = { path = "../metrique-writer-core", version = "0.1.9" }
 metrique-writer-macro = { path = "../metrique-writer-macro", version = "0.1.6" }
 metrique-core = { path = "../metrique-core", version = "0.1.11" }
+ordered-float = { version = "5.1.0", optional = true }
```
jlizen: Could be workspace dep?

metrique-writer/src/test_util.rs
```diff
@@ -173,6 +178,36 @@ impl Metric {
...
+        let mut out = Vec::with_capacity(self.num_observations() as usize);
+        self.distribution.iter().for_each(|obs| match obs {
+            Observation::Unsigned(v) => out.push(*v as f64),
+            Observation::Floating(v) => out.push(*v),
+            Observation::Repeated { occurrences, total } => {
+                for _ in 0..*occurrences {
+                    out.push(total / *occurrences as f64)
+                }
+            }
+            _ => unreachable!(),
```
jlizen: nit: prefer exhaustive match

metrique-writer/src/test_util.rs
```diff
@@ -173,6 +178,36 @@ impl Metric {
...
+
+    /// Returns all observations in a sorted Vec of f64, flatten repeated obsevations
+    pub fn flatten_and_sort(&self) -> Vec<f64> {
+        let mut out = Vec::with_capacity(self.num_observations() as usize);
+        self.distribution.iter().for_each(|obs| match obs {
+            Observation::Unsigned(v) => out.push(*v as f64),
+            Observation::Floating(v) => out.push(*v),
+            Observation::Repeated { occurrences, total } => {
+                for _ in 0..*occurrences {
+                    out.push(total / *occurrences as f64)
```
jlizen: nit: slightly more readable to calculate the per-occurrence value once before the loop

metrique-macro/src/lib.rs
```diff
@@ -283,6 +284,164 @@ pub fn metrics(attr: TokenStream, input: proc_macro::TokenStream) -> proc_macro:
...
+/// This macro enables combining multiple observations of the same metric into a single aggregated
+/// entry. It must be placed before `#[metrics]` and generates:
+/// - An `Aggregated{StructName}` struct with aggregated field types
+/// - An `AggregateEntry` trait implementation for merging observations
+///
+/// # Container Attributes
+///
+/// | Attribute | Type | Description | Example |
+/// |-----------|------|-------------|---------|
+/// | `raw` | Flag | Aggregates on the raw struct instead of the closed entry (default: aggregates on closed entry) | `#[aggregate(raw)]` |
```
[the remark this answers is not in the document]
jlizen: I don't mind `AggregateNoClose` i guess

metrique-aggregation/src/aggregator.rs
```diff
@@ -0,0 +1,255 @@
...
+
+use crate::traits::{
+    AggregateSink, AggregateSinkRef, AggregateStrategy, AggregateTy, FlushableSink, Key, KeyTy,
+    Merge, MergeRef,
+};
+use crate::value::NoKey;
+
+/// The Entry type you have when merging entries
```
jlizen: This is specifically for KeyedAggregator` no?

arielb1:
Reasons against "raw":

1. When reading, it's vague. What does `#[aggregate(raw)]` do? It could do anything. It's easier to remember what `#[aggregate(no_closed)]` does.
2. Positive superstitions - people can end up believing that `#[aggregate(raw)]` does things other than merely disabling closing. 
3. Negative superstitions - reviewers believing that "raw" means danger, something with disabling validations or compatibility checks and "should not be used". Tho `no_closed` feels a bit scary as well, tho less scary than "raw".

Tho `no_closed` feels overly complicated if you are not into the metrique ecosystem. Maybe a word like `direct` would work better? `#[aggregate(direct)]` feels OK. Is there a better name?

metrique-aggregation/src/aggregator.rs
```diff
@@ -0,0 +1,255 @@
...
+pub type AggregatedEntry<T> = crate::traits::AggregationResult<
+    <<<T as AggregateStrategy>::Key as Key<<T as AggregateStrategy>::Source>>::Key<'static> as CloseValue>::Closed,
+    <<<T as AggregateStrategy>::Source as Merge>::Merged as CloseValue>::Closed,
+>;
+
+/// Keyed aggregator that uses a HashMap to aggregate entries by key
+///
+/// This is the core aggregation logic without any threading or channel concerns.
```
jlizen: > This is the core aggregation logic without any threading or channel concerns

I would phrase this as, `This will generally be used to wrap an inner aggregator that handles threading or channel concerns (such as `MutexSink` or `WorkerSink`)?

metrique-aggregation/src/sink/worker.rs
```diff
@@ -0,0 +1,90 @@
...
+impl<T, Inner> WorkerSink<T, Inner>
+where
+    T: Send + 'static,
+    Inner: AggregateSink<T> + FlushableSink + Send + 'static,
+{
+    /// Create a new background thread sink
+    pub fn new(mut inner: Inner, flush_interval: Duration) -> Self {
+        let (sender, receiver) = channel();
+
+        let handle = thread::spawn(move || {
```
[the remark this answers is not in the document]
jlizen: Currently we're mostly runtime agnostic no? We don't have have `tokio::rt` as a dependency here, `metrique`, or `metrique-writer`?

For people that want lightweight, ephemeral sinks, I assume they want `MutexSink` anyway?

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+        }
+    };
+
+    let mut parsed_fields = Vec::new();
+
+    for field in fields {
+        let name = field
+            .ident
+            .clone()
+            .ok_or_else(|| Error::new(field.span(), "aggregate only supports named fields"))?;
```
jlizen: nit: just tag it `unreachable()!`, we guard against this immediately above

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+                        use_clone = true;
+                        Ok(())
+                    } else {
+                        Err(meta.error("unknown aggregate attribute"))
+                    }
+                })?;
+            }
+        }
+
+        if !is_key && strategy.is_none() {
```
jlizen: (plug for ignore)

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+        .iter()
+        .find(|attr| attr.path().is_ident("metrics"));
+
+    let derive_default = quote! { #[derive(Default)] };
+
+    Ok(quote! {
+        #metrics_attr
+        #derive_default
+        #[allow(clippy::type_complexity)]
+        // aggregated needs to be pub because it is used in a trait
```
jlizen: Is this true long term? Or just because we are hoisting other things to be `pub` visibility and you get lint issues?

metrique-aggregation/tests/aggregation.rs
```diff
@@ -0,0 +1,278 @@
+//! Test using the #[aggregate] macro
```
jlizen: I'm not seeing any tests with prefix, might be prudent to throw in?

I know we are just deferring to `#[metrics]` macro to handle those, but who knows what future edge case it might trip.

metrique-macro/src/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+
+#[derive(Debug)]
+struct ParsedAggregate {
+    fields: Vec<AggregateField>,
+}
+
+fn parse_aggregate_fields(input: &DeriveInput) -> Result<ParsedAggregate> {
+    let data_struct = match &input.data {
+        Data::Struct(s) => s,
+        _ => return Err(Error::new(input.span(), "aggregate only supports structs")),
```
jlizen: If you're anyway cutting follow-up issues, mind include enums in the list?

metrique-writer/src/test_util.rs
```diff
@@ -173,6 +178,36 @@ impl Metric {
...
+
+    /// Returns the total number of observations, correctly accounting for `Repeated`
+    pub fn num_observations(&self) -> u64 {
+        self.distribution
+            .iter()
+            .map(|obs| match obs {
+                Observation::Unsigned(_) => 1,
+                Observation::Floating(_) => 1,
+                Observation::Repeated { occurrences, .. } => *occurrences,
+                _ => unreachable!(),
```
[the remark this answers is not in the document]
arielb1: This is because of `non_exhaustive` - https://docs.rs/metrique-writer/latest/metrique_writer/enum.Observation.html - tho the unreachable should state that

metrique-aggregation/src/sink/mutex.rs
```diff
@@ -0,0 +1,59 @@
...
+where
+    Inner: AggregateSink<T>,
+{
+    fn merge(&self, entry: T) {
+        self.inner.lock().unwrap().merge(entry);
+    }
+}
+
+impl<Inner> CloseValue for MutexSink<Inner>
+where
```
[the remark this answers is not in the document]
rcoh: yeah there is some nuance for you how you close the value. You can either leave a `None` behind then sink becomes basically "dead" or you need someway of replacing the value. the aggregated value will basically always be default

metrique-aggregation/README.md
```diff
@@ -65,113 +81,19 @@ metrics.api_calls.insert(ApiCall {
...
-```
-
-Calls to the same endpoint will be aggregated together, while different endpoints remain separate.
-
-## Usage Patterns
-
-### Embedded Aggregation
-
-Use [`Aggregate<T>`] as a field in your metrics struct when a single unit of work fans out to multiple sub-operations:
+[`WokerSink`] performs aggregation in a background thread that periodically flushes aggregated data to a backing sink. [`MutexSink`] is alternative sink that manages concurrency with a mutex instead of a channel.
```
jlizen: This is an improvement, but it still is missing the "but which should I use"? Fine if we don't feel comfortable stating that yet.

metrique-aggregation/examples/split.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // Flush both sinks
+    println!("\nFlushing metrics...");
+    sink.flush().await;
+}
+
+#[tokio::main]
+async fn main() {
+    // Initialize tracing to see validation errors
```
jlizen: Do we still have any validation errors if we set `skip_all_validations(true)`?

Another option would be to just set `allow_ignored_dimensions(true)`:
https://docs.rs/metrique/latest/metrique/emf/struct.EmfBuilder.html#method.allow_ignored_dimensions

metrique-aggregation/examples/split.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl AggregateStrategy for AggregateByErrorsEndpoint {
+    type Source = ApiCallEntry;
+
+    type Key = AggregateByErrorsEndpoint;
+}
+
+#[derive(Debug, Hash, Clone, PartialEq, Eq)]
+#[metrics]
+pub struct ByErrorsEndpoint<'a> {
```
jlizen: Thanks for adding this to the example. Definitely a bit cumbersome/verbose for what is probably going to be a fairly common use case... Anyway we can iterate on it, the bones seem fine.

metrique-aggregation/examples/split.rs
```diff
@@ -0,0 +1,184 @@
...
+    // Create aggregator for precise metrics
+    let aggregate_by_endpoint = KeyedAggregator::<ApiCall>::new(ServiceMetrics::sink());
+    // also aggregate by errors
+    let aggregate_by_endoint_errors =
+        KeyedAggregator::<AggregateByErrorsEndpoint>::new(ServiceMetrics::sink());
+    // finally, we will also send the raw records to the sink as well
+    // in a more real use case, you might send this to another destiation
+    // (e.g. one that had a different set of dimensions atached), used sampling
+    // or was a compressed record for later inspection (perhaps it doesn't go to your metrics store)
+    let raw_sink = ServiceMetrics::sink();
```
jlizen: This is probably worth showing sampling on, it makes the example dramatically more realistic.

Though I can live with deferring it to: https://github.com/awslabs/metrique/issues/161

metrique-aggregation/README.md
```diff
@@ -1,74 +1,355 @@
...
+
+# fn main() {
+# use metrique::test_util::test_entry_sink;
+# let aggregated_sink = test_entry_sink().sink;
+# let raw_events_sink = test_entry_sink().sink;
+// Aggregator for precise counts
+let aggregator = KeyedAggregator::<QueueItem>::new(aggregated_sink);
+
+// Raw sink for sampling individual events
+let raw = EntrySinkAsAggregateSink::new(raw_events_sink);
```
jlizen: I think we really ought to include sampling in this example for realism.

metrique-aggregation/README.md
```diff
@@ -1,74 +1,355 @@
...
+# fn main() {
+let mut metrics = Metrics {
+    latency: Histogram::default(),
+};
+
+metrics.latency.add_value(Duration::from_millis(10));
+metrics.latency.add_value(Duration::from_millis(20));
+metrics.latency.add_value(Duration::from_millis(15));
+
+// write `metrics` to a sink to persist it...
```
jlizen: Why not just show that in the example? It's an extra twoish lines and reduces cognitive load maybe?

metrique-macro/src/lib.rs
jlizen: (This is done + tested)

metrique-macro/src/lib.rs
```diff
@@ -403,6 +404,177 @@ pub fn metrics(attr: TokenStream, input: proc_macro::TokenStream) -> proc_macro:
...
+/// - Multiple key fields create a tuple key: `(Field1, Field2, ...)`
+/// - Without key fields, all observations aggregate into a single entry
+/// - Key fields must implement `Clone`
+///
+/// # Aggregation Strategies
+///
+/// Each non-key field must specify an aggregation strategy that implements `AggregateValue<T>`:
+///
+/// ## Built-in Strategies
+///
```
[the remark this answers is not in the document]
jlizen: (Same question, do we need to mention histogram strategies?)

metrique-aggregation/README.md
```diff
@@ -1,74 +1,355 @@
...
+}
+# fn main() {}
+```
+
+**Output**: Multiple aggregated entries like `ItemType: "email", Priority: 1, ItemsProcessed: 1247, ProcessingTime: [histogram]`
+
+**Choosing between WorkerSink and MutexSink:**
+
+- **[`MutexSink`]** - Use when you have inputs from a smaller number of threads. Great for supporting `close_and_merge` with embedded metrics. Currently does not support automatic flushing.
+- **[`WorkerSink`]** - Use for sink-level aggregation from many producers across many threads. The channel-based design reduces contention and provides configurable flush timing.
```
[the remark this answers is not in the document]
rcoh: the sinks need a lot more configuration options

PR #169, opened 2026-01-08
feat: Add support for lifetimes with `#[metrics]`

rcoh:
📬 *Issue #, if available:*

✍️ *Description of changes:* In order to support aggregation, we need to support keys with lifetimes.

This adds supports for lifetimes to the `#[metrics]` macro. Currently, if you define a lifetime, we don't generate an `append_on_drop` wrapper. We could relax that in the future.

In order to make this work, I also need to improve our `CloseValue` impls. I was able to make them significantly more flexible than they are today which was a nice plus.

### Why do we need this?
In aggregation, we want to have key entries that have lifetimes:
```rust
#[metrics]
struct Key<'a> {
  f: Cow<'a, String>
}
```

This allows very efficient aggregation since we don't need to allocate a new key when the hashmap already has the key.

### DIFF
--- a/metrique-macro/src/enums.rs
+++ b/metrique-macro/src/enums.rs
@@ -254,7 +254,7 @@ pub(crate) fn generate_metrics_for_enum(
             &entry_name,
             &input.generics,
             variants,
-        ),
+        )?,
         _ => crate::entry_impl::generate_enum_entry_impl(
             &entry_name,
             &input.generics,
--- a/metrique-macro/src/value_impl.rs
+++ b/metrique-macro/src/value_impl.rs
@@ -9,7 +9,14 @@ pub(crate) fn generate_value_impl_for_enum(
     value_name: &Ident,
     generics: &syn::Generics,
     parsed_variants: &[MetricsVariant],
-) -> Ts2 {
+) -> Result<Ts2, syn::Error> {
+    if !generics.params.is_empty() {
+        return Err(syn::Error::new_spanned(
+            generics,
+            "generics and lifetimes are not supported for #[metrics(value)] enums",
+        ));
+    }
+
     let from_and_sample_group = crate::enums::generate_from_and_sample_group_for_enum(
         value_name,
         generics,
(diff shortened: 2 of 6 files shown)


metrique/tests/lifetimes.rs
```diff
@@ -0,0 +1,34 @@
+use assert2::check;
+use metrique::{test_util::test_metric, unit_of_work::metrics};
+use std::borrow::Cow;
+
+#[metrics]
+pub struct Foo<'a> {
```
jlizen: Probably good to cover all 4 types? (Entry enum/struct, Value enum/struct)

metrique-macro/src/structs.rs
```diff
@@ -141,18 +155,16 @@ fn generate_close_value_impls_for_struct(
...
-        quote! {
-            #[allow(deprecated)]
-            #entry {
-                #(#config,)*
-                #(#fields,)*
-            }
-        },
-    )
+
+    let impl_body = quote! {
```
jlizen: Nice cleanup

metrique-core/src/close_value_impls.rs
```diff
@@ -51,24 +51,37 @@ macro_rules! close_value {
...
-    f32,
-    f64,
-    u16,
-    u32,
-    u64,
-    u8,
-    usize,
-    &'static str,
-    SystemTime
+    bool, Duration, f32, f64, u16, u32, u64, u8, usize, SystemTime
```
jlizen: removing `&'static str` because we are manually implementing CloseValueRef with broader lifetimes?

jlizen (approved):
Some nits

metrique-macro/src/enums.rs
```diff
@@ -262,25 +270,38 @@ pub(crate) fn generate_metrics_for_enum(
...
-        MetricMode::RootEntry => {
+        MetricMode::RootEntry if !has_lifetimes(&input.generics) => {
             let on_drop_wrapper =
                 generate_on_drop_wrapper(vis, &guard_name, enum_name, &entry_name, &handle_name);
             quote! {
                 #on_drop_wrapper
             }
         }
-        MetricMode::Subfield
+        MetricMode::RootEntry
```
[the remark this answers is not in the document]
rcoh: (now uneeded)

Issue #103, opened 2025-09-12
Tokio task local time source override


TerminalWitchcraft:
Currently, `set_time_source` returns `ThreadLocalTimeSourceGuard` which cannot be sent between tokio threads since its not Send + Sync.

Maybe we can create a `set_time_source_tokio_local` version of this which is safe to send across await points.

This is useful for testing scenarios.

PR #193, opened 2026-01-27
Add `set_test_sink_on_runtime` to simplify testing with Tokio

rcoh:
✍️ *Description of changes:*
Currently, the guard utilizes a thread local. This works well for simple tests, but doesn't work on multithreaded tokio runtimes.

This adds another test-sink target `set_test_sink_on_runtime`.

It creates a `HashMap<RuntimeId, Sink>` which can then be used to lookup the correct sink based on the current active runtime.

We may want to use the same pattern for #103 

### DIFF
--- a/metrique-writer-core/src/global.rs
+++ b/metrique-writer-core/src/global.rs
@@ -607,6 +607,9 @@ macro_rules! global_entry_sink {
                         /// Returns a guard that will automatically remove the sink when dropped.
                         /// Only available when the `test-util` feature is enabled.
                         ///
+                        /// # Panics
+                        /// If this runtime already has a test sink installed.
+                        ///
                         /// # Example
                         #[doc = $crate::__macro_doctest!()]
                         /// # use metrique_writer::sink::global_entry_sink;
@@ -621,6 +624,7 @@ macro_rules! global_entry_sink {
                         ///     rt.block_on(async {
                         ///         // All appends on this runtime now go to the test sink
                         ///     });
+                        ///    // When the _guard is dropped, the sink is now detached and can be reattached again.
                         /// }
                         /// ```
                         #[track_caller]
@@ -654,7 +658,8 @@ macro_rules! global_entry_sink {
                         /// Only available when the `test-util` feature is enabled.
                         ///
                         /// # Panics
-                        /// Panics if called outside a tokio runtime context.
+                        /// - If called outside a tokio runtime context.
+                        /// - If a test sink is already installed on this runtime
                         ///
                         /// # Example
                         #[doc = $crate::__macro_doctest!()]
@@ -670,6 +675,8 @@ macro_rules! global_entry_sink {
                         ///     // `TestSink::sink()` will now always refer to the test sink on any thread on this runtime.
                         ///     // NOTE: that threads _outside_ this runtime (e.g. a background thread) will still NOT have this sink
                         ///     // installed.
+                        ///
+                        ///     // When _guard is dropped, the sink will be detached.
                         /// }
                         /// ```
                         #[track_caller]


metrique-writer-core/src/global.rs
```diff
@@ -542,6 +594,89 @@ macro_rules! global_entry_sink {
                             let _guard = Self::set_test_sink(sink);
                             f()
                         }
+
+                        /// Install a runtime-scoped test sink for a specific tokio runtime.
+                        ///
+                        /// This allows installing a test sink on a runtime from outside that runtime's context.
+                        /// The sink will be used by all tasks running on the specified runtime.
+                        ///
+                        /// **Note:** Most users should use [`set_test_sink_on_runtime`](Self::set_test_sink_on_runtime)
```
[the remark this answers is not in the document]
jlizen: +1, though I think that this comment belongs on `set_test_sink_on_runtime` rather than this method.

I'd also suggest including tokio in the names, probably:
` set_test_sink_on_tokio_runtime`
`set_test_sink_on_current_tokio_runtime`

metrique-writer-core/src/global.rs
```diff
@@ -542,6 +594,89 @@ macro_rules! global_entry_sink {
...
+                            if already_installed {
+                                panic!("A test sink was already installed for this runtime. You can only install one test sink per runtime at a time.");
+                            }
+
+                            $crate::global::TokioRuntimeTestSinkGuard::new(runtime_id, map.clone())
+                        }
+
+                        /// Install a runtime-scoped test sink for the current tokio runtime.
+                        ///
+                        /// Unlike `set_test_sink`, this guard is Send + Sync and works across
```
jlizen: From a user's perspective, is it really that interesting that the guard is `Send` + `Sync`? I might not mention it, personally, and let them see via docs.rs. It seems very niche to send the guard around instead of just holding it throughout your test.

Mostly what they care about is that the test sink will persist across threads within the same runtime.

metrique-writer-core/src/global.rs
```diff
@@ -542,6 +594,89 @@ macro_rules! global_entry_sink {
...
+                        /// threads within the same tokio runtime. The sink will be used by all
+                        /// tasks running on the current runtime.
+                        ///
+                        /// Returns a guard that will automatically remove the sink when dropped.
+                        /// Only available when the `test-util` feature is enabled.
+                        ///
+                        /// # Panics
+                        /// Panics if called outside a tokio runtime context.
```
jlizen: Or if the runtime already has a test sink installed?

metrique-writer-core/src/global.rs
```diff
@@ -481,6 +530,9 @@ macro_rules! global_entry_sink {
                         /// Returns a guard that will automatically restore the previous sink state when dropped.
                         /// Only available when the `test-util` feature is enabled.
                         ///
+                        /// **Note:** This guard is not Send/Sync. For multi-threaded tokio runtimes,
```
jlizen: I don't think the critical information here is that it is `Send/Sync` (that is niche and can be seen on docs.rs if they care). It is more that it won't keep sink installed across runtime threads, which affects tokio multi-threaded runtimes, including in tests.

metrique/docs/emf.md
```diff
@@ -119,7 +119,7 @@ Per entry dimensions can be set with the `emf::dimension_sets` attribute when de
 1. Only the root metric may define dimension_sets. Failing to do this will result in an error at runtime. This behavior may be improved in the future.
 2. There is no compile-time validation on the dimensions that are set. You must validate that that dimension sets are set as you expect. A good way to check this is by invoking the EMF formatter directly with `all_validations` enabled:
 
-    ```rust
+    ```rust,no_run
```
jlizen: Why `no_run`?

PR #194, opened 2026-01-30
feat: Add Debug derives to Slot, SlotGuard, and AppendAndCloseOnDrop

### LINKED ISSUE #190: implement Debug for Slot
We will want a custom Debug impl that still shows whether slot is open/closed/has data, even if `T: !Debug`

rcoh:
📬 *Issue #, if available:* Fixes #190 

✍️ *Description of changes:* Add `derive(Debug)` to Slot & friends. This also required updating the proc macro to pass through `Debug` when present.

### DIFF
--- a/metrique/src/slot.rs
+++ b/metrique/src/slot.rs
@@ -89,11 +89,23 @@ fn make_slot<T: CloseValue>(initial_value: T) -> (SlotGuard<T>, Waiting<T::Close
 ///     metrics.field_1 += 1;
 /// }
 /// ```
-#[derive(Debug)]
 pub struct Slot<T: CloseValue> {
     tx: Option<SlotGuard<T>>,
     rx: Option<Waiting<T::Closed>>,
     data: Option<T::Closed>,
+}
+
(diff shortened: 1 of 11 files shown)


metrique/src/slot.rs
```diff
@@ -257,6 +261,18 @@ pub struct SlotGuard<T: CloseValue> {
     parent_drop_mode: OnParentDrop,
 }
 
+impl<T: Debug + CloseValue> Debug for SlotGuard<T>
+where
+    T::Closed: Debug,
```
jlizen: This is fine, but is there a reason we wouldn't want to have a custom debug impl for `SlotI` instead of just the `#[derive(Debug)]`?

That way (I think) we could have one that is `T: Debug` and includes the inner contents, or `T: !Debug` which just shows whether it is still open or not

rcoh:
I did not opt for a opaque impl — you can't have both due to a lack of specialization and I think most users are fine with `#[derive(Debug)]` on their structs.

PR #197, opened 2026-02-01
ci: skip docs.rs check on release-* branches

rcoh:
The docs.rs check will always fail on release PRs since the versions haven't been published yet. Skip this check for branches starting with 'release-'.


jlizen (approved):
Makes sense

Issue #191, opened 2026-01-20
Add an example of showing how you can make a global or semi-global up/down counter


rcoh:
e.g. for tracking the number of currently outstanding requests

PR #207, opened 2026-02-17
Add example for global up/down counter

jasgin:
📬 *Issue #, if available:*

https://github.com/awslabs/metrique/issues/191

✍️ *Description of changes:*

Adds an example for a global up/down counter

- [X] I confirm that I've made a best effort attempt to update all relevant documentation.
- [X] I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/metrique/examples/global-counter.rs
+++ b/metrique/examples/global-counter.rs
@@ -1,10 +1,16 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! A global static counter, using in-flight requests as an example.
+//! A global counter, using in-flight requests as an example.
 //!
-//! See `scoped-counter.rs` for an alternative with better testing interactibility.
+//! [`GlobalCounter`] is an example of what you can do if you have want to initialize a global
+//! static counter or if you are already passing around an `Arc`-wrapped struct.
+//!
+//! For other most other non-static usage, doing something like [`RequestCounter`] will be more ergonomic
+//! and better for testing interactibility.
 
+use std::ops::Deref;
+use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


rcoh (commented):
ideally we'd show a more ergonomic pattern that had less boilerplate

metrique/examples/global-counter.rs
```diff
@@ -0,0 +1,51 @@
...
+    /*
+    {"_aws":{"CloudWatchMetrics":[{"Namespace":"Ns","Dimensions":[[]],"Metrics":[{"Name":"TotalRequests"}]}],"Timestamp":1771387941273},"TotalRequests":1}
+    {"_aws":{"CloudWatchMetrics":[{"Namespace":"Ns","Dimensions":[[]],"Metrics":[{"Name":"TotalRequests"}]}],"Timestamp":1771387941273},"TotalRequests":2}
+    */
+}
+
+async fn handle_request() {
+    let mut metrics = RequestMetrics::init();
+
+    metrics.total_requests = TOTAL_REQUESTS.fetch_add(1, Ordering::Relaxed) + 1;
```
rcoh: This doesn't use a guard and doesn't decrement?

metrique/examples/global-counter.rs
```diff
@@ -0,0 +1,85 @@
...
+use metrique::unit_of_work::metrics;
+use metrique::writer::{
+    AttachGlobalEntrySinkExt, FormatExt, GlobalEntrySink, sink::global_entry_sink,
+};
+global_entry_sink! { ServiceMetrics }
+
+/// Global counter to keep track of in-flight requests for the service's uptime
+static GLOBAL_REQUEST_COUNTER: LazyLock<GlobalCounter> = LazyLock::new(|| GlobalCounter::default());
+
+#[derive(Default)]
```
rcoh: Actually if we put it in a static you don't even need an arc. This is very similar to what @jlizen is working on in tokio-metrics

Not blocking here but we should consider this when we make an API

rcoh (approved):

metrique/examples/global-counter.rs
jlizen: I don't know if you need the lazy lock either without the arc. Just make GlobalCounter::new const (can't use default)

rcoh: Yeah exactly.

The reason to use the arc is you could embed it in a tower layer (which is another very common pattern).

In that case though you could arc the layer so, _shrug_

PR #224, opened 2026-03-11
feat: Add a pure JSON format

### LINKED ISSUE #192: Add a pure JSON backend (separate from EMF)
For usage that is not targeting EMF, a simpler format that is pure JSON would be helpful, easier to read as humans, and helpful for portability to other platforms.

There is a little bit of ambiguity to resolve (or to handle as a setting) around `Observation::Repeated` (and generally having multiple observations per metric). The EMF format currently produces it as a `{ Values: [....], Counts: [...] }`. Not clear that this is the right (or wrong) way for a generic JSON format but it is a point of contention.

yulnr:
Closes #192

## Summary

Implements a new pure JSON backend (new crate `metrique-writer-format-json`) sampled-output support.                                                                                         

Each entry is serialized as a single JSON line:                                                                                                                                                     
```json
{                                                                                                                                                                                                   
  "timestamp": 1705312800000,
  "metrics": {
    "Latency": { "value": 42.5, "unit": "Milliseconds" },
    "Count": { "value": 10 },
    "BackendLatency": { "value": { "total": 150, "count": 3 }, "unit": "Milliseconds" },
    "ResponseTimes": { "values": [1, 2, 3], "unit": "Milliseconds" }
  },
  "properties": {
    "Operation": "GetItem"
  }
}
```
                                                                                                   
- Single observations: "value": X                                                                                                                                                                   
- Multiple observations: "values": [...]                                                                                                                                                            
- Repeated observations: {"total": f64, "count": u64}
                                                                                                                                                                                                      
Observation format configurability (Histogram, Mean, etc.) was intentionally deferred to follow-up work per feedback. We're shipping a minimal version first and we'll revisit observation output format.

### Noteworthy decisions

I'm happy to be corrected on any of these:
- Dimensions, flags and config are intentionally out-of-scope for JSON v0
- Non-finite floats differ from EMF in this backend: EMF drops them, while pure JSON serializes them as JSON null. I don't have a strong opinion here, happy to change into dropping them if that's desired. My initial thought was that since we don't have EMF constraints, dropping them is not required. Also it might be a good point of information for some users, e.g. warning that something is wrong if they didn't expect nulls, or not missing observations.

### Implementation notes:
I used `ryu` for converting floats to strings, instead of `dtoa` (what `emf` is using). According to benchmarks `ryu` should perform better across the board nowadays (IIRC some time ago `dtoa` was still better for few-digits numbers). I'm happy to change to `dtoa` if there's a reason for it, and I wonder if really need both in these crates (I used `ryu` because Metrique already uses it somewhere else).

### Testing:

I tried to cover all the bases with the included tests, but I could be missing some cases that I'm not familiar with yet. I did fuzz test this to make sure it always renders valid JSON, I'll submit the fuzzer tomorrow. 

### DIFF
--- a/metrique-writer-format-json/README.md
+++ b/metrique-writer-format-json/README.md
@@ -24,8 +24,4 @@ Each entry is serialized as a single JSON line:
 }
 ```
 
-Single observations are emitted as `"value": X`, multiple as `"values": [...]`.
-Repeated observations (e.g. from histogram buckets) are emitted as
-`{"total": ..., "count": ...}`.
-
 [Format]: https://docs.rs/metrique-writer/latest/metrique_writer/format/trait.Format.html


metrique-writer-format-json/src/json.rs
```diff
@@ -0,0 +1,1080 @@
...
+    } else {
+        let mut buffer = ryu::Buffer::new();
+        let s = buffer.format_finite(v);
+        // Strip trailing ".0" for cleaner integer-like output
+        buf.push_str(s.strip_suffix(".0").unwrap_or(s));
+    }
+}
+
+/// Push a JSON-escaped string with surrounding quotes into the buffer.
+fn push_json_string(buf: &mut String, s: &str) {
```
yulnr: Heads-up, I wrote this JSON string escaper instead of using `serde_json::to_writer` (as EMF currently does). It should avoid some serializer overhead and covers required JSON escaping. Let me know if I'm missing anything else that would require to use that from serde here. 
I thought I might avoid it based on this comment:
https://github.com/awslabs/metrique/blob/cabb9253797d4c802cbda35a8d34b76e4d278a8c/metrique-writer-format-emf/src/json_string.rs#L11-L14

metrique-writer-format-json/src/json.rs
```diff
@@ -0,0 +1,1080 @@
...
+    }
+}
+
+/// Push a float value, clamping infinities and writing null for NaN.
+fn push_float(buf: &mut String, v: f64) {
+    let v = v.clamp(-f64::MAX, f64::MAX);
+    if v.is_nan() {
+        buf.push_str("null");
+    } else {
+        let mut buffer = ryu::Buffer::new();
```
jlizen: This will result in scientific notation, which can be pretty annoying to script against in the output json. Would it be worth using `dtoa` instead to get the nicer decimal readability?

metrique-writer-format-json/src/json.rs
```diff
@@ -0,0 +1,1080 @@
...
+    buf.push_str(&s[start..]);
+    buf.push('"');
+}
+
+/// A wrapper around [`Json`] that supports sampling. Datapoints are emitted with
+/// multiplicity equal to either `floor(1/rate)` or `ceil(1/rate)` to ensure
+/// statistics are unbiased.
+///
+/// See [`Json::with_sampling`].
+pub struct SampledJson<R = DefaultRng<ThreadRng>> {
```
jlizen: needs `#[derive(Debug)]`

README.md
```diff
@@ -24,7 +24,8 @@ struct RequestMetrics {
 }
 ```
 
-This currently supports exporting metrics in [Amazon EMF] format to CloudWatch and printing them locally with [`metrique::local::LocalFormat`.
+This currently supports exporting metrics in [Amazon EMF] format to CloudWatch and as plain JSON for non-AWS systems. 
+Metrics can be printed locally with [`metrique::local::LocalFormat`].
```
jlizen: (separate scope)
We should probably fix these README doc links to work both on github / crates.io and in rustdocs. I think optimizing for rustdocs is fine in the interim. Could use a backlog issue though.

metrique-writer-format-json/src/json.rs
```diff
@@ -0,0 +1,1080 @@
...
+                    buf.push_str(itoa::Buffer::new().format(occurrences.saturating_mul(mult)));
+                    buf.push('}');
+                }
+                RepeatedFormat::Mean => {
+                    push_float(buf, repeated_mean(total, occurrences));
+                }
+            }
+        }
+        _ => {
+            buf.push_str("null");
```
jlizen: ++ to serializing as `null`, generally seems like a better ux compared to chasing down missing metrics. I wish `NaN` or `+-Infinity` was valid json...

It might cause certain backends to fall over but that seems like a bridge we can cross with more configuration, if anybody asks.

metrique-writer-format-json/src/json.rs
```diff
@@ -0,0 +1,1081 @@
...
+    if occurrences == 0 {
+        0.0
+    } else {
+        total / occurrences as f64
+    }
+}
+
+/// Push a float value, clamping infinities and writing null for NaN.
+fn push_float(buf: &mut String, v: f64) {
+    let v = v.clamp(-f64::MAX, f64::MAX);
```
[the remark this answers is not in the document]
jlizen: lgtm

yulnr:
Thanks for feedback! I've removed the observation format configuration and kept the simpler behavior.

PR #234, opened 2026-03-14
`ArcSwap` integration in `metrique-core`

jlizen:
✍️ *Description of changes:*

Are we willing to take a public dependency on `arc-swap@1.x`? It makes for an awfully nice wrapper for occasionally changed globals. It's hard to imagine anybody implementing anything different. I don't see them breaking anytime soon.

Nicer to have the boilerplate in core if we are willing.

Right now it is behind a non-default feature flag. But it adds literally 0 other dependencies, if we care to turn it on by default (or ditch the feature flag entirely). 

The feature flag (`arcswap`) is re-exported from `metric-core` to `metrique`. We could also re-export it from `metrique-aggregation` if we someday have some cool eventually consistent bucket thing.

### DIFF
--- a/metrique/docs/concurrency.md
+++ b/metrique/docs/concurrency.md
@@ -223,7 +223,7 @@ use arc_swap::ArcSwap;
 #[metrics(subfield_owned, rename_all = "PascalCase")]
 #[derive(Clone)]
 struct FeatureFlags {
-    dark_mode_enabled: bool,
+    feature_xyz_enabled: bool,
 }
 
 // Owned ArcSwap: useful when the metrics struct owns the swap.
--- a/metrique/examples/global-counter.rs
+++ b/metrique/examples/global-counter.rs
@@ -1,17 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Global state patterns for metrics: counters and swappable feature flags.
-//!
-//! [`GlobalCounter`] is an example of what you can do if you have want to initialize a global
-//! static counter or if you are already passing around an `Arc`-wrapped struct.
-//!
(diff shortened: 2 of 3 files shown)


metrique/docs/concurrency.md
```diff
@@ -201,4 +201,47 @@ fn count_concurrent_ducks() {
...
+struct ServerState {
+    region: String,
+    cell: String,
+}
+
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    operation: &'static str,
+    #[metrics(flatten)]
+    server_state: &'static ArcSwap<ServerState>,
```
rcoh: Maybe we can find a better use case. This seems better handled by merge globals?

metrique/examples/global-counter.rs
```diff
@@ -80,16 +90,48 @@ impl Drop for RequestCounterGuard {
...
+// ---------------------------------------------------------------------------
+
+/// Process-lifetime server state, readable from any task.
+static SERVER_STATE: LazyLock<ArcSwap<ServerStateDimensions>> =
+    LazyLock::new(|| ArcSwap::from_pointee(read_server_state()));
+
+#[metrics(subfield_owned, rename_all = "PascalCase")]
+#[derive(Clone)]
+struct ServerStateDimensions {
+    region: String,
```
rcoh: These are unchanging in practice but maybe we could pick something like xyz_feature_enabled

metrique/examples/global-counter.rs
```diff
@@ -107,14 +149,15 @@ async fn main() {
 
     handle_request(scoped_request_counter.clone()).await;
     handle.await.unwrap();
-    handle_request(scoped_request_counter).await;
 
-    // EXAMPLE OUTPUT
```
rcoh: Why did we delete the example output?

jlizen:
Closing in favor of a newer, scope-creeped one

PR #235, opened 2026-03-15
feat: Shared state primitives, metrique-util crate, and documentation overhaul

jlizen:
## `State<T>` (crate: `metrique-util`, feature: `state`)

Atomically swappable value with snapshot-on-first-read semantics. The first `snapshot()` captures the current value via `OnceLock`; subsequent calls return the same `Arc<T>`. `clone()` shares the underlying swap but resets the snapshot, making it the natural way to create per-request handles from shared state.

```rust
let shared = State::new("v1".to_string());

let request = shared.clone(); // fresh handle, no snapshot yet
assert_eq!(*request.snapshot(), "v1"); // first snapshot captures "v1"

shared.store(Arc::new("v2".into()));

assert_eq!(*request.snapshot(), "v1"); // still returns captured value
assert_eq!(*request.latest(), "v2"); // bypasses snapshot, sees latest

let request2 = shared.clone(); // new handle, captures fresh
assert_eq!(*request2.snapshot(), "v2");
```

The backing `ArcSwap` is fully opaque. This allows changing the implementation without a breaking change and keeps the API minimal.

`OnceLock` was chosen for the snapshot over `Mutex` (lock on every read) or a second `ArcSwap` (would eliminate snapshot semantics entirely).

The inner `ArcSwap` is unconditionally wrapped in `Arc` so that `clone()` can share the swap cheaply, and so `latest()` can provide an escape hatch for hot paths that need to read the latest value without snapshot semantics. This means that we only support `ArcSwap<T>`, not `ArcSwapOption<T>`; users who need optionality can use `State<Option<T>>` as a mediocre "we have ArcSwapOption at home" equivalent. The ergonomics on keeping `ArcSwapAny`-like generic were just too icky for the common case... We can always add more APIs if needed.

`State` lives in the new `metrique-util` crate rather than `metrique` or `metrique-core`. It is not needed in core for trait coherence, and `metrique-util` serves as a home for additional utilities that don't belong in the core crates. Not re-exported from `metrique`; users depend on `metrique-util` directly.

## `OnceLock<T>` blanket `CloseValue` impl

Useful for any value initialized once at runtime: environment variables, `State<T>`, configuration loaded from disk, etc. Closes as `None` if uninitialized.

```rust
static NODE_GROUP: OnceLock<State<String>> = OnceLock::new();

#[metrics(subfield_owned)]
struct StaticState {
    node_group: &'static OnceLock<State<String>>,
}
```

## `CounterGuard` implements `CloseValue`

Reads the atomic counter value at close time. The guard still decrements on drop afterward (orthogonal to the close snapshot).

## `Counter::increment_scoped` and newly-const `Counter::new`

`const fn new` enables `static Counter` declarations. `increment_scoped()` returns a guard that decrements on drop, for tracking in-flight work. Replaces the custom `RequestCounter` from the old `global-counter.rs` example with a built-in primitive.

```rust
static IN_FLIGHT: Counter = Counter::new(0);

let _guard = IN_FLIGHT.increment_scoped(); // +1 now, -1 on drop
```

## `global-state.rs` example

Migrated from the handrolled RequestCounter to our new `Counter::increment_scoped` API.

Demonstrates both static/borrowed and owned patterns, with each flattened into per-request metrics. Background tasks atomically refresh application state on a loop, and per-request metrics capture snapshots via a cloned handle on the same struct.

## Documentation

- `cookbook.md`: added "shared state in unit-of-work" pattern, reworked periodic metrics to pitch attachment to unit-of-work over standalone emission.
- `concurrency.md`: updated for `State` snapshot-on-first-read semantics.

---

### DIFF
--- a/metrique-util/tests/state.rs
+++ b/metrique-util/tests/state.rs
@@ -10,11 +10,19 @@ use metrique::writer::sink::VecEntrySink;
 use metrique::writer::test_util;
 use metrique_util::State;
 
-#[metrics(subfield_owned)]
 #[derive(Clone, Debug, Default)]
+#[metrics(subfield)]
 struct AppConfig {
     feature_xyz_enabled: bool,
-    traffic_policy: String,
+    traffic_policy: TrafficPolicy,
+}
+
+#[derive(Clone, Copy, Debug, Default)]
+#[metrics(value(string))]
+enum TrafficPolicy {
+    #[default]
+    Default,
+    Canary,
 }
 
 #[metrics(rename_all = "PascalCase")]
@@ -33,7 +41,7 @@ fn state_flattened() {
         operation: "PutItem",
         config: State::new(AppConfig {
             feature_xyz_enabled: false,
-            traffic_policy: "default".into(),
+            traffic_policy: TrafficPolicy::Default,
         }),
         duck_count: 0,
     }
@@ -55,7 +63,7 @@ fn state_snapshot_on_first_load() {
     let vec_sink = VecEntrySink::new();
     let state = State::new(AppConfig {
         feature_xyz_enabled: false,
-        traffic_policy: "default".into(),
+        traffic_policy: TrafficPolicy::Default,
     });
 
     let metrics = MyMetrics {
@@ -71,7 +79,7 @@ fn state_snapshot_on_first_load() {
     // Update after the snapshot was captured.
     state.store(Arc::new(AppConfig {
         feature_xyz_enabled: true,
-        traffic_policy: "canary".into(),
+        traffic_policy: TrafficPolicy::Canary,
     }));
 
     drop(metrics);
@@ -89,7 +97,7 @@ fn state_across_config_reload() {
     let vec_sink = VecEntrySink::new();
     let state = State::new(AppConfig {
         feature_xyz_enabled: false,
-        traffic_policy: "default".into(),
+        traffic_policy: TrafficPolicy::Default,
     });
 
     // req1: clone and load before the swap, closed after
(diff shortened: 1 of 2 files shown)


metrique-core/Cargo.toml
```diff
@@ -8,9 +8,13 @@ description = "Library for working with unit of work metrics - core traits"
 repository = "https://github.com/awslabs/metrique"
 readme = "README.md"
 
+[features]
```
rcoh: This is a new type which means we don't have to put it in metrique core for coherence reasons. Could have it in metrique directly

metrique-core/src/atomics.rs
```diff
@@ -22,6 +22,13 @@ impl Counter {
         self.0.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
     }
 
+    /// Increments the count by 1, returning a guard that decrements the count
+    /// on drop, and the new value. Useful for tracking in-flight operations.
+    pub fn increment_scoped(&self) -> (CounterGuard<'_>, u64) {
```
rcoh: Nice

metrique-core/src/atomics.rs
```diff
@@ -33,6 +40,23 @@ impl Counter {
     }
 }
 
+/// A guard that decrements a [`Counter`] when dropped.
+///
+/// Returned by [`Counter::increment_scoped`].
+pub struct CounterGuard<'a>(&'a AtomicU64);
```
rcoh: Shouldn't counter guard impl close value?

jlizen:
@rcoh @yulnr this is ready for fresh eyes, PR description updated

PR #251, opened 2026-03-19
Prevent panic when sink is not attached

### LINKED ISSUE #139: Add a capability to drop entries without panicking if a sink is not attached
If a sink is not attached, append will panic:

```
impl<Q: AttachGlobalEntrySink> GlobalEntrySink for Q {
    #[track_caller]
    fn sink() -> BoxEntrySink {
        Q::try_sink().expect("sink must be `attach()`ed before use")
    }

    #[track_caller]
    fn append(entry: impl Entry + Send + 'static) {
        if Q::try_append(entry).is_err() {
            panic!("sink must be `attach()`ed before appending")
        }
    }
}
```

Test sink, and DevNullSink cover most of out use cases, but in some testing setups we don't even want to deal with attaching a thread local override and are Ok dropping entries on the floor.

Fluzko:
closes #139 

## Summary
There are some scenarios while testing that might be useful to avoid attaching a `sink`. This PR aims to support those cases under `test-util` flag through a `sink_or_noop` that will _silently_ discard the entries if not `sink`. If there's one, it will behave as a normal `sink`.


metrique-writer-core/src/global.rs
```diff
@@ -596,6 +596,63 @@ macro_rules! global_entry_sink {
...
+                        ///     let _my_metrics =
+                        ///         MyMetrics { operation: "test" }.append_on_drop(ServiceMetrics::sink_or_noop());
+                        ///     drop(_my_metrics);
+                        ///
+                        ///     assert_eq!(inspector.entries()[0].values["Operation"], "test");
+                        /// }
+                        /// ```
+                        pub fn sink_or_noop() -> BoxEntrySink {
+                            <Self as $crate::global::AttachGlobalEntrySink>::try_sink()
+                                .unwrap_or_else(BoxEntrySink::noop)
```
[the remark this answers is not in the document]
jv1i: I had the same concern when reviewing this, but I saw that this method is inside the `$crate::__test_util! { ... }` block that wraps `impl $name`, and when test-util is disabled `__test_util!` expands to nothing:
https://github.com/awslabs/metrique/blob/24c673d9c98977b2ad9793fd405f576ad8909d09/metrique-writer-core/src/global.rs#L993-L1007
So `sink_or_noop` wouldn't exist right? I tested this locally but perhaps I'm missing something
(still +1 to the idea of catching these in CI)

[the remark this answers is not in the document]
rcoh: oh you are right! should have expanded the context of the diff, my bad

metrique-writer-core/src/sink.rs
```diff
@@ -133,6 +133,21 @@ impl BoxEntrySink {
     pub fn new(sink: impl EntrySink<BoxEntry> + Send + Sync + 'static) -> Self {
         Self(Arc::new(Box::new(sink)))
     }
+
+    /// Returns a [`BoxEntrySink`] that silently discards all entries.
+    ///
+    /// Only available when the `test-util` feature is enabled.
+    #[cfg(feature = "test-util")]
+    pub fn noop() -> Self {
```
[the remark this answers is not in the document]
Fluzko: I think I get the idea, check [d74d49e](https://github.com/awslabs/metrique/pull/251/commits/d74d49e0521f809be73ac49908e12a315e225531) and let me know if it's what you meant

metrique-writer-core/src/global.rs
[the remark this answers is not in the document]
Fluzko: So as @yulnr mentioned, this is already protected by ` __test_util`. @rcoh, regarding [your comment](https://github.com/awslabs/metrique/pull/251#discussion_r2960550694) below, and having it async, it might be good to extract it from this place and leave it out of that test feature. Thoughts?

metrique-writer-core/src/sink.rs
```diff
@@ -133,6 +133,47 @@ impl BoxEntrySink {
...
```
rcoh: I'd probably prefer a name like `DropAllEntries` also we already have `DevNullSink`?

metrique-writer-core/src/sink.rs
```diff
@@ -215,3 +256,37 @@ impl<E: Entry, Q: EntrySink<E>> DerefMut for AppendOnDrop<E, Q> {
...
+            BoxEntrySink::lazy(move || Some(BoxEntrySink::new(MarkerSink(appended_clone.clone()))));
+        sink.append_any(TestEntry(1));
+        assert!(*appended.lock().unwrap());
+    }
+
+    #[test]
+    fn lazy_sink_discards_when_none() {
+        let sink = BoxEntrySink::lazy(|| None);
+        sink.append_any(TestEntry(1));
+    }
```
rcoh: this does not actually test the lazy resolution behavior

metrique/docs/testing.md
```diff
@@ -63,6 +65,33 @@ let _guard = ServiceMetrics::set_test_sink(sink);
 
 See `examples/testing.rs` and `examples/testing-global-queues.rs` for more detailed examples.
 
+### Ignoring metrics with `sink_or_noop`
```
rcoh: this is actually useful not just for testing, but for libraries who want to emit to a sink when its available

metrique-writer-core/src/global.rs
```diff
@@ -596,6 +596,63 @@ macro_rules! global_entry_sink {
...
+                        ///     let _guard = ServiceMetrics::set_test_sink(sink);
+                        ///
+                        ///     let _my_metrics =
+                        ///         MyMetrics { operation: "test" }.append_on_drop(ServiceMetrics::sink_or_noop());
+                        ///     drop(_my_metrics);
+                        ///
+                        ///     assert_eq!(inspector.entries()[0].values["Operation"], "test");
+                        /// }
+                        /// ```
+                        pub fn sink_or_noop() -> BoxEntrySink {
```
rcoh: not sure about the `sink_or_noop` name. wdyt @jlizen ?

jlizen: it's ok, but i might prefer something like `ServiceMetrics::with_optional_sink` or something

rcoh: Or maybe sink_or_discard? I don't like sink_or_noop

jlizen: That's fine too, sink or discard is a bit clearer about what is happening

metrique-writer-core/src/sink.rs
Fluzko: This is already removed, since the noop isn't used anymore. Also, DevNullSink acts differently from the current lazy approach

metrique-writer-core/src/sink.rs
Fluzko: you're right, enhanced it and added a few more, let me know

rcoh (approved):

PR #256, opened 2026-03-24
feat: Add tokio-metrics runtime metrics integration

### LINKED ISSUE #248: metrique <-> tokio-metrics bridge
We want first-class integration between metrique and tokio-metrics.
The key behavior here is making it so that `RuntimeMetrics` conditionally derives metrique entry.

Ideal API:

```rust
ServiceMetrics::attach_to_stream(
    Emf::builder("Ns".to_string(), vec![vec![]])
        .build()
        .output_to_makewriter(RollingFileAppender::new(
            Rotation::MINUTELY,
            &service_log_dir,
            "service_log.log",
        )),
).with_tokio_runtime_metrics()
```
(approximately)

yulnr:
Closes #248

## Depends on 
https://github.com/tokio-rs/tokio-metrics/pull/121 - once merged and released, will update `metrique-util/Cargo.toml` from git dep to crates.io version. The `[patch.crates-io]` section in the workspace `Cargo.toml` should needs to be removed.

## Summary

Adds a new feature `tokio-metrics-bridge` to `metrique-util` which adds a bridge for tokio-metrics.
 
**Main Changes**
- **metrique-writer-core**: Refactors `join` in `AttachHandle` into a shutdown registry. Background tasks register cleanup functions via `register_shutdown_fn` that run (in LIFO order) before the sink detaches on drop.
- **metrique-util**: New `tokio-metrics-bridge` feature with `AttachGlobalEntrySinkTokioMetricsExt::subscribe_tokio_runtime_metrics`. Spawns a reporter task that appends `RuntimeMetrics` snapshots at a configurable interval, automatically aborted on handle drop.

## Usage

```rust
use metrique_util::{AttachGlobalEntrySinkTokioMetricsExt, TokioRuntimeMetricsConfig};

// After attaching a sink:
let _handle = ServiceMetrics::attach_to_stream(emf.output_to(std::io::stderr()));

// Subscribe: reporter runs in the background, aborted when _handle drops.
let config = TokioRuntimeMetricsConfig::default().with_interval(Duration::from_secs(30));
ServiceMetrics::subscribe_tokio_runtime_metrics(config);
```
See `metrique/examples/tokio-metrics.rs` for a better example.

## How it works

The `RuntimeMetrics` struct in tokio-metrics derives `metrique_writer::Entry` (behind the `metrique-integration` feature flag), so each snapshot is appended directly to the sink as a structured metric record with all runtime fields (worker utilization, park counts, queue depths, steal counts, poll durations, etc.).

## Open question: duplicate subscriptions

Right now nothing prevents calling `subscribe_tokio_runtime_metrics` twice, which would just spawn two reporter tasks appending duplicate metrics. It won't break anything, but it's almost certainly users won't intend that to happen.

I left it unguarded for now since the options each have tradeoffs and I'd rather get some input:

- Panic on duplicate: track whether a sink type has already subscribed and panic on the second call. Needs a per-type flag.
- Just log a warning
- Do nothing 

Happy to add a guard if there is a preferred approach, otherwise I'm leaning towards having at least a warning. 

**_Update_:** this now warns of duplicate subscriptions

## Testing
- Adds some shutdown registry tests to metrique-writer-core
- Adds tests to metrique-util for the tokio-metrics-feature

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,9 @@ trybuild = "1.0"
 toml = "0.9"
 walkdir = "2"
 
+# External crates that depend on metrique pull it from crates.io, while workspace
+# members use local paths. Without these patches, Cargo treats them as separate
+# crates, causing duplicate type errors. Only affects workspace builds.
 [patch.crates-io]
 metrique = { path = "metrique" }
 metrique-core = { path = "metrique-core" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,13 +68,11 @@ trybuild = "1.0"
 toml = "0.9"
 walkdir = "2"
 
-# TODO: Remove before merge — tokio-metrics depends on metrique from
-# crates.io; patch to local to avoid duplicate crate errors.
 [patch.crates-io]
 metrique = { path = "metrique" }
 metrique-core = { path = "metrique-core" }
-metrique-writer-core = { path = "metrique-writer-core" }
-metrique-writer = { path = "metrique-writer" }
 metrique-macro = { path = "metrique-macro" }
 metrique-timesource = { path = "metrique-timesource" }
+metrique-writer = { path = "metrique-writer" }
+metrique-writer-core = { path = "metrique-writer-core" }
 metrique-writer-macro = { path = "metrique-writer-macro" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,8 +69,8 @@ toml = "0.9"
 walkdir = "2"
 
 # TODO: Remove before merge — tokio-metrics depends on metrique from
-# awslabs git; patch to local to avoid duplicate crate errors.
-[patch."https://github.com/awslabs/metrique"]
+# crates.io; patch to local to avoid duplicate crate errors.
+[patch.crates-io]
 metrique = { path = "metrique" }
 metrique-core = { path = "metrique-core" }
 metrique-writer-core = { path = "metrique-writer-core" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,10 +68,13 @@ trybuild = "1.0"
 toml = "0.9"
 walkdir = "2"
 
-# TODO: Remove before merge — temporary overrides so that tokio-metrics'
-# transitive metrique-writer dep resolves to the local workspace versions,
-# avoiding duplicate-crate errors.
-[patch.crates-io]
+# TODO: Remove before merge — tokio-metrics depends on metrique from
+# awslabs git; patch to local to avoid duplicate crate errors.
+[patch."https://github.com/awslabs/metrique"]
+metrique = { path = "metrique" }
 metrique-core = { path = "metrique-core" }
 metrique-writer-core = { path = "metrique-writer-core" }
 metrique-writer = { path = "metrique-writer" }
+metrique-macro = { path = "metrique-macro" }
+metrique-timesource = { path = "metrique-timesource" }
+metrique-writer-macro = { path = "metrique-writer-macro" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,11 @@ tracing-subscriber = "0.3.20"
 trybuild = "1.0"
 toml = "0.9"
 walkdir = "2"
+
+# TODO: Remove before merge — temporary overrides so that tokio-metrics'
+# transitive metrique-writer dep resolves to the local workspace versions,
+# avoiding duplicate-crate errors.
+[patch.crates-io]
+metrique-core = { path = "metrique-core" }
+metrique-writer-core = { path = "metrique-writer-core" }
+metrique-writer = { path = "metrique-writer" }
--- a/metrique-util/Cargo.toml
+++ b/metrique-util/Cargo.toml
@@ -30,6 +30,7 @@ tokio-metrics = { version = "0.5.0", optional = true, features = ["rt", "metriqu
 tracing = { workspace = true, optional = true }
 
 [dev-dependencies]
+assert2 = { workspace = true }
 metrique = { path = "../metrique", features = ["emf", "test-util"] }
 metrique-writer = { path = "../metrique-writer", features = ["test-util"] }
 metrique-writer-core = { path = "../metrique-writer-core", features = ["test-util"] }
--- a/metrique-util/Cargo.toml
+++ b/metrique-util/Cargo.toml
@@ -34,6 +34,7 @@ metrique = { path = "../metrique", features = ["emf", "test-util"] }
 metrique-writer = { path = "../metrique-writer", features = ["test-util"] }
 metrique-writer-core = { path = "../metrique-writer-core", features = ["test-util"] }
 tokio = { workspace = true, features = ["full", "test-util"] }
+rstest = { workspace = true }
 metrique-util = { path = ".", features = ["state"] }
 
 [package.metadata.docs.rs]
--- a/metrique-util/Cargo.toml
+++ b/metrique-util/Cargo.toml
@@ -8,13 +8,27 @@ description = "Additional utilities for metrique"
 repository = "https://github.com/awslabs/metrique"
 readme = "README.md"
 
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tokio_unstable)'] }
+
 [features]
 default = []
 state = ["dep:arc-swap"]
+tokio-metrics-bridge = [
+    "dep:metrique-writer-core",
+    "dep:tokio",
+    "dep:tokio-metrics",
+    "dep:tracing",
+]
 
 [dependencies]
 metrique-core = { path = "../metrique-core", version = "0.1.18" }
 arc-swap = { version = "1", optional = true }
+metrique-writer-core = { path = "../metrique-writer-core", version = "0.1.13", optional = true }
+tokio = { workspace = true, optional = true, features = ["time", "rt"] }
+# TODO: Update before merge, should point to tokio-metrics on crates.io
+tokio-metrics = { git = "https://github.com/yulnr/tokio-metrics", branch = "metrique-entry-support", optional = true, features = ["rt", "metrique-integration"] }
+tracing = { workspace = true, optional = true }
 
 [dev-dependencies]
 metrique = { path = "../metrique", features = ["emf", "test-util"] }
--- a/metrique-util/Cargo.toml
+++ b/metrique-util/Cargo.toml
@@ -15,7 +15,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tokio_unstable)'] }
 default = []
 state = ["dep:arc-swap"]
 tokio-metrics-bridge = [
-    "dep:metrique-writer-core",
+    "dep:metrique",
     "dep:tokio",
     "dep:tokio-metrics",
     "dep:tracing",
@@ -23,11 +23,10 @@ tokio-metrics-bridge = [
 
 [dependencies]
 metrique-core = { path = "../metrique-core", version = "0.1.18" }
+metrique = { path = "../metrique", version = "0.1.23", optional = true, default-features = false }
 arc-swap = { version = "1", optional = true }
-metrique-writer-core = { path = "../metrique-writer-core", version = "0.1.13", optional = true }
 tokio = { workspace = true, optional = true, features = ["time", "rt"] }
-# TODO: Update before merge, should point to tokio-metrics on crates.io
-tokio-metrics = { git = "https://github.com/yulnr/tokio-metrics", branch = "metrique-entry-support", optional = true, features = ["rt", "metrique-integration"] }
+tokio-metrics = { version = "0.5.0", optional = true, features = ["rt", "metrique-integration"] }
 tracing = { workspace = true, optional = true }
 
 [dev-dependencies]
--- a/metrique-util/README.md
+++ b/metrique-util/README.md
@@ -5,7 +5,9 @@ Additional utilities for [metrique].
 ## Features
 
 - `state`: Provides [`State<T>`], an atomically swappable shared value with snapshot-on-first-read semantics. Useful for shared runtime state (feature flags, config reloads, routing tables) that should appear on every metric record.
-- `tokio-metrics-bridge`: Subscribes [tokio-metrics](https://crates.io/crates/tokio-metrics) runtime snapshots to a global entry sink. The reporter task is automatically aborted when the `AttachHandle` is dropped.
+- `tokio-metrics-bridge`: Subscribes [tokio-metrics] runtime snapshots to a global entry sink. The reporter task is automatically aborted when the `AttachHandle` is dropped.
+
+[tokio-metrics]: https://crates.io/crates/tokio-metrics
 
 ## Usage
 
--- a/metrique-util/README.md
+++ b/metrique-util/README.md
@@ -5,6 +5,7 @@ Additional utilities for [metrique].
 ## Features
 
 - `state`: Provides [`State<T>`], an atomically swappable shared value with snapshot-on-first-read semantics. Useful for shared runtime state (feature flags, config reloads, routing tables) that should appear on every metric record.
+- `tokio-metrics-bridge`: Subscribes [tokio-metrics](https://crates.io/crates/tokio-metrics) runtime snapshots to a global entry sink. The reporter task is automatically aborted when the `AttachHandle` is dropped.
 
 ## Usage
 
--- a/metrique-util/src/dynamic_inflection.rs
+++ b/metrique-util/src/dynamic_inflection.rs
@@ -48,10 +48,10 @@ where
 
 #[cfg(test)]
 mod tests {
+    use assert2::check;
     use metrique::CloseValue;
     use metrique::unit_of_work::metrics;
     use metrique_core::DynamicNameStyle;
-    use assert2::check;
     use metrique_writer::test_util::to_test_entry;
     use rstest::rstest;
 
--- a/metrique-util/src/dynamic_inflection.rs
+++ b/metrique-util/src/dynamic_inflection.rs
@@ -51,6 +51,7 @@ mod tests {
     use metrique::CloseValue;
     use metrique::unit_of_work::metrics;
     use metrique_core::DynamicNameStyle;
+    use assert2::check;
     use metrique_writer::test_util::to_test_entry;
     use rstest::rstest;
 
@@ -89,8 +90,8 @@ mod tests {
             name_style: style,
         };
         let t = to_test_entry(entry);
-        assert_eq!(t.metrics[expected_prefixed], 42);
-        assert_eq!(t.metrics[expected_plain], 7);
+        check!(t.metrics[expected_prefixed] == 42);
+        check!(t.metrics[expected_plain] == 7);
     }
 
     #[metrics]
@@ -128,8 +129,8 @@ mod tests {
             name_style: style,
         };
         let t = to_test_entry(entry);
-        assert_eq!(t.metrics[request_key], 5);
-        assert_eq!(t.metrics[workers_key], 0);
-        t.metrics[park_key].as_u64();
+        check!(t.metrics[request_key] == 5);
+        check!(t.metrics[workers_key] == 0);
+        check!(t.metrics[park_key].as_u64() == 0);
     }
 }
--- a/metrique-util/src/dynamic_inflection.rs
+++ b/metrique-util/src/dynamic_inflection.rs
@@ -45,3 +45,91 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use metrique::CloseValue;
+    use metrique::unit_of_work::metrics;
+    use metrique_core::DynamicNameStyle;
+    use metrique_writer::test_util::to_test_entry;
+    use rstest::rstest;
+
+    use super::DynamicInflectionEntry;
+
+    #[metrics(subfield_owned)]
+    #[derive(Default)]
+    struct Inner {
+        count: u64,
+    }
+
+    #[metrics]
+    #[derive(Default)]
+    struct Outer {
+        #[metrics(flatten, prefix = "pfx_")]
+        inner: Inner,
+        other: u64,
+    }
+
+    #[rstest]
+    #[case::identity(DynamicNameStyle::Identity, "pfx_count", "other")]
+    #[case::pascal_case(DynamicNameStyle::PascalCase, "PfxCount", "Other")]
+    #[case::snake_case(DynamicNameStyle::SnakeCase, "pfx_count", "other")]
+    #[case::kebab_case(DynamicNameStyle::KebabCase, "pfx-count", "other")]
+    fn prefix_inflection(
+        #[case] style: DynamicNameStyle,
+        #[case] expected_prefixed: &str,
+        #[case] expected_plain: &str,
+    ) {
+        let entry = DynamicInflectionEntry {
+            entry: Outer {
+                inner: Inner { count: 42 },
+                other: 7,
+            }
+            .close(),
+            name_style: style,
+        };
+        let t = to_test_entry(entry);
+        assert_eq!(t.metrics[expected_prefixed], 42);
+        assert_eq!(t.metrics[expected_plain], 7);
+    }
+
+    #[metrics]
+    struct WithRuntimeMetrics {
+        #[metrics(flatten, prefix = "rt_")]
+        runtime: tokio_metrics::RuntimeMetrics,
+        request_count: u64,
+    }
+
+    #[rstest]
+    #[case::identity(
+        DynamicNameStyle::Identity,
+        "request_count",
+        "rt_workers_count",
+        "rt_total_park_count"
+    )]
+    #[case::pascal_case(
+        DynamicNameStyle::PascalCase,
+        "RequestCount",
+        "RtWorkersCount",
+        "RtTotalParkCount"
+    )]
+    fn runtime_metrics_with_prefix(
+        #[case] style: DynamicNameStyle,
+        #[case] request_key: &str,
+        #[case] workers_key: &str,
+        #[case] park_key: &str,
+    ) {
+        let entry = DynamicInflectionEntry {
+            entry: WithRuntimeMetrics {
+                runtime: tokio_metrics::RuntimeMetrics::default(),
+                request_count: 5,
+            }
+            .close(),
+            name_style: style,
+        };
+        let t = to_test_entry(entry);
+        assert_eq!(t.metrics[request_key], 5);
+        assert_eq!(t.metrics[workers_key], 0);
+        t.metrics[park_key].as_u64();
+    }
+}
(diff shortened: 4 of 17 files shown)


metrique-util/src/lib.rs
```diff
@@ -8,4 +8,11 @@
 #[cfg(feature = "state")]
 mod state;
 #[cfg(feature = "state")]
+#[cfg_attr(docsrs, doc(cfg(feature = "state")))]
```
jlizen: It looks like we are missing `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` from a few `lib.rs`, can we add those instead of manually annotating?

Ref: https://github.com/awslabs/metrique/commit/0f0e12a4bb9e6ba27f8c1c6956971452b1811117

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+use tokio_metrics::RuntimeMonitor;
+
+const DEFAULT_METRIC_SAMPLING_INTERVAL: Duration = Duration::from_secs(30);
+
+/// Configuration for Tokio runtime metrics bridge subscriptions.
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub struct TokioRuntimeMetricsConfig {
+    /// Sampling interval used by the reporter loop.
+    pub interval: Duration,
```
jlizen: Nit: why make this pub?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+use tokio::runtime::Handle;
+use tokio::task::JoinHandle;
+use tokio_metrics::RuntimeMonitor;
+
+const DEFAULT_METRIC_SAMPLING_INTERVAL: Duration = Duration::from_secs(30);
+
+/// Configuration for Tokio runtime metrics bridge subscriptions.
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub struct TokioRuntimeMetricsConfig {
```
jlizen, line 23: <Scope creep> 
This API to emit on an interval makes sense to me.

But, would it be worth having a different option, to fold the current runtime metrics into a given metrics struct? Some sort of `#[metrics(with_tokio_metrics)]` macro support?

Definitely not something to impl in this PR, but wondering if it merits a backlog (any opinion @rcoh ?)

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+impl TokioRuntimeMetricsConfig {
+    /// Return a config with a custom sampling interval.
+    pub fn with_interval(mut self, interval: Duration) -> Self {
+        self.interval = interval;
+        self
+    }
+}
+
+/// Extension methods for subscribing Tokio runtime metrics to a global entry sink.
+pub trait AttachGlobalEntrySinkTokioMetricsExt: AttachGlobalEntrySink + GlobalEntrySink {
```
jlizen: Probably these would benefit from some doc examples right? I guess our scraped `examples/*.rs` might show up?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+    rt: tokio::runtime::RuntimeMetrics,
+) {
+    use metrique_writer_core::value::MetricFlags;
+    use metrique_writer_core::{Entry, EntryWriter, Observation, Unit, unit::NegativeScale};
+
+    // Emitted as a separate entry alongside RuntimeMetrics because
+    // `poll_time_histogram` uses #[entry(ignore)] on RuntimeMetrics — the raw
+    // Vec<u64> counts need bucket ranges from the runtime handle to be
+    // meaningful, which Entry::write() doesn't have access to.
+    struct PollTimeHistogramEntry {
```
jlizen: We really want to keep the metrics all together rather than as separate entries. Can we not have a newtyped `PollTimeHistogramEntry` struct that we impl `CloseValue` for on the `tokio-metrics` side, which is part of the full struct?

metrique-util/Cargo.toml
```diff
@@ -8,13 +8,27 @@ description = "Additional utilities for metrique"
 repository = "https://github.com/awslabs/metrique"
 readme = "README.md"
 
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tokio_unstable)'] }
```
jlizen, line 12: We probably want to document tokio unstable unlocks more fields, right?

metrique-util/src/tokio_metrics_reporter.rs
jlizen: In general, re: discoverability, it feels like we are missing a breadcrumb somewhere, maybe in `metrique::_guide`?

metrique-util/src/tokio_metrics_reporter.rs
jlizen, line 23: This should be `#[must_use]`, right?

metrique-writer-core/src/global.rs
```diff
@@ -325,15 +336,23 @@ impl AttachHandle {
...
 
     /// Cause the attached global sink to remain attached forever.
     ///
     /// Note that this will prevent the sink from guaranteeing metric entries are flushed during
     /// shutdown. You *must* have another mechanism to ensure metrics are flushed.
-    pub fn forget(mut self) {
-        self.join = None;
+    pub fn forget(self) {
+        self.shutdown_registry.lock().unwrap().clear();
+        std::mem::forget(self);
```
jlizen: This means that `SHUTDOWN_REGISTRY` will stay alive and subsequent `register_shutdown_fn` will succeed but be non-ops. Consider also setting that to `None` in `forget()` (or dropping the stored arc before forgetting)

metrique-writer-core/src/global.rs
```diff
@@ -1050,100 +1019,123 @@ mod tests {
...
-        let _guard = ServiceMetrics::set_test_sink(sink);
-
-        ServiceMetrics::sink_or_discard().append(TestEntry);
-        assert_eq!(inspector.entries().len(), 1);
-        assert_eq!(inspector.entries()[0].metrics["BasicIntCount"], 1234);
-    }
+#[cfg(test)]
+mod shutdown_registry_tests {
+    use std::sync::Arc;
+    use std::sync::Mutex;
```
jlizen: nit: unused

metrique-util/README.md
```diff
@@ -1,10 +1,13 @@
...
+Additional utilities for [metrique](https://crates.io/crates/metrique).
 
 ## Features
 
 - `state`: Provides [`State<T>`], an atomically swappable shared value with snapshot-on-first-read semantics. Useful for shared runtime state (feature flags, config reloads, routing tables) that should appear on every metric record.
+- `tokio-metrics-bridge`: Subscribes [tokio-metrics] runtime snapshots to a global entry sink. The reporter task is automatically aborted when the `AttachHandle` is dropped.
+
+[tokio-metrics]: https://crates.io/crates/tokio-metrics
 
 ## Usage
```
jlizen, line 12: would be good to show the tokio-metrics usage in this section too

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+                let mut snapshot = snapshot;
+                let counts = std::mem::take(&mut snapshot.poll_time_histogram);
+                (snapshot, counts)
+            };
+            sink.append(snapshot);
+            #[cfg(tokio_unstable)]
+            emit_poll_time_histogram(&sink, histogram_counts, handle.metrics());
+            tokio::time::sleep(interval).await;
+        }
+        tracing::debug!("tokio runtime metrics reporter stopped");
```
jlizen, line 131: do we need to guard against the task panicking and have an error message?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+/// Extension methods for subscribing Tokio runtime metrics to a global entry sink.
+pub trait AttachGlobalEntrySinkTokioMetricsExt: AttachGlobalEntrySink + GlobalEntrySink {
+    /// Subscribe to Tokio runtime metrics, adding the subscription to this handle.
+    ///
+    /// The reporter task is automatically aborted when the [`AttachHandle`] is dropped.
+    ///
+    /// # Panics
```
jlizen: also panics if no sink has been attached before calling this, which seems likelier to trip people up

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn default() -> Self {
+        Self {
+            interval: DEFAULT_METRIC_SAMPLING_INTERVAL,
+        }
+    }
+}
+
+impl TokioRuntimeMetricsConfig {
+    /// Return a config with a custom sampling interval.
+    pub fn with_interval(mut self, interval: Duration) -> Self {
```
jlizen: nit: bit odd to have mut self here when interval is copy, I would have expected:

```
     pub fn with_interval(self, interval: Duration) -> Self {
         Self { interval, ..self }
     }
```

jlizen (commented):
> Right now nothing prevents calling subscribe_tokio_runtime_metrics twice, which would just spawn two reporter tasks appending duplicate metrics. It won't break anything, but it's almost certainly users won't intend that to happen.

Logging a warning seems fine to me for now

metrique-util/src/tokio_metrics_reporter.rs
jv1i: Agree! I meant to write a comment about this somewhere, must have forgotten. As it is now it's not very discoverable.
Also, will we want to mention/advertise these type integrations in the main readme? (same for sysinfo later, and metrics-rs)

metrique-util/src/tokio_metrics_reporter.rs
jv1i: Good idea, I wrapped the worker in a monitor task that awaits the JoinHandle and emits a tracing::error! on panic. (aborts both monitor and worker on shutdown).

metrique-writer-core/src/global.rs
jv1i: Good catch. I changed shutdown_registry to `Option<Arc<...>>` and `forget()` now sets it to `None`, same pattern as the original `join: Option<fn()>`. `register_shutdown_fn` now panics if used after `forget()`.

metrique-util/src/tokio_metrics_reporter.rs
jv1i: Ok I've added some doc examples and breadcrumb in the sinks guide (metrique::_guide::sinks), not sure if that's the best place for it or if we want a separate section.

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// Wrapper that roots an [`InflectableEntry`] into an [`Entry`] for `sink.append()`.
+struct RootedEntry<M: InflectableEntry>(M);
+
+impl<M: InflectableEntry> Entry for RootedEntry<M> {
+    fn write<'a>(&'a self, w: &mut impl EntryWriter<'a>) {
+        self.0.write(w);
+    }
+}
```
jv1i: Doing this and calling `snapshot.close()` is a workaround for not having access to `RootEntry` in metrique-util. If this is not ideal, perhaps we could move RootEntry down to `metrique-core` or `metrique-writer-core`?

yulnr:
Following up on https://github.com/tokio-rs/tokio-metrics/pull/121#discussion_r3016431358

With tokio-metrics using `#[metrics(subfield)]` without `rename_all`, I'm looking into how to let callers pick a name style.

`InflectableEntry` is generic over `NS: NameStyle`, so to expose this to callers we have some options:

**Generic method**
```rust
// Default (Identity — snake_case):
ServiceMetrics::subscribe_tokio_runtime_metrics(config);

// PascalCase:
ServiceMetrics::subscribe_tokio_runtime_metrics::<PascalCase>(config);
```

**Generic config**
```rust
let config = TokioRuntimeMetricsConfig::default()
    .with_name_style::<PascalCase>();
```
Config becomes `TokioRuntimeMetricsConfig<PascalCase>`

Now If we introduce a new public enum we could also do:

**Runtime enum on config**
```rust
let config = TokioRuntimeMetricsConfig::default()
    .with_name_style(MetricNameStyle::PascalCase);
```
(dispatches via `match` internally)
I think much better for users, only downside is we'd have a mirror of `NameStyle` enum from metrique-macro, but that should be easy to keep in sync.

**If we support folding via #[metrics] with automatic appending**

[Here](https://github.com/awslabs/metrique/pull/256#discussion_r2989175858) @jlizen mentioned some sort of macro support for folding RuntimeMetrics into another metrics struct (I'm assuming that would automatically pull in the RuntimeMetrics data). If we go that route then we could solve this in multiple ways at the struct level rather than subscription config level, one example: `#[metrics(rename_all = "PascalCase")]`

What already works as this stands now: since RuntimeMetrics is a subfield, users can flatten it into their own `#[metrics(rename_all = "PascalCase", with_tokio_metrics)]` struct and the naming is inherited. Though this is less likely to be used that way because users would need to write the monitor loop and append to sink themselves.

Do we have any preferences here, or other approaches I'm missing?

cc @rcoh

rcoh:
I think I have a slight preference for the enum. No issues with mirroring it. We could eventually have it set on the sink too for runtime-cased metrics.

Overall approach seems good to me

metrique-util/src/tokio_metrics_reporter.rs
rcoh: Can metrique-util depend on metrique?

jv1i: Actually yes, idk why I had the impression we might not want do that. 
I switched to metrique directly now. Also actually I could remove that RootedEntry now, since either way we have our own adapter to apply name style inflection. So I extracted that into a shared `DynamicInflectionEntry` in its own module (in metrique-util) so that sysinfo and other bridges can reuse it

metrique-util/src/dynamic_inflection.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use metrique::InflectableEntry;
+use metrique::writer::{Entry, EntryWriter};
+use metrique_core::{DynamicNameStyle, Identity, KebabCase, PascalCase, SnakeCase};
+
+use crate::MetricNameStyle;
+
+/// Adapter that bridges [`InflectableEntry`] to [`Entry`] by selecting the
```
rcoh, line 12: are there tests of how this interacts with prefixing?

metrique-util/src/tokio_metrics_reporter.rs
rcoh: I think the cleanest thing would be to have `EmbeddedTokioMetrics` that had a `::default()` impl so you could embed it as a field (rather than needing to drop into having macro support)

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Spawns a background task that periodically samples
+/// [`RuntimeMetrics`](tokio_metrics::RuntimeMetrics) and appends each snapshot to the sink.
+/// The task is automatically aborted when the [`AttachHandle`](metrique::writer::sink::AttachHandle) is dropped.
+///
+/// # `tokio_unstable`
+///
+/// When the runtime is built with `RUSTFLAGS="--cfg tokio_unstable"` and
+/// `enable_metrics_poll_time_histogram` is called on the runtime builder, each
+/// snapshot also includes a `poll_time_histogram` entry emitted as a distribution
+/// metric with bucket ranges from the runtime handle.
```
rcoh: I assume this works with and without tokio_unstable but you get a different set of metrics? we should be really clear about that

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn subscribe_tokio_runtime_metrics(config: TokioRuntimeMetricsConfig) {
+        // Guard against duplicate subscriptions within the same attach cycle.
+        // Reset in the shutdown fn so re-attaching can subscribe again cleanly.
+        static SUBSCRIBED: AtomicBool = AtomicBool::new(false);
+        if SUBSCRIBED.swap(true, Ordering::Relaxed) {
+            tracing::warn!(
+                "subscribe_tokio_runtime_metrics called more than once; \
+                 duplicate reporter task will emit duplicate metrics"
+            );
+        }
```
rcoh: why are we allowing this instead of ignoring it?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+
+impl<T: AttachGlobalEntrySink + GlobalEntrySink> AttachGlobalEntrySinkTokioMetricsExt for T {}
+
+fn spawn_tokio_runtime_metrics_task(
+    sink: BoxEntrySink,
+    config: TokioRuntimeMetricsConfig,
+) -> (tokio::task::AbortHandle, JoinHandle<()>) {
+    let interval = config.interval;
+    let name_style = config.name_style;
+    let worker = tokio::spawn(async move {
```
rcoh, line 120: we will eventually (but can probably punt) need to provide an overridable executor

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+        let entries = inspector.entries();
+        assert!(!entries.is_empty(), "expected entries");
+
+        let entry = &entries[0];
+        assert_eq!(entry.metrics["WorkersCount"], 1);
+        entry.metrics["TotalParkCount"].as_u64();
+        assert!(entry.metrics["Elapsed"] > 0.0);
+
+        #[cfg(tokio_unstable)]
+        assert!(entry.metrics["PollTimeHistogram"].num_observations() > 0);
```
rcoh: using `assert2::check!` will give nicer errors here

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+        );
+
+        tokio::time::sleep(Duration::from_millis(200)).await;
+
+        let entries = inspector.entries();
+        assert!(!entries.is_empty(), "expected entries");
+
+        let entry = &entries[0];
+        assert_eq!(entry.metrics["WorkersCount"], 1);
+        entry.metrics["TotalParkCount"].as_u64();
```
rcoh: this isn't really testing anything? I guess you are just trying to assert that the field is present and that it is a number?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+        // Guard against duplicate subscriptions within the same attach cycle.
+        // Reset in the shutdown fn so re-attaching can subscribe again cleanly.
+        static SUBSCRIBED: AtomicBool = AtomicBool::new(false);
+        if SUBSCRIBED.swap(true, Ordering::Relaxed) {
+            tracing::warn!(
+                "subscribe_tokio_runtime_metrics called more than once; \
+                 duplicate reporter task will emit duplicate metrics"
+            );
+        }
+        let sink = Self::sink();
```
rcoh: could we use the lazy_sink abstraction instead? then no panics.

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// The reporter task is automatically aborted when the [`AttachHandle`](metrique::writer::sink::AttachHandle) is dropped.
+    /// If the handle is [`forgotten`](metrique::writer::sink::AttachHandle::forget), the reporter runs indefinitely.
+    ///
+    /// # Panics
+    /// Panics if no sink has been attached yet, or if the underlying sink has been
+    /// detached (e.g. the `AttachHandle` was dropped or forgotten before this call).
+    fn subscribe_tokio_runtime_metrics(config: TokioRuntimeMetricsConfig) {
+        // Guard against duplicate subscriptions within the same attach cycle.
+        // Reset in the shutdown fn so re-attaching can subscribe again cleanly.
+        static SUBSCRIBED: AtomicBool = AtomicBool::new(false);
```
rcoh: This is actually shared across _all_ implementations (its not actually local to the monomorphized implementor). Honestly, I'd probably just drop validation step. It doesn't seem too critical in practice.

metrique-writer-core/src/global.rs
```diff
@@ -240,9 +247,17 @@ pub trait AttachGlobalEntrySink {
...
+    /// Registry of shutdown functions to call when the attach handle is dropped.
+    /// `None` after `forget()` is called.
+    shutdown_registry: Option<Arc<ShutdownRegistry>>,
 }
 
+#[doc(hidden)]
+pub type ShutdownFn = Box<dyn FnOnce() + Send>;
+
+#[doc(hidden)]
+pub type ShutdownRegistry = Mutex<Vec<ShutdownFn>>;
```
rcoh: what are we attempting to achieve by making these doc hidden? `register_shutdown_fn` is _not_ doc hidden an in practice, we would't be able to change these types. Probably worth just making newtypes for them for backwards compat and not trying to hide them in the API.

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -0,0 +1,281 @@
...
+            sink.append(DynamicInflectionEntry {
+                entry: snapshot.close(),
+                name_style,
+            });
+            tokio::time::sleep(interval).await;
+        }
+        tracing::debug!("tokio runtime metrics reporter stopped");
+    });
+    let worker_abort = worker.abort_handle();
+    let monitor = tokio::spawn(async move {
```
rcoh: do we actually need two separate tasks here? (Maybe?) Is this a general pattern that all bridge-reporters are going to need?

rcoh (commented):
Nice!

metrique-util/src/tokio_metrics_reporter.rs
jv1i: oh good catch, I'll do that

metrique-util/src/dynamic_inflection.rs
jv1i: I'll add some now, and perhaps others on top of the EmbeddedTokioMetrics approach you suggested (might finish this on Monday)

metrique-util/src/tokio_metrics_reporter.rs
jv1i: Good idea, would you be ok with doing this as a follow up? I can follow up with some work that does it for metrique in general, we do some hardcoded spawning in metrics-rs integration and in some of the internals as well. 
I'm also curious as to how to catch this on CI, tests or even dial9 warnings, I have some ideas I'd like to try out later

metrique-util/src/tokio_metrics_reporter.rs
jv1i: we can do that as well, I mentioned on the PR description some options for this and Jess agreed on warning, but I didn't think of ignoring entirely, I'm updating to just skip this as you mentioned in another comment

metrique-util/src/tokio_metrics_reporter.rs
jv1i: adding some better docs for that

metrique-util/src/tokio_metrics_reporter.rs
jv1i: yeah I can be clearer about that, I thought doing > 0 was also a bit redundant since that should be always true for unsigned, but that'd look more intentional at least. also maybe I can do something similar to what tokio-metrics and force and check and actual count

Cargo.toml
```diff
@@ -67,3 +67,12 @@ tracing-subscriber = "0.3.20"
 trybuild = "1.0"
 toml = "0.9"
 walkdir = "2"
+
+[patch.crates-io]
```
rcoh, line 74: Still needed?

rcoh (approved):

metrique-util/src/tokio_metrics_reporter.rs
jv1i: I added this earlier when Jess asked about guarding against task panics and having an error. Is not truly needed contract-wise, I now simplified to use a single abort handle so we only pass one in the abort closure. Other bridges could opt not to do this, but perhaps it's preferred that we catch these and use tracing to log errors rather than having a single task panic and go into stderr?

Cargo.toml
jv1i: I'm looking into this, because without it I get compilation errors:, note the `there are multiple different versions of crate metrique_core in the dependency graph`:
```rust
error[E0599]: no method named `close` found for struct `tokio_metrics::RuntimeMetrics` in the current scope
   --> metrique-util/src/tokio_metrics_reporter.rs:126:33
    |
126 |                 entry: snapshot.close(),
    |                                 ^^^^^
    |
note: there are multiple different versions of crate `metrique_core` in the dependency graph
   --> /Users/joulei/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/metrique-core-0.1.18/src/lib.rs:131:1
    |
131 | pub trait CloseValue {
    | ^^^^^^^^^^^^^^^^^^^^ this is the expected trait
    |
   ::: metrique-core/src/lib.rs:131:1
    |
131 | pub trait CloseValue {
    | -------------------- this is the trait that was imported
    = help: you can use `cargo tree` to explore your dependency tree
help: some of the expressions' fields have a method of the same name
```
cargo is pulling metrique-core from crates.io for tokio-metrics, and from path for metrique.

So at least we'll need it for local development until release. But I'm not sure yet if once this is released in crates.io, cargo will know how to unify the crates.io metrique deps from tokio-metrics, with the path deps that metrique has between crates. I'm trying to find docs on this or what others libs are doing (both serde and tokio have similar patches)

rcoh: Got it ok makes sense. Just add a comment

metrique-writer-core/src/global.rs
```diff
@@ -240,7 +247,47 @@ pub trait AttachGlobalEntrySink {
...
+
+impl ShutdownRegistry {
+    fn new(initial: ShutdownFn) -> Self {
+        Self(Mutex::new(vec![initial]))
+    }
+
+    /// Prepend a shutdown function so it runs before earlier ones (LIFO order).
+    #[doc(hidden)]
+    pub fn prepend(&self, f: ShutdownFn) {
+        self.0.lock().unwrap().insert(0, f);
```
rcoh: Nit: we could just append and drain backwards. Doesn't really matter though

metrique-writer-core/src/global.rs
```diff
@@ -240,7 +247,47 @@ pub trait AttachGlobalEntrySink {
...
+    }
+
+    pub(crate) fn drain_and_run(&self) {
```
rcoh: Should this consume self?

rcoh (approved):

metrique-writer-core/src/global.rs
jv1i: Made it do so now, needed Arc::try_unwrap in Drop since the registry lives behind an Arc

rcoh (approved):

PR #257, opened 2026-03-26
Derive debug through metrics macro

### LINKED ISSUE #250: project `#[derive(Debug)]` through `#[metrics(value(string)]`
The following code does not properly extend `Debug` onto the `OperationValue` generated struct:

```rust
#[metrics(value(string))]
#[derive(Debug)]
pub(crate) enum Operation {
    Flush,
    ProcessSegment,
}
```

this makes it annoying because derive(Debug) then stops working on metrics structs that contain this.

Fluzko:
closes #250 

## Summary
This PR aims to fix a wrong implementation of `derive`s for `Clone`, `Copy` and `Debug` for enums (and their closed versions [EnumName]Value)

After this work, derives that come **after** `#[metrics(value(string))]` will be correctly implemented and passed through if it corresponds.

Note:
There's a case where our metrics macro can't do much:
```rust
#[derive(Debug)]
#[metrics(value(string))] // << conflicting implementations
enum Operation {...}
```
The example above will cause a "Conflicting implementations" compile error, since we can't really catch derived/implemented macros that came **before** our macro. The rest of the cases should work as expected, and the error seems clearer now. Also added some docs around this.

For other macros, like the example below, this will be gracefully handled:
```rust
#[derive(Eq, Partial)]
#[metrics(value(string))]
enum Operation {...}

// Is the same as:

#[metrics(value(string))]
#[derive(Eq, Partial)]
enum Operation {...}
```


jlizen:
Catching up on #250 - Is there a reason that we opted to auto derive debug and clone? It feels misleading especially given the conflicting impl errors if the derive is out of macro scope. And if people care a lot about binary size, unwanted codegen isn't ideal.

I would have assumed that what we want is to instead respect the derives if they come before the metrics (though not sure how urgent that is compared to doc improvements).

(Cc @rcoh too)

Fluzko:
@jlizen, as per our discussion with @rcoh, I understood (I might be wrong) that we wanted those values to be auto-derived... but you got a point about the binary size. The discussion around adding those auto derives was because of their safety around the places where the `value(string)` can be called, they are always enums.

rcoh:
Binary size on derive(Debug) for value(string) isn't super relevant. 

I mentioned this because for value(string) it will just silently not imply debug on the value type if you put the derives in the wrong order (which is pretty annoying!) and it tripped up both me and AI.

So for value(string) there really isn't a good reason not to have debug.

I agree about the potential for a breaking change though

jlizen:
Yeah, the missing value impls based on ordering is pretty darn confusing, and no good mitigation unless we start having module level macro scope.

I think I'm sold on the auto derive on the value impl, but why do we need it on the original enum? That is likely the source of the conflicting impl right? Can't we sidestep that by forcing caller to specify explicitly, since it is locally obvious what is wrong if the required impl is missing? That makes all the autogenerated code do the right thing, without breakage risk (unless they manually implemented on the codegenned value type, which is kind of ridiculous)

rcoh:
Yeah precisely, the behavior we need is that the *Value type always gets the debug derive.

metrique-util/tests/state.rs
```diff
@@ -17,7 +17,7 @@ struct AppConfig {
     traffic_policy: TrafficPolicy,
 }
 
-#[derive(Clone, Copy, Debug, Default)]
+#[derive(Default)]
```
rcoh: This needs to work (without deleting it this)

metrique-macro/src/snapshots/metrique_macro__tests__simple_metrics_enum.snap
```diff
@@ -2,10 +2,12 @@
 source: metrique-macro/src/lib.rs
 expression: parsed_file
 ---
+#[derive(Debug, Clone, Copy)]
 enum Foo {
     Bar,
 }
 #[doc(hidden)]
+#[derive(Debug, Clone, Copy)]
```
rcoh, line 9: The end result is the Value type gets these derives (+ any others that are specified) but the user defined type gets exactly what they describe

rcoh (changes requested):

Fluzko:
@rcoh I think that with these changes, we're at the point we've discussed

metrique-util/tests/state.rs
```diff
@@ -17,7 +17,7 @@ struct AppConfig {
     traffic_policy: TrafficPolicy,
 }
 
-#[derive(Clone, Copy, Debug, Default)]
```
rcoh, line 20: was dropping this required?

rcoh (approved):

Issue #258, opened 2026-03-26
Support conditional metrics(...) inside cfg_attr(...)


yulnr:
## Summary

Metrique currently has an incomplete story for `metrics(...)` options nested inside `cfg_attr(...)` on fields/variants.

The integration work with `tokio-metrics` surfaced this. We needed conditional field behavior for `#[metrics]`, and found that currently, `metrics(...)` nested in `cfg_attr(...)` is not reliably treated as a `metrics` attribute input.

## How I ran into this

While integrating `#[metrics]` with `tokio-metrics` (inside a `macro_rules!`-generated struct), I used a conditional metrique attribute like:
```rust
#[cfg_attr(feature = "metrique-integration", metrics(ignore))]
```

This exposed two separate concerns:

1. **Attribute visibility/parsing**: `darling` parsing does not automatically see `metrics(...)` nested in `cfg_attr(...)`.
2. **Conditional semantics**: the above problem is easy to solve, but after exposing the nested `metrics(...)` for parsing, I realized that is not enough: we need to the correct evaluation for the `cfg_attr` condition, and as I understand it is not a straight forward problem to solve inside proc macros. 

Further down I linked a discussion from `bon` regarding solutions, at some point it references this code which explains the core issue: https://github.com/elastio/bon/blob/3217b4b0349f03f0b2a5853310f420c5b8b005a7/bon/src/private/mod.rs#L90-L112

## Test that demonstrates this

This it the main behavior we should support (or clearly reject/document):
```rust
#[metrics]
struct Example {
    foo: u64,
    // any() evals to false, `bar` should be ignored
    #[cfg_attr(any(), metrics(ignore))]
    bar: u64,
}

#[test]
fn cfg_attr_nested_metrics_not_applied_when_condition_false() {
    let m = Example { foo: 10, bar: 20 };
    let entry = test_util::to_test_entry(RootEntry::new(m.close()));
    assert!(entry.metrics.contains_key("foo"));
    assert!(!entry.metrics.contains_key("bar")); // <----- this assertion fails, bar doesn't get ignored
}
```
Or, said another way, the macro expands to:
```rust
pub struct ExampleEntry {
    #[deprecated(
        note = "these fields will become private in a future release. To introspect an entry, use `metrique::writer::test_util::test_entry`"
    )]
    #[doc(hidden)]
    foo: <u64 as metrique::CloseValue>::Closed,
    #[deprecated(
[...]

PR #259, opened 2026-03-26
fix(metrics): make metrics macro inside macro_rules more hygiene-safe and cfg-aware

yulnr:
## Summary

This updates `metrique-macro` so `#[metrics]` works on `macro_rules!`-generated structs (like `tokio-metrics::RuntimeMetrics`).

Main fixes:
- hygiene mismatches in generated impl bodies
- cfg propagation so generated code does not reference cfg-disabled fields

## Context

`tokio-metrics` defines `RuntimeMetrics` via `macro_rules!` and has fields gated by cfg
(for example, unstable metrics fields).  See: https://github.com/tokio-rs/tokio-metrics/blob/de140d346ad8f8994a900ddcbdb4d100cb710ec2/src/runtime.rs#L55-L89

Using `#[metrics]` as-is, I was hitting issues like:

- `cannot find value 'writer' in this scope`
- `expected value, found module 'self'`
- `no field '...' on type ...` when generated code referenced cfg-disabled fields

## What this changes

1. Hygiene fixes in generated impls
   - Use mixed-site internal idents for generated locals
   - Rebind receiver in generated method bodies and use that binding for generated field accesses

2. Propagate field cfg/cfg_attr to generated code paths
   - Apply field cfg attrs to generated entry fields, close initializers, write paths, and sample-group code
   - Sample-group generation now uses cfg-safe wrapping so cfg-disabled fields are not referenced

## Notes for `tokio-metrics`

Current direction in https://github.com/tokio-rs/tokio-metrics/pull/121 is to improve poll histogram representation directly in `tokio-metrics`, so field-level `#[cfg_attr(feature = "metrique-integration", metrics(ignore))]` is no longer required.
However another problem I found originally is that `metrics(...)` inside `#[cfg_attr(...)]` does not properly work either, so I'm opening an issue to track this here: https://github.com/awslabs/metrique/issues/258

## Testing

Added `metrics-inside-macro-rules.rs` with some tests, and updated snaps. Compilation passes for tokio-metrics with:
```rust
#[cfg_attr(feature = "metrique-integration", metrique::unit_of_work::metrics(subfield, rename_all = "PascalCase"))]
pub struct RuntimeMetrics {
```
(if we ignore that we don't have an impl Value for `poll_time_histogram: Vec<u64>`)

---

### DIFF
--- a/metrique/tests/ui/fail/bad_metrics_value.stderr
+++ b/metrique/tests/ui/fail/bad_metrics_value.stderr
@@ -131,24 +131,6 @@ error[E0277]: `FormattedValue<'_, u32, Foo, _>` is not a metric value
            and $N others
 
 error[E0277]: CloseValue is not implemented for &String
-  --> tests/ui/fail/bad_metrics_value.rs:76:1
-   |
-76 | #[metrics(subfield)]
-   | ^^^^^^^^^^^^^^^^^^^^ This type must implement `CloseValue`
-77 | struct WithString {
-78 |     field: String,
-   |     ----- required by a bound introduced by this call
-   |
-   = help: the trait `CloseValue` is not implemented for `&String`
-   = note: You may need to add `#[metrics]` to `&String` or implement `CloseValue` directly.
-   = note: if &String implements `Value` but not `CloseValue`, add `#[metrics(no_close)]`
-   = note: If this type is `&T`, is closed inside a flattened entry, and `T` implements `CloseValue`, consider using `#[metrics(subfield_owned)]`.
-help: consider dereferencing here
-   |
-78 |     *field: String,
-   |     +
-
-error[E0277]: CloseValue is not implemented for &String
   --> tests/ui/fail/bad_metrics_value.rs:78:5
    |
 78 |     field: String,
@@ -158,4 +140,7 @@ error[E0277]: CloseValue is not implemented for &String
    = note: You may need to add `#[metrics]` to `&String` or implement `CloseValue` directly.
    = note: if &String implements `Value` but not `CloseValue`, add `#[metrics(no_close)]`
    = note: If this type is `&T`, is closed inside a flattened entry, and `T` implements `CloseValue`, consider using `#[metrics(subfield_owned)]`.
-   = help: the trait `CloseValue` is implemented for `String`
+help: consider dereferencing here
+   |
+78 |     *field: String,
+   |     +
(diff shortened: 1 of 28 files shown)


metrique/tests/ui/fail/bad_metrics_value.stderr
```diff
@@ -131,16 +131,31 @@ error[E0277]: `FormattedValue<'_, u32, Foo, _>` is not a metric value
            and $N others
 
 error[E0277]: CloseValue is not implemented for &String
-  --> tests/ui/fail/bad_metrics_value.rs:78:5
+  --> tests/ui/fail/bad_metrics_value.rs:76:1
    |
+76 | #[metrics(subfield)]
```
rcoh: this appears to be a slight regression in diagnostics. not the end of the world

metrique/tests/ui/fail/enum_kitchen_sink.stderr
```diff
@@ -89,19 +104,34 @@ error[E0277]: CloseValue is not implemented for &TimestampOnClose
...
    |
    = help: the trait `CloseValue` is not implemented for `&String`
    = note: You may need to add `#[metrics]` to `&String` or implement `CloseValue` directly.
    = note: if &String implements `Value` but not `CloseValue`, add `#[metrics(no_close)]`
    = note: If this type is `&T`, is closed inside a flattened entry, and `T` implements `CloseValue`, consider using `#[metrics(subfield_owned)]`.
 help: consider dereferencing here
    |
-96 |     *value: String,
-   |     +
+94 | *#[metrics(subfield)]
```
rcoh: it looks like we are emitting the wrong span since it thinks this is `*#[metrics(subfield)]` and not `*value`

rcoh (commented):
looks like a small regression in diagnostics, probably caused by an incorrect span somewhere

metrique/tests/ui/fail/enum_kitchen_sink.stderr
jv1i: Fixed this one since I had a `quote_spanned!` that we didn't need anymore, I'll see if I can fix the other ones (the duplication and line number changes) or I'll update the .stderrs otherwise (didn't push them yet that's the reason for the build failure in CI)

jv1i: Alright after trying a lot of things I found a workaround that prevents these regressions. 
It's a bit hacky but inoffensive I think. For the CloseValue path I'm replacing `self` rebinding with a local macro token (`__metrique_self_expr!()`) used only inside the generated close() bodies. This seems to keep the macro_rules hygiene without the binding that was causing the line number drift and the references to `#[metrique(subfield)]`

see [b6a6a63](https://github.com/awslabs/metrique/pull/259/commits/b6a6a63a8f44343a8b998d5b6f2eb2723b0c8c5d)
(in the overall PR the stdrr files stay the same as current main)

rcoh (approved):

PR #260, opened 2026-03-30
fix(metrics): suppress dead_code and unreachable_pub on macro-generated items

yulnr:
## Summary

The `#[metrics]` macro generates type aliases (`*Guard`, `*Handle`)
and an `append_on_drop` method for every metrics struct. Consumers may never use or re-export them.

Crates that compile with `-D warnings` (like tokio-metrics) get hard errors from `dead_code` and `unreachable_pub` on these generated items.
                                                                                                                                                               
This adds `#[allow(dead_code, unreachable_pub)]` to the generated Guard, Handle, and append_on_drop items.  
  

yulnr:
I wont actually need this now for tokio-metrics since we're switching to `subfield` which won't generate the unused code. I'll open an issue to track this separately with a better solution.

PR #265, opened 2026-04-10
feat(core): add OwnedCounterGuard for cross-boundary tracking

### LINKED ISSUE #264: metrique: `Counter::increment_scoped` should support an owned guard variant
`Counter::increment_scoped` returns `CounterGuard<'a>` which borrows `&'a AtomicU64`. This works for same-scope usage but not when the guard needs to outlive the borrow, e.g. when tracking in-flight HTTP requests across a response body wrapper struct.

Pattern:
```rust
// Middleware increments on request arrival
let (guard, _) = counter.increment_scoped();
// Guard needs to live inside a response body wrapper that outlives the async block
let body = MetricsBody { inner: response_body, _guard: guard }; // can't: guard borrows counter
```

With `Arc<Counter>`, an owned guard could decrement on drop without a lifetime tie:

```rust
pub struct OwnedCounterGuard(Arc<Counter>);
impl Drop for OwnedCounterGuard {
    fn drop(&mut self) { /* fetch_sub */ }
}

impl Counter {
    pub fn increment_scoped_owned(self: &Arc<Self>) -> (OwnedCounterGuard, u64) { ... }
}
```

Without this, users fall back to manual `fetch_add`/`fetch_sub` on raw `AtomicU64`, losing the RAII safety that `increment_scoped` provides.

jlizen:
📬 *Issue #, if available:*

Closes #264

✍️ *Description of changes:*

### Summary

`CounterGuard` borrows `&'a AtomicU64`, which prevents it from being stored in structs that outlive the borrow (e.g. response body wrappers in tower middleware). Users currently fall back to manual `fetch_add`/`fetch_sub`, losing RAII safety.

This adds `OwnedCounterGuard`, which holds an `Arc<Counter>` instead of a reference. It has the same `Drop` behavior (saturating decrement) and the same `CloseValue` impls as `CounterGuard`, but can be freely moved across async boundaries.
a📬 *Issue #, if available:*

Closes #264

✍️ *Description of changes:*

### Summary

`CounterGuard` borrows `&'a AtomicU64`, which prevents it from being stored in structs that outlive the borrow (e.g. response body wrappers in tower middleware). Users currently fall back to manual `fetch_add`/`fetch_sub`, losing RAII safety.

This adds `OwnedCounterGuard`, which internally shares the counter and can be freely moved across async boundaries. It has the same `Drop` behavior (saturating decrement) and the same `CloseValue` impls as `CounterGuard`.

The guard is `Clone`. Cloning shares the underlying counter without incrementing it; each clone independently decrements on drop. Callers are responsible for ensuring the counter has been incremented a corresponding number of times.

The constructor is `Counter::increment_scoped_owned(self: &Arc<Self>)`, mirroring the existing `Counter::increment_scoped(&self)` API.

`OwnedCounterGuard` is re-exported from both `metrique-core` and the top-level `metrique` crate.

### Testing

Unit tests covering increment/drop semantics, saturation at zero, `CloseValue` behavior, cross-thread movement, and multiple concurrent guards on the same counter.

The constructor is `Counter::increment_scoped_owned(self: &Arc<Self>)`, mirroring the existing `Counter::increment_scoped(&self)` API. The caller wraps their `Counter` in an `Arc`, which is the natural choice for shared state in async contexts.

`OwnedCounterGuard` is re-exported from both `metrique-core` and the top-level `metrique` crate.

### Testing

Unit tests covering increment/drop semantics, saturation at zero, `CloseValue` behavior, cross-thread movement, and multiple concurrent guards on the same counter.

### DIFF
--- a/metrique-core/src/atomics.rs
+++ b/metrique-core/src/atomics.rs
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
+
 use core::sync::atomic::{AtomicBool, AtomicU8, AtomicU16, AtomicU32, AtomicU64, AtomicUsize};
 
 use crate::CloseValue;
@@ -38,11 +40,28 @@ impl Counter {
     pub fn set(&self, i: u64) {
         self.0.store(i, std::sync::atomic::Ordering::SeqCst);
     }
+
+    /// Increments the count by 1, returning an owned guard that decrements the
+    /// count on drop, and the new value.
+    ///
+    /// Unlike [`increment_scoped`](Self::increment_scoped), the returned
+    /// [`OwnedCounterGuard`] can be moved across async boundaries or stored
+    /// in structs without lifetime constraints.
+    pub fn increment_owned(self: &Arc<Self>) -> (OwnedCounterGuard, u64) {
+        let count = self.0.fetch_add(1, std::sync::atomic::Ordering::Relaxed) + 1;
+        (
+            OwnedCounterGuard {
+                counter: Arc::clone(self),
+            },
+            count,
+        )
+    }
 }
 
 /// A guard that decrements a [`Counter`] when dropped.
 ///
 /// Returned by [`Counter::increment_scoped`].
+#[must_use]
 pub struct CounterGuard<'a>(&'a AtomicU64);
 
 impl Drop for CounterGuard<'_> {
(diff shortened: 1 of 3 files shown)


metrique-core/src/atomics.rs
```diff
@@ -38,6 +40,20 @@ impl Counter {
...
     }
+
+    /// Increments the count by 1, returning an owned guard that decrements the
+    /// count on drop, and the new value.
+    ///
+    /// Unlike [`increment_scoped`](Self::increment_scoped), the returned
+    /// [`OwnedCounterGuard`] can be moved across async boundaries or stored
+    /// in structs without lifetime constraints.
+    ///
+    /// The guard is [`Clone`]. Cloning does not increment the counter, but
```
rcoh: This seems wrong - I don't think the guard should be clone. Or if it is, the last one to drop should decrement.

At the very least these are confusing docs.

metrique-core/src/atomics.rs
```diff
@@ -38,6 +40,20 @@ impl Counter {
...
+    /// each clone will independently decrement on drop.
+    pub fn increment_scoped_owned(self: &Arc<Self>) -> (OwnedCounterGuard, u64) {
```
rcoh: This isn't scoped anymore, just increment_owned?

metrique-core/src/atomics.rs
```diff
@@ -75,6 +91,55 @@ impl CloseValue for CounterGuard<'_> {
...
+///
+/// Unlike [`CounterGuard`], this guard can be moved across async boundaries
+/// or stored in structs without lifetime constraints.
+///
+/// Cloning a guard shares the underlying counter without incrementing it.
+/// Each clone will independently decrement on drop, so callers should
+/// ensure the counter has been incremented a corresponding number of times.
+///
+/// Returned by [`Counter::increment_scoped_owned`].
+pub struct OwnedCounterGuard(Arc<Counter>);
```
rcoh: Must_use

metrique-core/src/atomics.rs
```diff
@@ -75,6 +91,55 @@ impl CloseValue for CounterGuard<'_> {
...
+        Self(Arc::clone(&self.0))
+    }
+}
+
+impl Drop for OwnedCounterGuard {
+    fn drop(&mut self) {
+        self.0
+            .0
+            .fetch_update(
+                std::sync::atomic::Ordering::Relaxed,
```
rcoh, line 113: This seems wrong

metrique-core/src/atomics.rs
```diff
@@ -151,4 +218,80 @@ mod tests {
...
+        assert_eq!(counter.0.load(std::sync::atomic::Ordering::Relaxed), 1);
+        drop(g2);
+        assert_eq!(counter.0.load(std::sync::atomic::Ordering::Relaxed), 0);
+    }
+
+    #[test]
+    fn owned_counter_guard_clone_decrements_independently() {
+        let counter = Arc::new(Counter::new(0));
+        let (guard, _) = counter.increment_scoped_owned();
+        // Manually bump so there's headroom for two decrements.
```
rcoh: Yeah this clearly doesn't make sense. It would go to -2 otherwise

rcoh (changes requested):

metrique-core/src/atomics.rs
jlizen: i can name it

rcoh (approved):

Issue #233, opened 2026-03-14
Add `Value` impl for `Vec<V: Value>` (property lists)


jlizen:
## Motivation

We have non-metric data that is naturally a list of strings (e.g. `["cargo", "spm"]` for active plugins). We want to emit this as a structured property so it is easy to query in telemetry backends.

Today, `ValueWriter` only supports `string()` and `metric()`. There is no way to emit structured non-metric properties like arrays or objects. The workaround is to emit a comma-separated string or other custom serialized wrapper via a `ValueFormatter`:

```rust
struct CommaSeparated;
impl ValueFormatter<Vec<String>, NotLifted> for CommaSeparated {
    fn format_value(writer: impl ValueWriter, value: &Vec<String>) {
        writer.string(&value.join(","));
    }
}

#[derive(Metrics)]
struct MyMetrics {
    #[metrics(format = CommaSeparated)]
    active_plugins: Vec<String>,
}
```

## Proposal

Add a blanket `impl<V: Value> Value for Vec<V>`, backed by a new `values()` method on `ValueWriter`:

```rust
fn values<'a, V: Value + 'a>(self, values: impl IntoIterator<Item = &'a V>);
```

This follows the same `IntoIterator` pattern as the existing `metric()` method. Each element writes itself through the existing `string()`/`metric()` path. These are flat (non-recursive) collections of primitive `Value` types.

This is safe to add: orphan rules prevent downstream crates from implementing `Value` for `Vec<T>`, so there are no conflicting impls. And since `Vec<String>` doesn't implement `Value` today, nobody can be using it in a metrics struct.

With this, `Vec<String>` just works as a field type:

```rust
let emf = Emf::all_validations("MyNamespace".to_string(), vec![vec![]]);

#[derive(Metrics)]
struct MyMetrics {
    active_plugins: Vec<String>,
}

metrics.active_plugins = vec!["cargo".into(), "spm".into()];
// EMF output: "active_plugins": ["cargo","spm"]
```

### Default implementation for `values()`

The default impl joins elements with commas and calls `self.string()`. This is a reasonable representation for line-based formats and matches what you'd do manually today with a pre-joined string.

### Existing format support

[...]

PR #266, opened 2026-04-10
 feat(core): add Value impl for Vec<V> with native EMF/json array support 

jlizen:
📬 *Issue #, if available:* #233

✍️ *Description of changes:*

Adds `Value` impl for `Vec<V: Value>` and `[V: Value]` so vector properties emit as native JSON arrays in EMF and comma-joined strings in other formats.

Adds a `values()` method to `ValueWriter` with a default comma-join implementation. EMF overrides it to write a JSON array. Elements that write nothing (e.g. `None` in a `Vec<Option<String>>`) are skipped in both paths. Also adds `CloseValue` for `Vec<V>` so it works with the `#[metrics]` derive macro.

Tested with new integration tests covering EMF array output, default comma-join behavior, empty vecs, and None-element skipping.

### DIFF
--- a/metrique-writer-core/src/value/mod.rs
+++ b/metrique-writer-core/src/value/mod.rs
@@ -85,19 +85,19 @@ pub trait ValueWriter: Sized {
     }
 
     /// Write a list of values. Formats that support native arrays (e.g. EMF) can override this
-    /// to emit a structured representation. The default joins each element's string representation
-    /// with `", "`, skipping elements that write nothing (e.g. `None`).
+    /// to emit a structured representation. The default comma-joins each element's string
+    /// representation, skipping elements that write nothing (e.g. `None`).
     fn values<'a, V: Value + 'a>(self, values: impl IntoIterator<Item = &'a V>) {
         let mut buf = String::new();
-        let mut elem = String::new();
         for value in values {
-            elem.clear();
-            value.write(StringCapture(&mut elem));
-            if !elem.is_empty() {
-                if !buf.is_empty() {
-                    buf.push_str(", ");
-                }
-                buf.push_str(&elem);
+            let before = buf.len();
+            if !buf.is_empty() {
+                buf.push(',');
+            }
+            let after_sep = buf.len();
+            value.write(StringCapture(&mut buf));
+            if buf.len() <= after_sep {
+                buf.truncate(before);
             }
         }
         self.string(&buf);
@@ -105,6 +105,7 @@ pub trait ValueWriter: Sized {
 }
 
 /// Adapter that captures a [`Value`]'s string representation into a buffer.
+/// Only captures string values; metrics and errors are ignored.
 struct StringCapture<'a>(&'a mut String);
 
 impl ValueWriter for StringCapture<'_> {
@@ -114,25 +115,12 @@ impl ValueWriter for StringCapture<'_> {
 
     fn metric<'a>(
         self,
-        distribution: impl IntoIterator<Item = Observation>,
+        _distribution: impl IntoIterator<Item = Observation>,
         _unit: Unit,
         _dimensions: impl IntoIterator<Item = (&'a str, &'a str)>,
         _flags: MetricFlags<'_>,
     ) {
-        use std::fmt::Write;
-        if let Some(obs) = distribution.into_iter().next() {
-            match obs {
-                Observation::Unsigned(v) => {
-                    let _ = write!(self.0, "{v}");
-                }
-                Observation::Floating(v) => {
-                    let _ = write!(self.0, "{v}");
-                }
-                Observation::Repeated { total, .. } => {
-                    let _ = write!(self.0, "{total}");
-                }
-            }
-        }
+        // Intentionally a no-op: values() is for string properties only.
     }
 
     fn error(self, _error: ValidationError) {}
--- a/metrique-writer-core/src/value/mod.rs
+++ b/metrique-writer-core/src/value/mod.rs
@@ -106,7 +106,7 @@ pub trait ValueWriter: Sized {
 
 /// Adapter that captures a [`Value`]'s string representation into a buffer.
 /// Only captures string values; metrics and errors are ignored.
-struct StringCapture<'a>(&'a mut String);
+pub(crate) struct StringCapture<'a>(pub(crate) &'a mut String);
 
 impl ValueWriter for StringCapture<'_> {
     fn string(self, value: &str) {
--- a/metrique-writer-core/src/value/mod.rs
+++ b/metrique-writer-core/src/value/mod.rs
@@ -86,14 +86,19 @@ pub trait ValueWriter: Sized {
 
     /// Write a list of values. Formats that support native arrays (e.g. EMF) can override this
     /// to emit a structured representation. The default joins each element's string representation
-    /// with `", "`.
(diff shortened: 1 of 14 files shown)


metrique-writer-core/src/value/mod.rs
```diff
@@ -83,6 +83,59 @@ pub trait ValueWriter: Sized {
     fn invalid(self, reason: impl Into<String>) {
         self.error(ValidationError::invalid(reason))
     }
+
+    /// Write a list of values. Formats that support native arrays (e.g. EMF) can override this
+    /// to emit a structured representation. The default joins each element's string representation
+    /// with `", "`, skipping elements that write nothing (e.g. `None`).
```
rcoh: Do we want `, ` or just comma? Are you concerned about values that contain commas?

metrique-writer-core/src/value/mod.rs
```diff
@@ -83,6 +83,59 @@ pub trait ValueWriter: Sized {
...
+    fn values<'a, V: Value + 'a>(self, values: impl IntoIterator<Item = &'a V>) {
+        let mut buf = String::new();
+        let mut elem = String::new();
+        for value in values {
+            elem.clear();
+            value.write(StringCapture(&mut elem));
```
rcoh: I think you can reuse buf here? Probably doesn't super matter

metrique-writer-core/src/value/mod.rs
```diff
@@ -83,6 +83,59 @@ pub trait ValueWriter: Sized {
...
+                buf.push_str(&elem);
+            }
+        }
+        self.string(&buf);
+    }
+}
+
+/// Adapter that captures a [`Value`]'s string representation into a buffer.
+struct StringCapture<'a>(&'a mut String);
+
```
rcoh, line 111: I feel like we must have this type somewhere already

metrique-writer-core/src/value/mod.rs
```diff
@@ -83,6 +83,59 @@ pub trait ValueWriter: Sized {
...
+impl ValueWriter for StringCapture<'_> {
+    fn string(self, value: &str) {
+        self.0.push_str(value);
+    }
+
+    fn metric<'a>(
```
rcoh, line 117: Interesting. I guess I wouldn't have assumed this behavior. I'm a little worried about this triggering automatically?

I would definitely rather this be strings and stringy types only.

Need to think about it more

rcoh (commented):
Fuzzing this would be a good idea

metrique-writer-core/src/value/mod.rs
jlizen: just comma should be fine. if somebody wants to pass in commas to the default impl, I can live with them seeing `",,,,,,,,"`. Any reasonable format will anyway have something like `[",",","]`

metrique-writer-core/src/value/mod.rs
jlizen: I'm good to apply it to stringly types only. It's just pretty problematic to not have something like this for them, since it is likely break any reasonable parsing otherwise (since we don't have eg a universal "null" across formats)

metrique-writer-core/src/value/mod.rs
jlizen: surprisingly, no, closest is `ToString`, but that is a `ValueFormatter`, not `ValueWriter`

jlizen:
- Added fuzzer support (no findings)
- Only drop `None` on stringly types in `StringCapture`
- Add native json `values()` support for json formatter as well
- Properly bridge the inner `values()` impl when type erasing with `BoxEntry`

metrique-writer-format-emf/src/emf.rs
```diff
@@ -1668,6 +1668,49 @@ impl ValueWriter<'_, '_> {
...
+                    let mean = if occurrences == 0 {
+                        0.0
+                    } else {
+                        total / occurrences as f64
+                    };
+                    if let Some(v) = clamp_to_finite(mean, "") {
+                        ValueWriter::write_float(self.0, v);
+                    }
+                }
+                _ => {}
```
rcoh: does this properly split by comma?

metrique-writer-format-json/src/json.rs
```diff
@@ -215,6 +215,26 @@ struct JsonValueWriter<'b, 'c> {
...
+    }
+
+    fn metric<'a>(
+        self,
+        _distribution: impl IntoIterator<Item = Observation>,
+        _unit: Unit,
+        _dimensions: impl IntoIterator<Item = (&'a str, &'a str)>,
+        _flags: MetricFlags<'_>,
+    ) {
+    }
```
rcoh, line 253: emf supports numbers by json does not, seems like a surprising gap

metrique-writer-format-emf/src/emf.rs
```diff
@@ -1668,6 +1668,49 @@ impl ValueWriter<'_, '_> {
...
+    }
+
+    fn metric<'a>(
+        self,
+        distribution: impl IntoIterator<Item = Observation>,
+        _unit: Unit,
+        _dimensions: impl IntoIterator<Item = (&'a str, &'a str)>,
+        _flags: MetricFlags<'_>,
+    ) {
+        if let Some(obs) = distribution.into_iter().next() {
```
rcoh: I don't see any tests of this codepath. Does it properly add commas?

rcoh: oh I see, it only takes the first element. Seems worth documenting (or just drop this)

rcoh (approved):

metrique-writer-format-emf/tests/misc.rs
```diff
@@ -122,3 +122,89 @@ fn test_background_queue_with_invalid_metric() {
...
+    let output: serde_json::Value = serde_json::from_str(&sink.dump()).unwrap();
+    assert_json_diff::assert_json_eq!(output["Plugins"], serde_json::json!(["only"]));
+}
+
+#[test]
+fn test_vec_emits_json_array_through_boxed_entry() {
+    let sink = TestSink::default();
+    let mut stream = Emf::all_validations("App".into(), vec![vec![]]).output_to(sink.clone());
+    let boxed = VecEntry {
+        plugins: vec!["a".into(), "b".into()],
```
rcoh, line 204: can you also test `u64` here?

metrique-writer-format-emf/src/emf.rs
jlizen: Currently the EmfArrayElementWriter (and json equivalent) are only ever called once per vec (internally at least).

But we might be passed a custom `Value` impl with repeated observations, we should support it.

metrique-writer-format-json/src/json.rs
jlizen: Will add

metrique-writer-format-emf/src/emf.rs
jlizen: Yep, the outer values loop handles it.

jlizen:
@rcoh I updated to:
- allow Observations in` values()` for emf/json/default
    - json/emf - nested array
    - default - flatten into root comma-separated string
- support numerics in all types
    - json/emf: preserve as numeric, but don't add any extra format handling (metric representation, units, etc)
    - default - include inside the single string

That work for you?

PR #271, opened 2026-04-14
refactor: use workspace dependencies for inter-crate deps

jlizen:
📬 *Issue #, if available:*

✍️ *Description of changes:*

Cutting over to workspace-wide local crate version management, rather than explicit path + version blocks inside crates.

I left root at `0.x` to avoid needing to change it every minor version bump.

The `check_docs.rs` support was pretty gnarly because `workspace = true` essentially resolves to workspace patches - but then, our docs.rs. packaged copies have different paths. Previously, the sibling paths were directly embedded in the packaged bundles, so we didn't have this problem.

We fixed it by modifying the `workspace.dependencies` to point to the packaged copy instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,22 @@ members = [
 ]
 
 [workspace.dependencies]
+# Internal subcrates
+metrique = { version = "0.1", path = "metrique", default-features = false }
+metrique-aggregation = { version = "0.1", path = "metrique-aggregation" }
+metrique-core = { version = "0.1", path = "metrique-core" }
+metrique-macro = { version = "0.1", path = "metrique-macro" }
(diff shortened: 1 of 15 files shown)


jlizen:
Ah, i'll need to set `default-features = true` locally if we want `default-features = false` globally... let me see...

rcoh (approved):
lgtm, please double check that cargo publish --dryrun works

jlizen:
`cargo publish --dryrun` at the workspace root does NOT work since it can't figure out proper publish ordering for these inter-workspace dependencies. That is a well-understood gap that release-plz helps with.

`release-plz update` does figure out proper ordering, and each individual crate can be packaged via dryrun when following that ordering.

So, LGTM.

PR #273, opened 2026-04-14
docs: reframe terminology to lead with "wide events"

jlizen:
📬 *Issue #, if available:*

✍️ *Description of changes:*

Reframes documentation to lead with "wide events" as the primary concept, positioning unit-of-work as the most common type of wide event that metrique is particularly tailored for.

The previous framing implied metrique was exclusively for unit-of-work metrics, but it's really a wide event library: periodic gauges, background jobs, and other non-request-scoped metrics are equally valid uses. "Wide events" is also a more widely recognized term in the observability community, making it easier for new users to understand what metrique does and how it compares to alternatives like OpenTelemetry.

The `pub mod unit_of_work` API surface, example filenames, and the "Unit-of-work" cookbook pattern name are all intentionally unchanged.


metrique-aggregation/Cargo.toml
```diff
@@ -4,25 +4,25 @@ version = "0.1.11"
 edition = "2024"
 rust-version = "1.89" # build.yml
 license = "Apache-2.0"
-description = "Library for working with unit of work metrics - aggregation"
```
rcoh, line 7: maybe "library for aggregating metrique events"?

rcoh (approved):

Issue #263, opened 2026-04-10
metrique-util: State<T> should snapshot via CloseValue, not Clone


jlizen:
`State<T>` requires `T: Clone` for snapshotting. This prevents wrapping structs that contain `Slot` fields (which are oneshot channels and can't be cloned). `State` should support snapshotting via `T::close()` instead, which `Slot` handles correctly (returns `None` if unfilled). This would allow `State<MyEnvironment>` where `MyEnvironment` contains `Slot<Option<String>>` fields for progressive population.

jlizen:
Actually, we want `CloseValueRef`, which would allow using a `OnceLock` inside

jlizen:
Closing as discussed here: https://github.com/awslabs/metrique/pull/274#issuecomment-4248351967

PR #274, opened 2026-04-14
feat(util): add StateRef for non-Clone types in State

jlizen:
📬 *Issue #, if available:*

#263 

✍️ *Description of changes:*

`State<T>` requires `T: Clone` for its `CloseValue` impls because it extracts the value from an internal `Arc` via `Arc::unwrap_or_clone`. This prevents using `State` with structs containing non-Clone fields like `OnceLock`, which are useful for progressive population (fields that start empty and are filled during request processing).

Rather than changing `State`'s bounds (which would break existing uses like `State<String>`, since `&String` intentionally doesn't implement `CloseValue`), this adds a new `StateRef<T>` type that requires `T: CloseValueRef` instead of `T: Clone + CloseValue`. `StateRef` closes by reference through the `Arc`, so no cloning is needed.

Both types share the same snapshot semantics via an internal `StateInner` struct. The choice between them is simple: use `State` when `T` is `Clone`, use `StateRef` when `T` implements `CloseValue` for both owned and borrowed (e.g. `#[metrics(subfield)]` structs containing `OnceLock` fields).

The global-state example is updated to demonstrate progressive population with `OnceLock` fields in a `StateRef`-wrapped struct.

### DIFF
--- a/metrique-util/src/lib.rs
+++ b/metrique-util/src/lib.rs
@@ -8,7 +8,7 @@
 #[cfg(feature = "state")]
 mod state;
 #[cfg(feature = "state")]
-pub use state::{LatestRef, State};
+pub use state::{LatestRef, State, StateRef};
 
 #[cfg(feature = "tokio-metrics-bridge")]
 mod dynamic_inflection;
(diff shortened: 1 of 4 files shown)


jlizen:
Ehhh this is probably too niche at this point. My only motivating example is OnceLock, which is basically usable directly, as long as you don't care if OnceLock containing value shows up in metrics but wasn't available in handler. Which, I don't.

Let's hold off unless new use cases arise.

Issue #240, opened 2026-03-17
feat: support deferred sink attachment with bounded entry buffering


jlizen:
## Problem

Some sinks require async initialization that may take hundreds of milliseconds (e.g. credential exchange, network handshakes). Today, `attach_to_stream` must be called before any code calls `sink()` or `append()`, otherwise the process panics. This forces applications to block startup on sink initialization, delaying readiness to serve traffic.

## Desired Behavior

Allow attaching a sink at startup that buffers entries while the underlying stream is still initializing, then transparently drains into the real stream once it's ready. If initialization fails, buffered entries are dropped and the stream becomes a no-op.

From the caller's perspective, there's no separate "pending" phase; the stream just works, buffering entries until the worker is ready to drain them.

The buffer should be bounded (and ideally consistent with existing metrique backpressure behavior: drop oldest when full).

## Use Cases

- Sinks that require credential vending (e.g. Cognito, STS) before they can publish metrics to a remote service
- Sinks that depend on network resources that may be slow to initialize
- Applications where startup latency is critical and telemetry is best-effort

## Relationship to #225

The executor/spawner abstraction proposed in #225 may provide a natural path to solving this.

If the async `EntryIoStream` backend manages its own runtime or accepts a spawner, the stream could handle deferred initialization internally: the `EntryIoStream` exists and accepts entries immediately, while the async submission worker is still spinning up in the background.

PR #275, opened 2026-04-14
 feat(util): add pending_sink for deferred sink attachment with bounded buffering

jlizen:
📬 *Issue #, if available:*
#240 

✍️ *Description of changes:*

Adds `pending_sink::new(capacity)` in `metrique-util` for deferred sink attachment with bounded entry buffering. This lets applications start emitting metrics immediately during startup, even when the underlying sink requires async initialization (credential exchange, network handshakes, etc.).

`pending_sink::new()` returns a `(BoxEntrySink, PendingSinkResolver)` pair. The sink can be used (and attached to a `GlobalEntrySink`) right away. While the resolver is outstanding, entries are buffered in a bounded lock-free ring buffer. When the buffer is full, the oldest entry is dropped, consistent with `BackgroundQueue` backpressure behavior. Calling `resolver.resolve(real_sink)` drains the buffer into the real sink and switches to direct forwarding. If the resolver is dropped without calling `resolve` (initialization failure), the sink becomes a no-op and buffered entries are discarded.

The implementation lives in `metrique-util` behind a `pending-sink` feature flag (gating optional deps on `metrique-writer-core` and `crossbeam-queue`). It uses `OnceLock` for the one-time state transition and `AtomicBool` for cancellation, so the hot path (`append` after resolution) is a single atomic load with no locking.

### DIFF
--- a/metrique-util/src/pending_sink.rs
+++ b/metrique-util/src/pending_sink.rs
@@ -1,0 +1,335 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Deferred sink attachment with bounded entry buffering.
+//!
+//! See [`new()`] for details.
+
+use std::sync::{
+    Arc, OnceLock,
+    atomic::{AtomicBool, AtomicUsize, Ordering},
+};
+
+use crossbeam_queue::ArrayQueue;
+use metrique_writer_core::{
+    entry::BoxEntry,
+    sink::{BoxEntrySink, EntrySink, FlushWait},
+};
+
+struct Inner {
+    buffer: ArrayQueue<BoxEntry>,
+    sink: OnceLock<BoxEntrySink>,
+    cancelled: AtomicBool,
+    /// Number of append() calls currently in the buffering path (between the
+    /// sink.get() check and the force_push completion). resolve() waits for
+    /// this to reach zero before draining, preventing stranded entries.
+    buffering: AtomicUsize,
+}
+
+struct PendingSink(Arc<Inner>);
+
+impl EntrySink<BoxEntry> for PendingSink {
+    fn append(&self, entry: BoxEntry) {
+        if let Some(sink) = self.0.sink.get() {
+            sink.append(entry);
+        } else if !self.0.cancelled.load(Ordering::Acquire) {
+            self.0.buffering.fetch_add(1, Ordering::AcqRel);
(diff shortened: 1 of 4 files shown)


metrique-util/src/pending_sink.rs
```diff
@@ -0,0 +1,284 @@
...
+#[must_use = "if dropped without calling resolve(), buffered entries will be discarded"]
+pub struct PendingSinkResolver(Option<Arc<Inner>>);
+
+impl PendingSinkResolver {
+    /// Drain all buffered entries into `sink` and switch to direct forwarding.
+    ///
+    /// After this call, new entries appended to the associated [`BoxEntrySink`] will
+    /// go directly to `sink`. This method consumes the resolver.
+    ///
+    /// Note: there is a brief window where concurrent `append()` calls may still
```
rcoh: there is a race condition that can drop entries during the switch window (hard to trigger IRL but it is possible, see test below)
```
  1. Thread A: append() sees sink.get() → None
  2. Thread B: resolve() sets the sink via OnceLock::set()
  3. Thread B: drain loop runs and finishes (buffer empty)
  4. Thread A: force_push(entry) — entry is now stranded in the buffer
```

```rust
 #[test]
  fn stress_no_entry_loss_on_concurrent_resolve() {
      use std::sync::atomic::AtomicUsize;
      use std::sync::Barrier;

      for _ in 0..1000 {
          let n = 100;
          let threads = 4;
          let expected = n * threads;

          let (sink, resolver) = new(expected + 1024);
          let barrier = Arc::new(Barrier::new(threads + 1));

          let (real_sink, appended, _) = collector();

          let handles: Vec<_> = (0..threads)
              .map(|_| {
                  let sink = sink.clone();
                  let barrier = barrier.clone();
                  std::thread::spawn(move || {
                      barrier.wait();
                      for i in 0..n {
                          sink.append_any(TestEntry(i as u64));
                      }
                  })
              })
              .collect();

          barrier.wait();
          resolver.resolve(real_sink);

          for h in handles {
              h.join().unwrap();
          }

          let got = appended.lock().unwrap().len();
          assert_eq!(
              got, expected,
              "lost {} entries in this iteration",
              expected - got
          );
      }
  }
```

jlizen: Thanks, I added an `AtomicUsize` counter that tracks in-flight appenders between the `sink.get()` check and the `force_push()`. `resolve()` sets the sink then waits on the counter reaching 0 before draining the buffer.

rcoh (approved):

PR #282, opened 2026-05-04
docs(design): entry descriptor + field tag system

jlizen:
📬 *Issue #, if available:*

✍️ *Description of changes:*

Related: [dial9-tokio-telemetry#346](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/346) (the first concrete consumer).

### Summary

Design-only PR. No implementation yet.

Adds a small system on top of metrique's existing `Entry` / `Value` / `CloseValue` traits that lets sinks introspect entry structure and lift structural context out of closed entries:

- An **entry descriptor** (`&'static EntryDescriptor`) describing each macro-derived entry's closed shape: ordered fields, tags, optionality, lists, dynamic-key maps, units, and sources.
- A **source** system (`#[metrics(source(T))]` attribute, `SourceTag` trait with `type Snapshot` and optional `register_descriptor` hook) that lets entries declare typed structural capabilities and lets sinks extract snapshots from the closed entry via `desc.source::<T>(..)`.
- A **field tag** system (`#[metrics(default_field_tag(...))]`, `#[metrics(field_tag(...))]`, `skip(T)`) that lets sinks define their own opt-in markers and lets users apply them at struct or field scope.
- A narrow **`no_write`** attribute for fields that must participate in close but not in normal emission.

None of this changes the existing `Entry`, `Value`, or `CloseValue` traits. Sinks that do not need descriptors pay no runtime cost.

The immediate driver is dial9's per-request tracing sink (see the related PR). The design is generic: OTEL, privacy-tiered export, audit, and other observability sinks can use the same mechanism with their own tag types.

### Documents

In-tree permanent:

- `docs/entry-descriptors.md`: keeper. API reference for the descriptor model, `SourceTag`, field tags, `no_write`, architecture, units, list/flex shapes, validation, future evolution.

Review-only (deleted as part of sign-off):

- `docs/entry-descriptors-review.md`: problem statement, core requirements, non-goals, tradeoffs, rationale for the combination of pieces, field-tag resolution rules, rejected alternatives, feasibility checks, evolution path for hand-written `Entry` impls.
- `docs/entry-descriptors-impl-plan.md`: sequencing, module-level work (`metrique-writer-core`, `metrique-macro`, `metrique-core`), public API signatures, testing plan, risks.

### Downstream consumers

dial9-tokio-telemetry is the first consumer; see the related PR for the sink-side design that consumes `EntryDescriptor`, `SourceTag`, the `InTrace` / `InternString` field tags, and the new `List` and `Flex` shapes. That PR cannot merge until this one lands on a released version.


docs/entry-descriptors.md
```diff
@@ -0,0 +1,426 @@
...
+pub struct Export;               // field tag
+pub struct RequestContext;       // source tag
+pub struct RequestCtx {
+    pub request_id: String,
+    pub started_at_monotonic_ns: u64,
+}
+
+// Application entry uses them.
+#[metrics(source(audit::RequestContext))]
+#[metrics(default_field_tag(skip(audit::Export)))]
```
rcoh: written out english explaining what these things to would be really helpful for understanding the design. 

possibly an issue with the nomenclature that I don't understand them by looking as is

docs/entry-descriptors.md
```diff
@@ -0,0 +1,426 @@
+# Entry descriptors, sources, and field tags
+
+A system on top of metrique's existing `Entry` / `Value` / `CloseValue` traits that lets sinks introspect entry structure and lift structural context out of closed entries.
```
rcoh: can you add a glossary of the new terms here?

docs/entry-descriptors.md
```diff
@@ -0,0 +1,426 @@
...
+
+Because the descriptor is `#[non_exhaustive]` all the way through, future metrique versions can add `KnownShape` variants without breaking hand-written `DescribeEntry` implementors, and new descriptor-aware sinks can introspect older descriptors without compilation breaks.
```
rcoh: sort of, but not really in practice. e.g. if you add `&'static str` and consumers ignore it, you can't _actually_ switch to using it.

The best way around this issue is to provide accessor methods and suggest that consumers use those, e.g. `.as_str()`

docs/entry-descriptors.md
```diff
@@ -0,0 +1,426 @@
...
+
+### The Opaque trapdoor
+
+A field whose closed shape is `FieldShape::Opaque` is fully functional through `Entry::write` (every `Value` impl works; EMF and JSON handle it fine), but descriptor-aware sinks that selected it via a tag have no wire-level shape guarantee for it. Typical sinks skip opaque fields with a diagnostic and continue. This is the price of letting user types implement `Value` without a parallel descriptor hook.
+
+The most common current Opaque case is distribution-typed fields: `metrique_aggregation::Histogram<T>`, `SharedHistogram<T>`, and user-defined types that emit multiple `Observation`s with the `Distribution` flag. The descriptor has no way to represent "this field emits 0..N observations of an inner scalar type." Such fields are safe to use on EMF/JSON sinks today. Tagging them for a descriptor-aware sink produces a diagnostic and skips the field on that sink; see "Future evolution" for the planned `FieldShape::Distribution` variant.
```
rcoh: interesting. I guess we could either:

1. Add a `Value` into dial9 that can hold multiple values
2. Do something like we do for LocalFormat where we push rollups

in either case, I think simply omitting this is fine

docs/entry-descriptors.md
```diff
@@ -0,0 +1,426 @@
...
+
+Lookup goes through the erased entry vtable, not by widening `BoxEntry`:
+
+```rust
+trait ErasedEntry {
+    fn write_erased(&self, w: &mut dyn ErasedEntryWriter);
+    fn inner_any(&self) -> &(dyn Any + Send + 'static);
+    fn descriptor(&self) -> Option<&'static EntryDescriptor>;
+}
+```
```
rcoh: hm...why not add Descriptor onto the trait with a default `None` impl? Not totally clear why we want another trait here (but I could be convinced)

rcoh (commented):
the key ideas I'm aligned with.

My gut-check feedback is this is a lot of new traits and I wonder what they are all for and if they are all needed.

I feel like given our current N=1 use case we go as simple as possible then expand when N=2 (otel?)

jlizen:
From offline discussion, my plan is to scope down to only static tag system + entry descriptor system.

The source system is a nicety for dial9 but not strictly needed. The main thing we give up is certain validations are later. We might want to add a smoke test to help lint against mistakes.

We can see of Otel cares more, or if static tag system suffices there too.

docs/entry-descriptors.md
jlizen: Yeah there definitely are paths to handle it, but planning to punt for V1

docs/entry-descriptors.md
jlizen: You make a good point, i'll refactor to use accessors and keep lifetimes self-bounded

docs/entry-descriptors.md
jlizen: Agreed, thanks

jlizen:
This is ready for fresh eyes @rcoh / @yulnr , I'll get started on the schema field extensions in the meantime.

There is a glossary now, and also a changelog since last review.

rcoh (approved):

yulnr (approved):

rcoh (approved):

Issue #205, opened 2026-02-17
Improve reporting of the BackgroundQueueBuilder metrics


rcoh:
Right now, `BackgroundQueueBuilder` can emit metrics via metrics.rs but this is really quite inflexible and hard to deal with.

Consider instead having a trait of the individual metric actions that customers that can provide — we can still impl that trait for metrics.rs but this will make it much easier to capture metrics from the BackgroundQueue.

PR #283, opened 2026-05-04
sysinfo bridge

### LINKED ISSUE #255: sysinfo <-> metrique bridge
[sysinfo](https://crates.io/crates/sysinfo) metrics can be a bit subtle to collect properly. It would be good to have a bridge in metrique to make it easy.

```
ServiceMetrics::attach_to_stream(
    Emf::builder("Ns".to_string(), vec![vec![]])
        .build()
        .output_to_makewriter(RollingFileAppender::new(
            Rotation::MINUTELY,
            &service_log_dir,
            "service_log.log",
        )),
).with_sysinfo()
```

See also: #248

fmzbl:
closes: #255

adds a sysinfo-bridge feature to metrique-util that periodically samples system and current-process metrics (CPU, memory, swap, load average, uptime) and appends them to a global entry sink. same shape and lifecycle as the tokio-metrics bridge from #256 : configurable interval and name style. includes tests, an example, and README/docs entries.


metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,503 @@
...
+        network_errors_on_transmitted: net.errors_on_transmitted,
+        network_total_errors_on_transmitted: net.total_errors_on_transmitted,
+
+        component_count: components.list().len() as u64,
+        component_max_temperature,
+        component_max_temperature_recorded,
+
+        process_memory: process.map(|p| p.memory()).unwrap_or(0),
+        process_virtual_memory: process.map(|p| p.virtual_memory()).unwrap_or(0),
+        process_cpu_usage: process.map(|p| p.cpu_usage()).unwrap_or(0.0),
```
jv1i, line 431: A note about `cpu_usage()` and it's delta behavior, from the sysinfo docs:
```
/// ⚠️ To start to have accurate CPU usage, a process needs to be refreshed
/// **twice** because CPU usage computation is based on time diff (process
/// time on a given time period divided by total system time on the same
/// time period).
///
/// ⚠️ If you want accurate CPU usage number, better leave a bit of time
/// between two calls of this method (take a look at
/// [`MINIMUM_CPU_UPDATE_INTERVAL`][crate::MINIMUM_CPU_UPDATE_INTERVAL] for
/// more information).
```
So we might want to do an initial refresh and sleep (of at least `MINIMUM_CPU_UPDATE_INTERVAL`), before the loop, so that the initial values are always valid. Maybe we do a refresh of everything before the loop just to be safe? call `refresh_cpu_all()`, `system.refresh_processes(..)`, `networks.refresh(true)`, etc.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,503 @@
...
+        Self { interval, ..self }
+    }
+
+    /// Set the name style for emitted metric fields.
+    ///
+    /// Defaults to [`MetricNameStyle::Identity`].
+    pub fn with_name_style(self, name_style: MetricNameStyle) -> Self {
+        Self { name_style, ..self }
+    }
+}
```
jv1i, line 107: Do we want to add some configuration so that users can opt-in to certain metrics instead of tracking everything from sysinfo by default? e.g. `with_disks`, `with_networks`, `with_components`, etc.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// Number of thermal/component sensors being tracked.
+    pub component_count: u64,
+    /// Maximum current temperature across all components, in degrees Celsius.
+    /// `0.0` if no component reports a temperature.
+    pub component_max_temperature: f32,
+    /// Maximum recorded temperature across all components, in degrees Celsius.
+    /// `0.0` if no component reports a max.
+    pub component_max_temperature_recorded: f32,
+
+    // ----- Current process -----
```
jv1i: Should we add to the struct-level doc some note about per-process values? Something that makes clear that we are not iterating over all processes, e.g.:
```
/// `process_*` fields measure the process running this reporter: i.e. your                                                                                                                          
/// service binary. Child processes, forked workers, and sidecars are not included.
```
I think tracking only the main process is ok for a v1 but perhaps we should be very clear about what fields prefixed `process_*` mean

metrique-util/src/lib.rs
```diff
@@ -10,15 +10,25 @@ mod state;
...
 
+#[cfg(feature = "sysinfo-bridge")]
+mod sysinfo_reporter;
+#[cfg(feature = "sysinfo-bridge")]
+pub use sysinfo_reporter::{AttachGlobalEntrySinkSysinfoExt, SysinfoMetrics, SysinfoMetricsConfig};
+// `MetricNameStyle` is shared between the two bridges. When tokio-metrics-bridge
+// is enabled it is re-exported above; otherwise expose it directly here.
+#[cfg(all(feature = "sysinfo-bridge", not(feature = "tokio-metrics-bridge")))]
+pub use metrique_core::DynamicNameStyle as MetricNameStyle;
+
```
jv1i, line 32: Can we drop the re-export from `tokio_metrics_reporter.rs` and just have one re-export here?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,503 @@
...
+/// A snapshot of system-wide and current-process metrics sampled from
+/// [`sysinfo`].
+///
+/// Field names mirror sysinfo's API verbatim. See the [sysinfo docs] for
+/// platform-specific behavior and the meaning of each value.
+///
+/// [sysinfo docs]: https://docs.rs/sysinfo
+#[metrics]
+pub struct SysinfoMetrics {
+    // ----- CPU -----
```
jv1i: We might want to make this `#[metrics(subfield)]`/`#[metrics(subfield_owned)]` to later be folded/flattened into other metric structs.

yulnr (commented):
Looking good! leaving some comments from a quick first pass

jv1i:
Thinking about avoiding noise:

I wonder if we should skip some values that are constant like `num_cpus`, `physical_core_count`, etc. that I think shouldn't change during runtime. We could either:
- always include them (now)
- include them under config (opt-in) 
- skip them entirely
- emit them only at startup separately

I'm leaning towards make them opt-in or skipping them, but not strong opinion.  

Also, we have some cumulative+delta pairs like `network_received` (delta) and `network_total_received` (cumulative), same for `process_disk_read_bytes`/`process_disk_total_read_bytes`. Should we include both as it is now, or just one of them (cumulative or delta) and let consumers downstream calculate the other?

fmzbl:
ill skip them. idk about the last thing, its what sysinfo already gives, as a user i would expect them probably.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,670 @@
...
+            .map(|p| p.disk_usage().total_read_bytes)
+            .unwrap_or(0),
+        process_disk_written_bytes: process.map(|p| p.disk_usage().written_bytes).unwrap_or(0),
+        process_disk_total_written_bytes: process
+            .map(|p| p.disk_usage().total_written_bytes)
+            .unwrap_or(0),
+        process_open_files: process
+            .and_then(|p| p.open_files())
+            .map(|v| v as u64)
+            .unwrap_or(0),
```
jv1i: Should we make these fields `Option` instead, so that if process is `None` these fields are also `None`? (using Option will skip None fields). Defaulting could be interpreted as `cpu_usage: 0` = no cpu usage, when it in reality it might mean "we couldn't access the process info" (not sure how likely it is for process to be None).

jv1i:
I'm not entirely sure if summing across all network interfaces, or disks is actually very useful for users (i.e. how can the use that information). (sorry for not catching this earlier)

The one solution I can think of, would be to do a manual `impl Entry for SysinfoMetrics` (rather than using 
`#[metrics]`) so that in `write()` we can iterate over each disk/interface and emit individual entries using `AllowSplitEntries`.  

It could look something like this:
```rust
pub struct NetworkInterfaceMetrics {
    pub interface_name: String,     // becomes `InterfaceName` dimension
    pub received: u64,
    pub transmitted: u64,
    pub total_received: u64,
    pub total_transmitted: u64,
    // ... packets_*, errors_*, totals
}
```

```rust
struct SysinfoEntry { // no longer users #[metrics]
    scalars: DynamicInflectionEntry<<SysinfoMetrics as CloseValue>::Closed>, // separates scalar values from others
    disks: Vec<DiskMountMetrics>,
    networks: Vec<NetworkInterfaceMetrics>,
    components: Vec<ComponentSensorMetrics>,
    name_style: MetricNameStyle,
}

impl Entry for SysinfoEntry {
    fn write<'a>(&'a self, w: &mut impl EntryWriter<'a>) {
        w.config(&const { AllowSplitEntries::new() });

        self.scalars.write(w);

        for n in &self.networks {
            let dim = ("InterfaceName", n.interface_name.as_str());
            w.value("network_received",
                    &n.received.with_dimension(dim.0, dim.1.to_owned()));
            // ... transmitted, packets_*, errors_*, totals (same pattern)
        }
        for d in &self.disks { ... }
        for c in &self.components { ... }
    }
}
```
This would result in multiple entries:
```rust
// Scalars
{"_aws":{"CloudWatchMetrics":[{"Namespace":"Sysinfo","Dimensions":[[]],"Metrics":[{"Name":"cpu_usage"},{"Name":"used_memory"},/*…*/]}],"Timestamp":…},"cpu_usage":12.5,"used_memory":8000000000,/*…*/}

// eth0
{"_aws":{"CloudWatchMetrics":[{"Namespace":"Sysinfo","Dimensions":[["InterfaceName"]],"Metrics":[{"Name":"network_received"},{"Name":"network_transmitted"},/*…*/]}],"Timestamp":…},"InterfaceName":"eth0","network_received":4500000,"network_transmitted":120000,/*…*/}

// lo
{"_aws":{/*…same metrics, dimensioned by InterfaceName…*/},"InterfaceName":"lo","network_received":12000,"network_transmitted":12000,/*…*/}

// root mount
{"_aws":{"CloudWatchMetrics":[{"Namespace":"Sysinfo","Dimensions":[["MountPoint"]],"Metrics":[{"Name":"disk_total_space"},{"Name":"disk_available_space"}]}],"Timestamp":…},"MountPoint":"/","disk_total_space":1.0e11,"disk_available_space":4.0e10}
```

Not sure off the top of my head if the value of that is worth the implementation, if we should keep the current per-category totals or drop it for v1.

cc @jlizen I think you recently had to manually integrate sysinfo, if that was with metrique, what did you do for per-item dimensioned metrics? or do we think this is important for current users?

jlizen:
> cc @jlizen I think you recently had to manually integrate sysinfo, if that was with metrique, what did you do for per-item dimensioned metrics? or do we think this is important for current users?

For network interfaces, I just captured whether they were up or down, written out as `Vec<String>` where `String` is eg: `eth0:up`. I didn't capture anything else to do with dimensionality, just rolled up across.

I'd suggest starting with rollups for v1 and figuring out what more detailed API makes sense from there.

Not to say that it is a good way to do it, here's sample code from my usage. I didn't spend very long thinking about it.

This was my aggregate emissions struct:
```
 /// System load summary, sampled every 2s on a background thread.
	/// Per-sample wide events are written to `sysinfo.jsonl` for time-series
	/// reconstruction. These scalar max values provide a quick summary on the
	/// per-invocation EMF record.
	///
	/// CPU usage is the average across all logical cores (0-100%). Memory values
	/// come from the OS (not cgroup-aware). Short invocations may
	/// have zero samples (Option fields will be None).
	#[metrics(subfield_owned, rename_all = "PascalCase")]
	pub struct SysinfoMetrics {
	    /// Number of logical CPU cores.
	    pub sysinfo_cpu_cores: u64,
	    /// Number of samples collected.
	    pub sysinfo_sample_count: u64,
	    /// Total system memory in MB.
	    pub sysinfo_total_memory_mb: u64,
	    /// Peak global CPU usage (%) across all samples.
	    pub sysinfo_cpu_max_pct: Option<f64>,
	    /// Peak used memory (MB) across all samples.
	    pub sysinfo_mem_used_max_mb: Option<u64>,
	    /// Peak network RX bytes/sec across all sample intervals.
	    pub sysinfo_net_rx_max_bytes_per_sec: Option<u64>,
	    /// Peak network TX bytes/sec across all sample intervals.
	    pub sysinfo_net_tx_max_bytes_per_sec: Option<u64>,
	    /// Interface names and up/down state at period start, e.g. "eth0:up,lo:up".
	    pub sysinfo_net_interfaces: Option<String>,
	    /// Peak open file descriptor count observed.
	    pub sysinfo_fd_max: Option<u64>,
	    /// Soft RLIMIT_NOFILE at build start.
	    pub sysinfo_fd_limit: Option<u64>,
	    /// Peak FD utilization (fd_max / fd_limit).
	    pub sysinfo_fd_utilization: Option<f64>,
	}
```

And this was my per-snapshot:
```
	/// Per-sample EMF record emitted to the same metrics file as the invocation record.
	struct SysinfoSampleMetrics {
	    #[metrics(timestamp)]
	    sampled_at: Timestamp,
	    cpu_pct: Option<f64>,
	    mem_used_mb: u64,
	    net_rx_bytes_per_sec: Option<u64>,
	    net_tx_bytes_per_sec: Option<u64>,
	    fd_count: Option<u64>,
	}
```

fmzbl:
i added a couple fields from that list, it should be ready now

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Total swap space in bytes.
+    pub total_swap: u64,
+    /// Used swap space in bytes.
+    pub used_swap: u64,
+    /// Free swap space in bytes.
+    pub free_swap: u64,
+
+    // ----- Load average -----
+    /// 1-minute load average. `0.0` on platforms without load average support.
+    pub load_average_one: f64,
```
jlizen: These `0.0` are potentially very confusing, especially if viewing metrics across services that do and don't have platform support.

We should instead use `Option<f64>` and have it be none if there is no support.

jlizen: Same goes for the other places that don't have support (`RLIMIT_NOFILE`, etc)

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+        process_disk_total_read_bytes: process
+            .map(|p| p.disk_usage().total_read_bytes)
+            .unwrap_or(0),
+        process_disk_written_bytes: process.map(|p| p.disk_usage().written_bytes).unwrap_or(0),
+        process_disk_total_written_bytes: process
+            .map(|p| p.disk_usage().total_written_bytes)
+            .unwrap_or(0),
+        process_open_files: process
+            .and_then(|p| p.open_files())
+            .map(|v| v as u64)
```
jlizen: nit: why not just store it as `usize` instead of casting it?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+/// use std::time::Duration;
+///
+/// let _handle = ServiceMetrics::attach_to_stream(emf.output_to(std::io::stderr()));
+///
+/// let config = SysinfoMetricsConfig::default()
+///     .with_interval(Duration::from_secs(30))
+///     .with_name_style(MetricNameStyle::PascalCase);
+/// ServiceMetrics::subscribe_sysinfo_metrics(config);
+/// ```
+pub trait AttachGlobalEntrySinkSysinfoExt: AttachGlobalEntrySink + 'static {
```
jlizen, line 274: Great example, thanks. Though, i wonder if it might be more discoverable on `subscribe_sysinfo_metrics`?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+/// Extension methods for subscribing system metrics to a global entry sink.
+///
+/// Spawns a background task that periodically samples [`SysinfoMetrics`] from
+/// [`sysinfo`] and appends each snapshot to the sink. The task is automatically
+/// aborted when the [`AttachHandle`](metrique::writer::sink::AttachHandle) is
+/// dropped.
+///
+/// # Example
+///
+/// ```rust,ignore
```
jlizen: nit: why not `no_run`?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+        if let Err(err) = worker.await
+            && !err.is_cancelled()
+        {
+            tracing::error!("sysinfo metrics reporter panicked: {err}");
+        }
+    });
+    abort
+}
+
+#[cfg(test)]
```
jlizen, line 534: We should also have tests for:
  - process_memory > 0 (the test process itself is always
  using memory)
  - process_run_time > 0 or process_run_time >= 0 (depending
  on timing sensitivity)
  - process_cpu_usage exists as a key (though I suspect since we are starting tests with time paused, it might show as `0.0`

This would give us some confidence that process fields are actually emitted, without testing full sysinfo internals

metrique/examples/sysinfo.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jlizen: Would it be worth having just a consolidated `extras.rs` that demonstrates attaching tokio metrics, sysinfo metrics, whatever N+1 will be?

These individual examples are simple enough that they don't have a lot of value over just the doc comments, but I'd still like to keep the discoverability

jlizen (changes requested):
Overall impl looks great. The three blocking concerns are quickies:
- Nowhere should we default to `0` as a sentinel value for "could not load" / "not enabled". Instead, let's use `Option<T>`
- I'd like to see some simple probing of process-level metrics in tests
- all the metrics fields should be `non_exhaustive`

Rest are optional nits

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+            track_disks: false,
+            track_networks: false,
+            track_components: false,
+        }
+    }
+}
+
+impl SysinfoMetricsConfig {
+    /// Return a config with a custom sampling interval.
+    pub fn with_interval(self, interval: Duration) -> Self {
```
jlizen, line 73: We should probably mention the default here, and also that if you set the interval less than ~200ms, cpu readings will be unreliable (since we call `refresh_cpu_all` inside `sample()`.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+/// Field names mirror sysinfo's API verbatim. See the [sysinfo docs] for
+/// platform-specific behavior and the meaning of each value.
+///
+/// `process_*` fields measure the process running this reporter: i.e. your
+/// service binary. Child processes, forked workers, and sidecars are not
+/// included.
+///
+/// [sysinfo docs]: https://docs.rs/sysinfo
+#[metrics(subfield_owned)]
+pub struct SysinfoMetrics {
```
jlizen, line 122: This should be `#[non-exhaustive]`

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Component (thermal sensor) metrics. `None` (entirely omitted from
+    /// output) unless [`SysinfoMetricsConfig::with_components`] is set.
+    #[metrics(flatten)]
+    pub components: Option<ComponentMetrics>,
+}
+
+/// Aggregated disk metrics across all mounted disks. Emitted as part of
+/// [`SysinfoMetrics`] when [`SysinfoMetricsConfig::with_disks`] is set.
+#[metrics(subfield_owned)]
+pub struct DiskMetrics {
```
jlizen, line 205: `#[non_exhaustive]`

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Sum of total space across all mounted disks, in bytes.
+    pub total_disk_space: u64,
+    /// Sum of available space across all mounted disks, in bytes.
+    pub available_disk_space: u64,
+}
+
+/// Aggregated network metrics across all interfaces. Emitted as part of
+/// [`SysinfoMetrics`] when [`SysinfoMetricsConfig::with_networks`] is set.
+#[metrics(subfield_owned)]
+pub struct NetworkMetrics {
```
jlizen, line 218: `#[non_exhaustive]`

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Transmit errors across all interfaces since the previous refresh.
+    pub network_errors_on_transmitted: u64,
+    /// Cumulative transmit errors across all interfaces.
+    pub network_total_errors_on_transmitted: u64,
+}
+
+/// Component (thermal sensor) metrics. Emitted as part of [`SysinfoMetrics`]
+/// when [`SysinfoMetricsConfig::with_components`] is set.
+#[metrics(subfield_owned)]
+pub struct ComponentMetrics {
```
jlizen, line 254: `#[non_exhaustive]`

metrique-util/src/sysinfo_reporter.rs
fmzbl: it fails, it would need the global_entry_sync!

jlizen: Couldn't the example just import `ServiceMetrics`? But even if we need more setup, having it hidden with a leading `#` is good so we can make sure the API doesn't go out of sync with the example.

fmzbl:
thanks for the feedback, everything should be covered and ordered in commits.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+    /// [`AttachHandle`](metrique::writer::sink::AttachHandle) is dropped. If
+    /// the handle is [`forgotten`](metrique::writer::sink::AttachHandle::forget),
+    /// the reporter runs indefinitely.
+    ///
+    /// If no sink has been attached yet, entries are silently discarded until
+    /// one is attached.
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
```
jlizen: Not understanding why this can't be `no_run` with whichever metrique dependencies imported?

metrique/docs/sinks.md
```diff
@@ -222,11 +222,21 @@ Available integrations (via [`metrique-util`]):
   `metrique-util` and call [`subscribe_tokio_runtime_metrics`] to
   start appending [`RuntimeMetrics`] snapshots (worker
   utilization, queue depths, poll durations, and more).
+- **System metrics** — enable the `sysinfo-bridge` feature on
+  `metrique-util` and call [`subscribe_sysinfo_metrics`] to start
```
jlizen, line 226: Can we get this down to two or three example metrics, and "and more"? Listing everything is a bit verbose and gets stale.

metrique-util/README.md
```diff
@@ -6,9 +6,11 @@ Additional utilities for [metrique].
 
 - `state`: Provides [`State<T>`], an atomically swappable shared value with snapshot-on-first-read semantics. Useful for shared runtime state (feature flags, config reloads, routing tables) that should appear on every metric record.
 - `tokio-metrics-bridge`: Subscribes [tokio-metrics] runtime snapshots to a global entry sink. The reporter task is automatically aborted when the `AttachHandle` is dropped.
+- `sysinfo-bridge`: Subscribes [sysinfo] system and current-process snapshots to a global entry sink — CPU usage, memory and swap, load average, uptime, and current-process memory/CPU/disk/runtime/open-file counters. Disk space, network rx/tx, and thermal sensor metrics are opt-in via `SysinfoMetricsConfig::with_disks` / `with_networks` / `with_components`. The reporter task is automatically aborted when the `AttachHandle` is dropped.
```
jlizen:
```suggestion
- `sysinfo-bridge`: Subscribes [sysinfo] system and current-process snapshots to a global entry sink, capturing metrics like CPU usage, disk space, and network rx/tx. The reporter task is automatically aborted when the `AttachHandle` is dropped.
```

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+            SysinfoMetricsConfig::default()
+                .with_interval(Duration::from_millis(50))
+                .with_networks(),
+        );
+
+        tokio::time::sleep(Duration::from_millis(500)).await;
+
+        let entry = inspector.entries().last().cloned().unwrap();
+        check!(entry.metrics.get("disk_count").is_none());
+        check!(entry.metrics["network_interface_count"] >= 0);
```
jlizen: Was this showing up as 0 in tests?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+        Sink::subscribe_sysinfo_metrics(
+            SysinfoMetricsConfig::default()
+                .with_interval(Duration::from_millis(50))
+                .with_disks(),
+        );
+
+        tokio::time::sleep(Duration::from_millis(500)).await;
+
+        let entry = inspector.entries().last().cloned().unwrap();
+        check!(entry.metrics["disk_count"] >= 0);
```
jlizen: Was this showing up as 0 in tests?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+        }
+    });
+    let abort = worker.abort_handle();
+
+    // Spawn a monitor to log panics
+    tokio::spawn(async move {
+        if let Err(err) = worker.await
+            && !err.is_cancelled()
+        {
+            tracing::error!("sysinfo metrics reporter panicked: {err}");
```
jlizen: Better to use `{err:#}` to capture any future error chains

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+            agg.network_total_errors_on_received += data.total_errors_on_received();
+            agg.network_errors_on_transmitted += data.errors_on_transmitted();
+            agg.network_total_errors_on_transmitted += data.total_errors_on_transmitted();
+        }
+        agg
+    });
+
+    let components_metrics = components.map(|c| {
+        c.refresh(true);
+        let (max_cur, max_recorded) = c.list().iter().fold((0.0f32, 0.0f32), |(cur, max), comp| {
```
jlizen: Why cast to 0 here rather than initializing as None? Looks like another 0 default.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+            (
+                cur.max(comp.temperature().unwrap_or(0.0)),
+                max.max(comp.max().unwrap_or(0.0)),
+            )
+        });
+        ComponentMetrics {
```
jlizen, line 403: I'm somewhat uncomfortable with this compression, we probably want to capture WHICH component was warmest, but also probably if somebody is enabling temperatures as an opt-in thing, probably they want to see the distinct values per component per cycle.

I guess that's pretty obnoxious today without something like MultiFlex, but maybe we could at least include the component name as a property for max temperature and max recorded?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+    });
+    let abort = worker.abort_handle();
+
+    // Spawn a monitor to log panics
+    tokio::spawn(async move {
+        if let Err(err) = worker.await
+            && !err.is_cancelled()
+        {
+            tracing::error!("sysinfo metrics reporter panicked: {err}");
+        }
```
jlizen, line 515: We should emit a metric on panics no? Probably fine to wait until #205 though and leave behind a TODO to use whichever strategy.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+    let track_disks = config.track_disks;
+    let track_networks = config.track_networks;
+    let track_components = config.track_components;
+    let worker = tokio::spawn(async move {
+        tracing::debug!("sysinfo metrics reporter started");
+        let mut system = System::new();
+        let mut disks = track_disks.then(Disks::new_with_refreshed_list);
+        let mut networks = track_networks.then(Networks::new_with_refreshed_list);
+        let mut components = track_components.then(Components::new_with_refreshed_list);
+        let pid = sysinfo::get_current_pid().ok();
```
jlizen: Should this be a debug or trace log if unable to get a pid? What are the cases where it is expected?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+
+        total_memory: system.total_memory(),
+        used_memory: system.used_memory(),
+        available_memory: system.available_memory(),
+        free_memory: system.free_memory(),
+        total_swap: system.total_swap(),
+        used_swap: system.used_swap(),
+        free_swap: system.free_swap(),
+
+        load_average_one: LOAD_AVERAGE_SUPPORTED.then_some(load.one),
```
jlizen, line 423: It makes me a bit nervous that we have no unit tests for these load numbers, was there not a good way to test them (that wouldn't involve way too long tests)? Could you share some info about what manual testing you have done? Ideally an affirmation that we saw non-0 values output for the ones that only have >= 0 assertions would be good too.

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+
+/// Component (thermal sensor) metrics. Emitted as part of [`SysinfoMetrics`]
+/// when [`SysinfoMetricsConfig::with_components`] is set.
+#[metrics(subfield_owned)]
+#[non_exhaustive]
+pub struct ComponentMetrics {
+    /// Number of thermal/component sensors being tracked.
+    pub component_count: u64,
+    /// Maximum current temperature across all components, in degrees Celsius.
+    /// `0.0` if no component reports a temperature.
```
jlizen: We want to have these cases be None no?

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+    pub component_max_temperature: f32,
+    /// Maximum recorded temperature across all components, in degrees Celsius.
```
jlizen: Is this is a gauge that captures the lifetime of the process? Whereas the previous is per cycle? Wording is a bit confusing

jlizen (changes requested):
Looking good, mostly some questions around testing.

The main load bearing feedback is the remaining 0 default on the thermals (and adding the relevant component name as a property).

jlizen:
Oh, also - this panics if called outside a tokio context yes? We should document that clearly (probably also should be for tokio-metrics bridge, but it is more self evident).

Ideally we would have support for a std thread based worker too though I don't think it is critical initial scope. Do you think that we should have a defensive constructor entry point to make it easier to add later @rcoh? Ie have it be a builder or have the single function entry point mention the tokio worker in the name?

metrique-util/src/sysinfo_reporter.rs
fmzbl: ill create one that works

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,724 @@
...
+fn spawn_sysinfo_metrics_task(
+    sink: BoxEntrySink,
+    config: SysinfoMetricsConfig,
+) -> tokio::task::AbortHandle {
+    let interval = config.interval;
+    let name_style = config.name_style;
+    let track_disks = config.track_disks;
+    let track_networks = config.track_networks;
+    let track_components = config.track_components;
+    let worker = tokio::spawn(async move {
```
rcoh: we cannot spawn this onto a tokio task. sysinfo routinely blocks for multiple milliseconds.

this should either be a `spawn_blocking` or a std::thread::thread

one reasonable option could be using `spawn_blocking` in a tokio context and falling back to a std::thread outside of a tokio context.

rcoh (changes requested):

metrique-util/src/sysinfo_reporter.rs
fmzbl: it just checks if its present, i can maybe test the presence of the 3 categories

metrique-util/src/sysinfo_reporter.rs
fmzbl: it just checks if its present, im gonna check the whole prefix

metrique-util/src/sysinfo_reporter.rs
fmzbl: you are right, ill make it an option

metrique-util/src/sysinfo_reporter.rs
fmzbl: sure, good idea

metrique-util/src/sysinfo_reporter.rs
fmzbl: i ran the example in extras.rs and checked the values were non 0, ill add a test to check for them

metrique-util/src/sysinfo_reporter.rs
fmzbl: yes, but ill improve the docs

jlizen (approved):
Thanks for all the iterations! looks ready to me

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -0,0 +1,842 @@
...
+                .copied()
+                .or_else(|| payload.downcast_ref::<String>().map(String::as_str))
+                .unwrap_or("<non-string panic>");
+            tracing::error!("sysinfo metrics reporter panicked: {msg}");
+        }
+    };
+
+    if Handle::try_current().is_ok() {
+        tokio::task::spawn_blocking(worker);
+    } else {
```
rcoh, line 520: A small improvement I think you can call spawn_blocking on the handle itself if I'm not mistaken

rcoh (approved):

rcoh (approved):

PR #284, opened 2026-05-06
make worker sink exit

fmzbl:

### DIFF
--- a/metrique-aggregation/src/sink/worker.rs
+++ b/metrique-aggregation/src/sink/worker.rs
@@ -3,7 +3,7 @@
 use std::{
     marker::PhantomData,
     sync::Arc,
-    sync::mpsc::{Sender, channel},
+    sync::mpsc::{RecvTimeoutError, Sender, channel},
     thread,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


rcoh (approved):

metrique-aggregation/src/sink/worker.rs
```diff
@@ -96,3 +100,34 @@ where
...
+        let sink = WorkerSink::<(), _>::new(NoopSink, Duration::from_secs(60));
+        let handle = Arc::clone(&sink._handle);
+        drop(sink);
+
+        let deadline = Instant::now() + Duration::from_secs(5);
+        while !handle.is_finished() {
+            if Instant::now() >= deadline {
+                panic!("worker thread still running 5s after senders were dropped");
+            }
+            thread::sleep(Duration::from_millis(10));
```
rcoh, line 129: This is a race condition, you need to validate that the flush actually happened, this test will be flaky

jlizen: Ah sorry I thought it was just needing the merge button. Will need a separate PR for the fix @fmzbl

Issue #110, opened 2025-09-24
Add OTel support


rcoh:
Tracking issue for OTel support. This will require adding the ability to have "aggregation hints" (via flags, probably, similar to the emf::HighResolution) so that we know which type of OTel metric to create.

jlizen:
Some thoughts:

Some useful docs:
- https://opentelemetry.io/docs/concepts/signals/metrics/
- https://opentelemetry.io/docs/specs/otel/metrics/data-model
- https://docs.rs/opentelemetry/latest/opentelemetry/#how-metrics-work-in-opentelemetry

Since russell made this issue, we've added an aggregation api that has a good amount of overlap: 
- https://docs.rs/metrique-aggregation/0.1.9/metrique_aggregation/#sink-level-aggregation

It supports driving aggregation sinks with various strategies, which would be similar to the otel meters. 

And then, for instruments, we have a fair amount of primitives that map cleanly over (our `Counter` is the same as theirs, for instance). 

We'll want to understand more about the semantics of grouping multiple instruments in otel emissions, since we normally want to emit all our metrics in a bundle. We also will want to see how we can attach properties to requests, or if that even makes sense with the units that otel allows.

jlizen:
It's not clear to me how useful it is to just write to disk.

I believe the collector does have a way to read from disk, but it is less common than pushing directly to the collector over a socket.

Ref: https://opentelemetry.io/docs/collector/architecture/

We should do a little research to see what all is involved. If we need to do our own exporting, that probably wants #225 . 

One option for that would be using `opentelemetry-proto` and managing our own connection to the collector.

`opentelemetry-otlp` exists but I suspect it would be pretty difficult to use outside of the `opentelemetry` input API.

jlizen:
Another consideration (which I think Russell was alluding to with the Format config) is supporting cumulative vs delta encoded metrics. I imagine that delta encoded might involve building some new capabilities into `metrique-aggregation` as that is not a strategy that we currently support?

rcoh:
The hinting needed is picking which otel primitive we are emitting to. You can "guess" from the aggregation strategy but I think it would probably be cleaner to provide a flag that we wrap during aggregation so that the formatter knows what OTel type to publish

Fluzko:
I've did a first pass and trying to tackle users-interface first, so then we can start deep-diving into specifics per topic.


## Base user interface

High-level overview of the user interface expected.

```rust

use std::time::Duration;
[...]

Issue #100, opened 2025-09-11
It is unclear why one want to use metrique over other alternatives


weihanglo:
The README and the the first page of <https://docs.rs/metrique/0.1.5/metrique/> explains a lot around how to use metrique. However, it says little about why we should, like performance wins, and better structural metrics definitions.

We might want to have a paragraph in the very front of the README to show the pros and cons and tradeoff of using metriques, and optionally comparing with other metric libraries.

PR #288, opened 2026-05-12
OTel integration design

Fluzko:
📬 *Issue #, if available:*
#110 

✍️ *Description of changes:*
This is a follow up to the plan proposed in a comment at #100, with some adressed comments from both @jlizen & @rcoh. 
There's a coded-approach, reflecting this proposal as a draft in [here](https://github.com/jlizen/metrique/pull/1)

### DIFF
--- a/docs/otel-plan-status.md
+++ b/docs/otel-plan-status.md
@@ -1,9 +1,11 @@
-# `metrique-otel` — current status and plan
+# `metrique-otel` —
+
+current status and plan
 
 Follow-up to the [original design comment](https://github.com/awslabs/metrique/issues/110#issuecomment-4400001687)
 and both [Jess's](https://github.com/awslabs/metrique/issues/110#issuecomment-4422783595) and [Russell's](https://github.com/awslabs/metrique/issues/110#issuecomment-4412942585) feedback.
-Captures where the `otel-integration` branch landed, what's intentionally deferred,
-and what changes once [#282 (Entry Descriptors)](https://github.com/awslabs/metrique/pull/282)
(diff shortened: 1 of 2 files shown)


docs/otel-plan-status.md
```diff
@@ -0,0 +1,129 @@
...
+
+#[metrics]
+#[derive(Default)]
+struct RequestMetrics {
+    time: Timer,
+    operation: String,                  // metric attribute on every observation
+    request_count: Counter<u64>,        // OTel Counter::add
+    queue_depth:   UpDownCounter<i64>,  // OTel UpDownCounter::add
+    latency_ms:    Histogram<Duration>, // OTel Histogram::record
+    cpu_usage:     Gauge<f64>,          // OTel Gauge::record
```
rcoh: generally, we will have better UX if these are attached during the CloseValue (so these wrappers only apply to the `Entry` type, not the user facing metrics type.

rcoh (commented):
seems like this basically makes sense. Starting w/ an MvP that uses the off-the-shelf OTel APIs is probably reasonable. The only reason not to is the cases when they are fundamentally the wrong shape for what we actually need to emit metrics efficiently

Fluzko:
Made some progress thanks to #282 over [my draft](https://github.com/Fluzko/metrique/pull/3). Updates [@7685](https://github.com/awslabs/metrique/pull/288/commits/7685416cf10982207522a9cad1dfd61b1c03745c)

rcoh:
Left some brief thoughts on the impl. I'm directionally aligned

Issue #160, opened 2026-01-01
[macro] No inheritance of container-level rename_all and prefix to flattened children


jlizen:
# Container-level prefix/rename_all not applying to flattened subfields

## Current Behavior

When a parent struct has a container-level `prefix` or `rename_all` attribute, these attributes do **not** apply to fields from flattened subfield structs. Only field-level attributes on the flatten apply.

### Example

```rust
#[metrics(subfield)]
struct Child {
    child_field: u32,
}

#[metrics(prefix = "api_")]
struct Parent {
    #[metrics(flatten)]
    child: Child,
    direct_field: u32,
}
```

**Current output:**
- `child_field` (no prefix)
- `api_direct_field` (prefixed)

**Expected output:**
- `api_child_field` (prefixed)
- `api_direct_field` (prefixed)

## Workaround

Reproduce the container prefix as a field-level prefix instead:

```rust
#[metrics(prefix = "api_")]
struct Parent {
    #[metrics(flatten, prefix = "api_")]
    child: Child,
    direct_field: u32,
}
```

## Impact

- Inconsistent behavior: container-level attributes apply to direct fields but not flattened fields
- Verbose workarounds required when you want consistent prefixing across all fields
- Affects **all** flattened structs (both with and without `subfield`/`subfield_owned`)

### `rename_all`
Flattened subfields should inherit the parent's `rename_all`, but the child's own container-level `rename_all` takes precedence if specified.

### `prefix`
All prefixes should be chained in order: `parent_container_prefix` + `flatten_field_prefix` + `child_container_prefix`.

This would make the behavior more intuitive and consistent.

jlizen:
I will maintain the same behavior for now in entry enums, ref #156

jlizen:
But, it seems like rename_all clearly should be inherited? I think that is just a bug to leave it out even if it is technically breaking existing behavior. Without that, you have to annotate every subfield entry. You only really should have to annotate root, when are the two ever going to diverge...

And if they do for some reason, you can let child override with cleaner semantics than prefix merging.

arielb1:
The reason prefix is not inflected is that prefix-on-containers existed before we had transitive prefix support, and we are not up for releasing a metrique 0.2 for that (at least it's only an update to metrique, not metrique-core)

jlizen:
I am not sure where I was seeing rename_all support not being transitive, I agree that it looks an awful lot like it does.

[...]

Issue #30, opened 2025-07-21
Support setting EMF Storage resolution and NoMetric from metrics via macro


rcoh:
EMF has the option for `HighStorageResolution`: https://docs.rs/metrique-writer-format-emf/latest/metrique_writer_format_emf/type.HighStorageResolution.html

However, we don't properly implement `CloseValue` for `ForceFlag` which would allow it to be used from structs with `#[metrics]`

1. Implement `CloseValue` for `ForceFlag`
2. Consider adding a macro attribute like `emf::high_resolution` to do this transparently.

Once 1) is done, the following code should work:

```rust
#[metrics(
    rename_all = "PascalCase",
    emf::dimension_sets = [
        ["OperationName"]
    ]
)]
pub struct SimpleHighResolutionMetric {
    /// The name of the operation being measured
    operation_name: String,
    
    #[metrics(unit = Count)]
    event_count: HighStorageResolution<Counter>,
}
```

arielb1:
(1) has been done (#33). Can we close this issue?

rcoh:
```rust
pub struct SimpleHighResolutionMetric {
    /// The name of the operation being measured
    operation_name: String,
    
    #[metrics(unit = Count, flag = HighStorageResolution)]
    event_count: Counter,
}
```

when the entry is generated we generate `HighStorageResolution<Counter::Closed>`

PR #289, opened 2026-05-13
feat: Entry descriptors

jlizen:
Follows  #290 which added the write-path flag mechanism (`ForceFlag`/`FlagConstructor`).

This is most of the `metrique`-side groundwork for the dial9 integration (besides enumerating native field shapes).

### Summary

Implements the entry descriptor system described in `docs/entry-descriptors.md`. Sinks can now call `Entry::descriptors()` to get compile-time structural metadata (field names, flags, units, timestamp) without observing a live write.

Deviations from the design doc:

- **All shapes are `Opaque`.** `FieldShape` variants exist but the macro always emits `Opaque`. In a subsequent PR, I will add that in shapes (Scalar, Optional, Distribution).
- **Flatten-site `default_flags`** is an addition not in the original design. `#[metrics(flatten, default_flags(X))]` lets a parent apply flags to a library child's fields. The child can still opt out with `flags(skip(X))` at field level. This required `skipped_flags` on `FieldDescriptor` and `extra_flags` on `DescriptorRef`. I think it is the right choice, since otherwise a caller has no way to enable dial9 emission on a field they are flattening into their struct, containing a foreign type.
- **Single static per entry** rather than one per style. Each `FieldDescriptor` stores a names array indexed by the parent's style. Simpler, smaller binary, same runtime cost. I also added const `Styles` struct to encapsulate ordering, style names, etc.

Includes a descriptor-aware sink e2e test (`descriptor_sink.rs`) exercising the intended consumption pattern. Also, a guide recipe in `Extending metrique` docs. No user-facing descriptor-aware sinks ship yet, dial9 will be the first.

What's left:
- Typed `FieldShape`
- (Further off, breaking change) migrate native sinks to descriptor API for flags, deprecate write-path `MetricFlags`

### Migration path

This is additive. Existing sinks and formats are unaffected: `Entry::descriptors()` has a defaulted impl returning `Unavailable`, so nothing breaks.

For sink authors using flags or other descriptor features:
1. Call `entry.descriptors()` and handle `Unavailable` gracefully (fall back to write-path-only behavior).
2. Read flags from `FieldView::flags()` rather than downcasting `MetricFlags` at write time.
3. Cache per-entry-type work (schema registration, field mappings) keyed on `DescriptorId`.

In a future major version, we can migrate native sinks/formats like EMF and Otel to use `descriptors()` rather than requiring explicit `ForceFlags` in field types. This mostly improves ergonomic for hand (non-macro) implemented Entry structs. That would be a mildly breaking change since it will default to no available descriptors, and ignore field-specified `ForceFlags`. It is not urgent and we should assess consumer usage first.

### Open question

Previously the #[aggregate] macro generated a key struct that only had a bare `#[metrics]`, so key fields (dimensions) kept their Rust field names even when `rename_all` was set. I consider this a bug.

The `#[aggregate]` macro now forwards the parent's `#[metrics(...)]` attribute (including `rename_all`) to the generated key struct.  Now they follow the same convention as non-key fields.

This seems more consistent, but it changes the wire names of dimensions for anyone using `#[aggregate]` with `rename_all`. Should we preserve the old behavior for the key struct and only forward non-rename attributes? And add it to the heap similar to https://github.com/awslabs/metrique/issues/160?


metrique-core/src/inflectable_entry_impls.rs
```diff
@@ -33,6 +37,14 @@ impl<NS: NameStyle, T: InflectableEntry<NS>> InflectableEntry<NS> for Option<T>
             itertools::Either::Right([].into_iter())
         }
     }
+
+    fn descriptors(&self) -> impl Iterator<Item = DescriptorRef<'_>> {
+        if let Some(entry) = self.as_ref() {
+            itertools::Either::Left(entry.descriptors())
+        } else {
+            itertools::Either::Right(std::iter::empty())
+        }
```
rcoh: I think `Option` already has an iter impl on it

metrique-core/src/namestyle.rs
```diff
@@ -39,6 +39,28 @@ pub trait NameStyle: private::NameStyleInternal {
     type InflectAffix<ID: MaybeConstStr, PASCAL: MaybeConstStr, SNAKE: MaybeConstStr, KEBAB: MaybeConstStr>: MaybeConstStr;
 }
 
+// Style index constants used by descriptor codegen to select the right static.
+// The macro crate mirrors these in `metrique-macro/src/inflect.rs` (DESCRIPTOR_STYLES
+// and DESCRIPTOR_STYLE_NAMES). Both must stay in sync.
+#[doc(hidden)]
+pub const STYLE_PRESERVE: u8 = 0;
```
rcoh: metrique-macro can depend on metrique-core, I don't think there is a need to duplicate this.

metrique-core/src/namestyle.rs
```diff
@@ -39,6 +39,28 @@ pub trait NameStyle: private::NameStyleInternal {
...
+pub const STYLE_PASCAL: u8 = 1;
+#[doc(hidden)]
+pub const STYLE_SNAKE: u8 = 2;
+#[doc(hidden)]
+pub const STYLE_KEBAB: u8 = 3;
+
+// Compile-time assertion: style constants must be sequential starting from 0.
+// If you add or reorder styles, update metrique-macro/src/inflect.rs
+// (DESCRIPTOR_STYLES, DESCRIPTOR_STYLE_NAMES, descriptor_index) to match.
+const _: () = {
```
rcoh: this isn't testing anything useful

metrique/tests/ui/fail/conflicting_field_tags.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use metrique::unit_of_work::metrics;
+
+struct MyTag;
+
+// Conflicting field_tag: same tag both present and skipped on one field
+#[metrics]
+struct ConflictingFieldTag {
```
rcoh: field_tag seems somewhat orthogonal to the main PR?

metrique-writer-core/src/entry/mod.rs
```diff
@@ -185,6 +185,15 @@ pub trait Entry {
         MergedRef(self, other)
     }
 
+    /// Returns descriptors for this entry in write order.
+    ///
+    /// Each descriptor covers a contiguous segment of the `Entry::write` output.
+    /// Simple entries yield one descriptor. Composed entries (like aggregation results)
+    /// yield multiple. Hand-written entries return an empty iterator by default.
+    fn descriptors(&self) -> impl Iterator<Item = DescriptorRef<'_>> {
```
rcoh: I wonder if we need a sentinel that we are hitting this default impl eg. `iter::once(DescriptorRef::undefined())`

it may make it easier to debug issues when descriptors are not present because of a manual entry impl especially given how we combine descriptors.

metrique-macro/src/inflect.rs
```diff
@@ -31,6 +31,31 @@ pub(crate) enum NameStyle {
 }
 
 impl NameStyle {
+    /// Ordered array matching the `STYLE_*` index constants in `metrique-core`.
+    /// Index 0 = Preserve, 1 = PascalCase, 2 = SnakeCase, 3 = KebabCase.
+    /// See also: `metrique-core/src/namestyle.rs` STYLE_* constants.
+    pub(crate) const DESCRIPTOR_STYLES: [NameStyle; 4] = [
```
rcoh: metrique-macro can actually just depend on metrique-core

rcoh (commented):
given the non-trivial risk of emitting a broken descriptor (e.g. that doesn't compile) for some complicated struct in the wild, I wonder if an opt-out or temporary opt-in flag for generating a non-trivial descriptor impl is prudent.

metrique/tests/ui/fail/conflicting_field_tags.rs
jlizen: I've updated to extend the pre-existing `ForceFlags` infrastructure. I wanted to make sure our approach would work for it given all the operator chaining, and I also wanted to get tags out for Facu to use with Otel.

metrique-writer-core/src/entry/mod.rs
jlizen: Agreed, I've updated it to force users to always interactive with a descriptors enum that treats the OVERALL descriptor chain as unavailable if any individual segment is unavailable. This also keeps ergonomics decent since you can just match on the enum variant once, and then get the inner regular iterator.

```
pub enum Descriptors<'a> {
    /// Descriptors are available for this entry.
    Available(AvailableDescriptors<'a>), // a custom iterator
    /// This entry has not implemented descriptor support.
    Unavailable,
}
```

metrique-core/src/inflectable_entry_impls.rs
jlizen: Thanks, fixed the existing sample group ones as well. Note quite enough to get rid of itertools... close...

metrique/tests/ui/fail/enum_cfg_fields.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use metrique::unit_of_work::metrics;
+
+// cfg on any field inside a struct enum variant is not supported
```
jlizen, line 6: This was already broken for enum writes, I just added a custom error to make it clearer than the arcane rustc error.

It is not trivial to fix compared to struct arms since we have to access fields by destructuring in match. We would have to do something like, generate opposing match arms for every cfg combination, which quickly gets combinatorial and might not be a good idea.

Let's wait until anybody asks for this before looking at it more.

jlizen:
@rcoh ready for new eyes, see comments above and below. Here's the change digest:

## Notes on feedback

1. **Default impl sentinel** (rcoh): `Entry::descriptors()` and `InflectableEntry::descriptors()` both return `Descriptors` enum (`Available`/`Unavailable`). Callers must match. `Descriptors::chain()` propagates `Unavailable` through flatten composition.

2. **Option .iter()** (rcoh): Replaced `Either` with `.iter().flat_map()` / direct forwarding for `descriptors()` and `sample_group()`.

3. **Style constant duplication** (rcoh): Macro depends on `metrique-core`. Style selection uses `NameStyle::DESCRIPTOR_STYLE_INDEX` (const on each type). No parallel arrays.

4. **field_tag orthogonality** (rcoh): Replaced with `flags(X, Y, skip(Z))` integrated with `FlagConstructor`. `#[metrics(flags(X))]` records in descriptor AND wraps in `ForceFlag` at write time. Addresses #30.

5. **metrique-macro depends on metrique-core** (rcoh): Done.

## Additional changes

- **MSRV 1.91**: Required for `const TypeId::of()`.
- **Builder API**: `EntryDescriptor::builder(name, fields)`, `FieldDescriptor::builder(name)`, `TimestampDescriptor::new(name)`, `FieldFlag::new::<T>()`. No `__metrique_private_new` anywhere.
- **Flags scoping**: `default_flags(X)` applies to direct fields only. `flags(...)` on flatten is a compile error. `skip(...)` in `default_flags` is a compile error.
- **Style propagation**: Child's `rename_all` takes precedence. Parent's style propagates to children without their own `rename_all`. Descriptor field names match write-path output.
- **Opaque API**: `AvailableDescriptors`, `DescriptorIter` hide internal SmallVec. `Descriptors::chain()` and `map_available()` for composition.
- **Descriptor forwarding**: All Entry/InflectableEntry wrapper impls (`&T`, `Box`, `Arc`, `Cow`, `Option`, `ForceFlag`, `Merged`, `DynamicInflectionEntry`) properly forward. `NoKey`/`EmptyEntry` return Available-empty. Maps return Unavailable (documented).
- **EMF integration**: `emf::flags::{HighStorageResolution, NoMetric}` re-exports. Tests verify descriptor + write-path + EMF JSON output.
- **FlagsList parsing**: Uses `from_meta` with direct token parsing (darling workaround for Flag field interaction, documented in code).

## Intentionally out of scope

- Field shapes (all `Opaque`)
- `entry_name` attribute
- User-facing sink authoring guide

docs/entry-descriptors.md
```diff
@@ -1,33 +1,45 @@
...
 - **`FieldShape`**: the closed/emitted shape of a field (scalar, optional, list, dynamic-key map, or opaque). Describes what the sink will see, not the raw Rust type.
-- **`DescriptorRef`**: the handle returned by `Entry::descriptor()`. Opaque; carries a stable `DescriptorId` for cache keying and a borrow of the underlying `EntryDescriptor`.
+- **`DescriptorRef`**: the handle yielded by `Entry::descriptors()`. Provides field access via `FieldView`, carries a stable `DescriptorId` for cache keying.
 - **`DescriptorId`**: an opaque identifier for a descriptor, stable within a single process lifetime. Used by sinks to cache derived data.
 
+### Relationship between flags, FlagConstructor, ForceFlag, and MetricFlags
+
+A **`FlagConstructor`** is a type whose `construct()` method returns a `MetricFlags<'static>` value. Format crates (like EMF) define these to represent format-specific options (e.g. `HighStorageResolutionCtor`). They also provide a `flags` module re-exporting the constructor under a user-friendly name (e.g. `emf::flags::HighStorageResolution`).
```
jlizen: I'm not sure if we'll want to "re-export" in future cases versus just not having the ugly `HighStorageResultionCtor` (which stands for constructor)... anyway...

docs/entry-descriptors.md
```diff
@@ -1,33 +1,45 @@
...
+
+**`#[metrics(flags(T))]`** where `T: FlagConstructor` does two things:
+1. Records `T`'s `TypeId` in the field's descriptor (so sinks can introspect it via `FieldView::flags()`)
+2. Wraps the field's value in `ForceFlag<_, T>` at write time (so the flag flows through `ValueWriter::metric()` to the format)
+
+**`ForceFlag<V, T>`** is the write-path mechanism. It intercepts `Value::write` and injects `T::construct()` into the `MetricFlags` parameter of `ValueWriter::metric()`. Using `ForceFlag<V, T>` as a field type directly only affects the write path; the descriptor is unaware of it. Prefer `#[metrics(flags(T))]` to get both.
```
jlizen: I'm still uncomfortable with the idea of "you only get the flag in the descriptor if you use macro attribute"... But, I do think if we replace all examples with using the macro, that's what everybody will do. And, none of our existing tags are sensitive to not being in the descriptor, for legacy code.

I'll be sure to demonstrate properly setting tags in the custom entry impl that has a descriptor impl, in a coming PR, which I can throw one of these macro-managed tags into for clarity on how do it properly.

jlizen: Thought about this more. I am thinking now to have newer sinks use the entry descriptor api on the write path, but leave old sinks on the uglier kind where you need the explicit type API.

And then, next major version, we can cut old sinks over to the entry descriptor API. The write path handling is buggy anyway, you can't have multiple types of flags at least for EMF.

jlizen:
@rcoh this is ready for fresh eyes, PR description updated. Note the open question on whether we should bugfix the lack of rename_all propagation to `#[aggregate]`-generated Key entries.

CHANGELOG.md
```diff
@@ -7,88 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [0.1.26](https://github.com/awslabs/metrique/compare/metrique-v0.1.25...metrique-v0.1.26) - 2026-06-09
```
rcoh, line 10: Looks like an accidental delete of a bunch of changelog entries

metrique-aggregation/src/traits.rs
```diff
@@ -321,6 +321,16 @@ impl<Ns: NameStyle, A: InflectableEntry<Ns>, B: InflectableEntry<Ns>> Inflectabl
...
         self.aggregated.write(w);
     }
+
+    fn descriptors(&self) -> metrique_writer_core::Descriptors<'_> {
+        match (self.key.descriptors(), self.aggregated.descriptors()) {
+            (
+                metrique_writer_core::Descriptors::Available(a),
+                metrique_writer_core::Descriptors::Available(b),
+            ) => metrique_writer_core::Descriptors::available(a.into_iter().chain(b)),
+            _ => metrique_writer_core::Descriptors::Unavailable,
```
rcoh, line 335: Is this the right merging?

metrique-aggregation/tests/descriptor.rs
```diff
@@ -0,0 +1,105 @@
...
+impl MetricOptions for ExportOpt {}
+struct Export;
+impl FlagConstructor for Export {
+    fn construct() -> MetricFlags<'static> {
+        MetricFlags::upcast(&ExportOpt)
+    }
+}
+
+#[aggregate]
+#[metrics(rename_all = "PascalCase", default_flags(Export))]
```
rcoh, line 23: Agree we should fix rename all. Could mv it without much pain and prevent breaking anyone surprisingly

metrique-aggregation/tests/descriptor.rs
```diff
@@ -0,0 +1,105 @@
...
+struct RequestMetrics {
+    #[aggregate(key)]
+    #[metrics(flags(skip(Export)))]
+    operation: &'static str,
+    #[aggregate(strategy = Sum)]
+    count: u64,
+}
+
+#[test]
+fn aggregation_result_yields_two_descriptors() {
```
rcoh, line 33: Side note: we should be able to use this to improve my PR to respect dimensions by handling dimensions via entry descriptors

metrique-macro/src/inflect.rs
```diff
@@ -139,6 +149,20 @@ mod test {
     use super::name_contains_uninflectables;
     use crate::{NameStyle, inflect::name_ends_with_delimiter};
 
+    #[test]
+    fn descriptor_styles_exhaustive() {
+        // Exhaustive match: adding a new NameStyle variant causes a compile error here,
+        // forcing you to update NameStyle::ALL and descriptor_index().
+        fn _assert_exhaustive(s: NameStyle) -> usize {
+            match s {
+                NameStyle::Preserve => 0,
```
rcoh, line 170: Seems like this test should take the index and look it up in all?

metrique-macro/src/lib.rs
```diff
@@ -38,7 +38,7 @@ use crate::inflect::{name_contains_dot, name_contains_uninflectables, name_ends_
 /// | Attribute | Type | Description | Example |
 /// |-----------|------|-------------|---------|
 /// | `rename_all` | String | Changes the case style of all field names | `#[metrics(rename_all = "PascalCase")]` |
-/// | `prefix` | String | Adds a prefix to all field names (prefix gets inflected) | `#[metrics(prefix = "api_")]` |
+/// | `prefix` | String | Adds a prefix to all direct field names (prefix gets inflected). Does not propagate to flattened children. | `#[metrics(prefix = "api_")]` |
```
rcoh, line 41: Agree this would be nice to fix but not here

rcoh (commented):
Seems good!

CHANGELOG.md
jlizen: thanks, rebase misfire

metrique-aggregation/tests/descriptor.rs
jlizen: Good point, just is a metrique-aggregation bump, will do

metrique-macro/src/lib.rs
jlizen: (And to be clear not new behavior, just improved docs -  https://github.com/awslabs/metrique/issues/160)

docs/entry-descriptors.md
```diff
@@ -185,153 +124,131 @@ Flattening an `Option<SubEntry>` into a parent entry propagates optionality to e
 
 `#[metrics(ignore)]` fields are not part of the descriptor. They do not emit, do not close, and do not appear in `fields()`.
 
-### The Opaque trapdoor
+### The Opaque shape
```
Fluzko, line 127: This is the one item that gates a descriptor-driven OTel sink, so flagging it early.

Otel histograms arrive 2 ways today:
1. A user `flags(otel::Histogram)` tag on a field. This shows up in the descriptor as a flag, so it is fine.
2. The canonical path: `KeyedAggregator` -> `Histogram<T>`, where the closed value emits the runtime `Distribution` flag and our translator maps that to a histogram instrument. This is the common case and it is what most users hit.
 
2nd path is invisible to descriptors: the field lowers to `FieldShape::Opaque`, and `Distribution` is a property of the closed value at write time, not a user-applied flag. So a sink that resolves instrument kind purely from `descriptors()` silently misses aggregated histograms, which is, arguably, our highest-traffic shape.

Practical consequence woul be that until `FieldShape::Distribution` (or a `DescribeValue`/ `DescribeFlags` hook so the histogram value self-reports as distribution-shaped) lands, an otel migration has to stay hybrid for descriptors for kind/unit/attributes, write-path `Distribution` for the
aggregated histogram case. That is workable, but it means "migrate native sinks to descriptors
and deprecate write-path flags" cannot fully cover otel until this exists (correct me).

Is `FieldShape::Distribution` + the `DescribeValue` trait the intended mechanism, and is there an ordering where it could come soon after this PR?

docs/entry-descriptors.md
```diff
@@ -185,153 +124,131 @@ Flattening an `Option<SubEntry>` into a parent entry propagates optionality to e
...
+Each enum variant gets its own static descriptor containing only that variant's fields (plus the tag field if present). The `descriptors()` method matches on self and yields the active variant's descriptor. Different variants produce different `DescriptorId`s. The descriptor name includes the variant (e.g., `"MyEnum::Read"`).
+
+For variants with flatten fields, the variant's base descriptor is followed by the flatten children's descriptors (same chaining pattern as structs). A generated enum iterator type (same pattern as sample_group) unifies the different return types across match arms.
 
-Per-variant descriptors (where each variant emits its own narrower schema) are a future evolution that requires `DescriptorRef` to back with `Arc<EntryDescriptor>` rather than `&'static`; the opaque handle leaves that open.
+Sinks see different descriptor sequences depending on which variant is active. Each segment has its own `DescriptorId`, so per-segment caching works naturally. Sinks that want a single cache key for the whole entry can hash the sequence of ids.
+
+### Aggregated entries
```
Fluzko, line 171: Descriptors expose shape and, for `AggregationResult`, the segment split (key segment first, aggregated segment second), but nothing labels a field or segment as a dimension vs a measurement. Today our sink uses a heuristic ("every string field becomes an attribute on all metrics"), which is lossy in both directions since a `u64` dimension cannot be an attribute, and a string cannot be a measurement.

With the segment split we can special-case `AggregationResult` (first segment = dimensions), but only if we already know the entry is an aggregation result. But generically a sink cannot tell "this segment is the key."

Fluzko: For instance, a segment "role" on the descriptor (`AggregationResult`'s key descriptor marks its fields as keys/dimensions). This is sink-agnostic and would benefit EMF dimension handling too (kind of?).

or is the intent that dimension selection stays a per-sink flag concern?

docs/entry-descriptors.md
```diff
@@ -185,153 +124,131 @@ Flattening an `Option<SubEntry>` into a parent entry propagates optionality to e
...
+
+`AggregationResult` writes key fields then aggregated fields. Its `descriptors()` implementation chains the key entry's descriptor followed by the aggregated entry's descriptor. Both are generated by `#[metrics]` on the respective structs; no additional descriptor generation is needed in the aggregate macro.
+
+Sinks walking `descriptors()` see two segments in write order: key fields first, aggregated fields second. Each segment's descriptor carries its own flags, units, and field names. Flatten on key fields is rejected at compile time.
 
 ### `Entry::write` order contract
```
Fluzko, line 177: The "exactly one `EntryWriter::value()` per `FieldDescriptor`, in order" guarantee is what would let a sink index its cached plan by position instead of by name. but it think it worth the clarifitcation of for a `None` optional field (or any field whose `Value` writes nothing), is `value()` still invoked (with the `ValueWriter` then calling neither `string()` nor `metric()`), or is the callback skipped entirely?

This is important for correctness of any position-indexed consumer.Because if `value()` is skipped for `None`, position indexing desyncs and sinks must key by name instead. If `value()` always
fires and only the inner `ValueWriter` no-ops, position indexing is safe as long as the sink counts `value()` calls rather than emitted metrics. 
Either is fine, but the contract should state which, since the doc currently implies strict 1:1 positional correspondence.

docs/entry-descriptors.md
```diff
@@ -45,128 +56,56 @@ pub struct Export;
...
+
+impl Entry for MyMetrics {
+    fn descriptors(&self) -> Descriptors<'_> {
+        Descriptors::available(std::iter::once(DescriptorRef::from_static(&DESC)))
+    }
+    // ...
 }
 ```
 
 ### Forward compatibility
```
Fluzko, line 100: This might not be that important, but it's about a performance need. `base_name()` returns `&'static str`, which lets us build otel attribute/instrument keys once and reuse them with no per-entry allocation (today we `to_owned()` the key on every attribute).
But "Forward compatibility" says accessors are deliberately conservative (borrows tied to the handle) to allow `Arc`-backed storage later, which would drop `'static` and break any consumer caching those keys.

Could names commit to `'static` (they already are in the generated statics), or get a dedicated `'static`/interned name accessor? Units and flags are cheap `Copy`. Worth to note that names are the one place the allocation cost is real for us.

Fluzko (commented):
Reviewing this from the perspective of a descriptor consumer not line-by-line on the macro. 
The foundation is exactly what an otel sink wants: reading instrument kind, unit, and field names from a static keyed by `DescriptorId` lets us build a per-entry-type "wire plan" once and drop the per-write `MetricFlags` downcast we do today in `OtelValueWriter::metric()`. `FieldFlag::construct()` returning our existing `MetricFlags` means we can adopt descriptors without changing our flag types at all, which is great.

Leaving some comments, one gap and a few asks before this is something otel can fully lean on, but love it 🚀

metrique-aggregation/src/traits.rs
jlizen: Yes, we intentionally remove access to descriptors if we see any opaque entries, since we lose all guarantees on field ordering. I'll comment more clearly.

docs/entry-descriptors.md
jlizen: Yep, it will be a fast follow, just trying to keep PR size contained

docs/entry-descriptors.md
jlizen: Yes, it is called with `None`, I'll document more clearly

docs/entry-descriptors.md
jlizen: Currently the accessor yield `&'static` via `base_name()`, so I think that's what you need. The more complex lifetimes are more for constructing the descriptor configuration via non macro-generated entries.

For instance, backing it with an arcswap (which would also yield `&'static str` on access

jlizen: Will document this more clearly

docs/entry-descriptors.md
jlizen: Today this would be handled via a flag, which if desired  could be owned by metrique for use across sinks. We could think about a more sophisticated API if desired.

jlizen:
Addressed feedback: https://github.com/awslabs/metrique/pull/289/commits/b3b9f6245f194369b219022d4f5328be4508b2b5

rcoh (approved):

PR #290, opened 2026-05-14
feat(macro): Add flags(...) macro attribute for per-field ForceFlag

### LINKED ISSUE #14: Support the `NoMetric` flag in metrique
Currently, the EMF format supports a flag [`NoMetric`](https://github.com/awslabs/metrique/blob/main/metrique-writer-format-emf/src/emf.rs#L1315) that allows making a field that would otherwise a metric into a property (effectively suppressing it from the metrics array in EMF). We need something like `#[metrique(property)]` to support this from the metrique frontend

### LINKED ISSUE #30: Support setting EMF Storage resolution and NoMetric from metrics via macro

jlizen:
📬 *Issue #, if available:*

Closes: #30 
Closes: #14

✍️ *Description of changes:*

Minimal subset of #289 to get tags working for OTel.

Adds `#[metrics(flags(...))]` field attribute support to the `#[metrics]` macro. This provides an alternative to wrapping field types in `ForceFlag<T, Ctor>`:

```rust
use metrique::emf::flags::{HighStorageResolution, NoMetric};

#[metrics(emf::dimension_sets = [["Op"]], rename_all = "PascalCase")]
struct Metrics {
    op: String,
    #[metrics(flags(HighStorageResolution))]
    latency: u64,
    #[metrics(flags(NoMetric))]
    debug_count: u64,
}
```

The macro wraps the field value in `ForceFlag` at write time during codegen. Multiple flags can be specified on a single field: `flags(A, B)`.

Also adds `emf::flags` module re-exports (`HighStorageResolution`, `NoMetric`) providing ergonomic names for the `FlagConstructor` types.

**Deferred to keep scope minimal:**
- Container-level `default_flags(...)` (apply a flag to all fields in a struct)
- `flags(...)` on `flatten`/`flatten_entry` fields (requires `InflectableEntry` impl for `ForceFlag`)
- `InflectableEntry` blanket impl for `ForceFlag` (needed for the above)

I think these probably belong in this API, but Otel doesn't need them, so will keep this PR small.

### Testing

Integration tests covering: HighStorageResolution on individual field, NoMetric exclusion from metrics array, multiple flags on a single field, enum variant flags, a user-defined generic `FlagConstructor`, and backwards-compat verification that `ForceFlag<T, Ctor>` type wrappers still work. All tests validate actual EMF JSON output. Compile-fail UI tests for flags on flatten, flatten_entry, timestamp, and ignore fields.

### DIFF
--- a/metrique-macro/src/entry_impl.rs
+++ b/metrique-macro/src/entry_impl.rs
@@ -227,13 +227,27 @@ fn generate_field_writes(
                 let (extra, name) = make_inflect_metric_name(root_attrs, field);
                 let field_access = field_access(&field.ident);
                 let value = crate::value_impl::format_value(format, field_span, field_access);
+
+                let wrapped_value = if field.attrs.flags.is_empty() {
+                    quote! { #value }
+                } else {
+                    let flags = &field.attrs.flags;
+                    let mut expr = quote! { #value };
+                    for path in flags {
+                        expr = quote! {
+                            ::metrique::writer::value::ForceFlag::<_, #path>::from(#expr)
+                        };
+                    }
+                    quote! { &#expr }
+                };
+
                 quote_spanned! {field_span=>
                     ::metrique::writer::EntryWriter::value(#writer_ident,
                         {
                             #extra
                             ::metrique::concat::const_str_value::<#name>()
                         }
-                        , #value);
+                        , #wrapped_value);
                 }
             }
         };
--- a/metrique-macro/src/lib.rs
+++ b/metrique-macro/src/lib.rs
@@ -691,6 +691,29 @@ impl From<RawTag> for Tag {
     }
 }
 
+/// Wrapper for parsing `flags(Path1, Path2, ...)` as a single darling field.
+///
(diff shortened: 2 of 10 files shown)


rcoh (approved):
Awesome

PR #291, opened 2026-05-15
OTel integration

Fluzko:
## Summary

Adds a new `metrique-otel` crate that lets `metrique` entries record directly onto an OpenTelemetry `SdkMeterProvider`, with the OTLP/gRPC exporter wired up by default. Also adds the small `metrique-writer-core` change required to use flag-tagged values inside `metrique-aggregation`'s `Sum` strategy.

## What's new

**`metrique-otel::OtelSink`**
- Implements `EntrySink<E>` for any `E: Entry`.
- `OtelSink::with_otlp_default()` builds a sink wired to OTLP/gRPC using the standard `OTEL_*` environment variables (`OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_SERVICE_NAME`, ...).
- `OtelSinkBuilder` lets callers supply a custom `SdkMeterProvider`, a `Resource`, or an `InstrumentationScope` name.
- `flush_async()` drives `force_flush` on the meter provider on a blocking task and returns a `FlushWait` compatible with the existing `EntrySink` plumbing.

**Instrument kind via flags** (`metrique_otel::flags`)
- New tag types `Counter`, `UpDownCounter`, `Histogram`, `Gauge` applied with `#[metrics(flags(...))]` decide which OTel instrument kind a field records to.
- A private `OtelOptions` rides along inside `MetricFlags` so the sink can downcast to recover the kind at write time. Conflicting flags on the same value (e.g. `Counter` + `Gauge`) fail to merge, matching the existing `MetricOptions` contract.

**Using standard aggregation strategies**
- `metrique-otel` reuses `metrique-aggregation`'s existing strategies; there is no separate `metrique-otel::aggregate` module. Pair a strategy with the `flags(...)` tag that selects the OTel instrument kind:
  ```rust
  #[aggregate(strategy = Sum)]
  #[metrics(flags(Counter))]
  request_count: u64,

  #[aggregate(strategy = Sum)]
  #[metrics(flags(UpDownCounter))]
  in_flight: i64,

  #[aggregate(strategy = KeepLast)]
  #[metrics(flags(Gauge))]
  queue_depth: f64,

  #[aggregate(strategy = Histogram<Duration>)]
  latency: Duration,
  ```
- Histogram fields do not need a flag: the OTel translator routes any `Distribution`-shaped observation to a histogram instrument.
- Non-OTel sinks (EMF, log-line, ...) ignore `flags(...)`, so the same struct works in mixed topologies.

**Entry layout**
- String fields on an entry become OTel attributes attached to every metric in that entry. Per-metric `dimensions` take precedence over entry-level attributes on key collisions. Order of declaration does not matter because observations are buffered until `finish()`.
- Aggregator `#[aggregate(key)]` fields naturally end up as OTel attributes via the same mechanism.

**Instrument caching**
- `InstrumentCache` is a lock-free `papaya::HashMap` keyed by `(scope, name, instrument_kind)`. Steady-state lookups never take a lock; first-sight inserts use `get_or_insert_with` and are race-safe.
- Instrument units are fixed at creation, mirroring the OTel SDK's own behavior.

**Unit mapping**
- `metrique::Unit` is translated to the UCUM-flavored strings OTel expects (`ms`, `By`, `%`, `1` for dimensionless, etc.). Unknown future variants fall back to `1`.

**Examples**
- `otlp_aggregated`: recommended `KeyedAggregator -> WorkerSink -> OtelSink` topology.
- `otlp_alongside_emf`: same entries fanned out to both EMF and OTLP.
- `otlp_multi_entry`: multiple entry types sharing one sink.
- `otlp_resource_attributes`: custom OTel `Resource` via the builder.
- `otlp_views_and_temporality`: custom OTel views and delta/cumulative temporality.

## `metrique-writer-core` change

`ForceFlag<T, FLAGS>` now forwards `AddAssign` and `SubAssign` (both `Self` and `T` rhs) to its inner value. This is what lets a `ForceFlag<u64, Counter>` be summed by `metrique-aggregation`'s `Sum` strategy without unwrapping. No behavior change for existing callers.

## Design notes

- The sink does not implement `Drop`. Users can pass an externally-owned `SdkMeterProvider`, and shutting it down implicitly would be surprising. An explicit `OtelSink::shutdown` can be added later if a use case appears.
- Counters clamp negative observations at 0 rather than panicking; OTel counters are monotonic and producing an out-of-spec observation should not crash the producer.
- Histograms record the mean once when handed a pre-summed `Observation::Repeated` (count + total). Faithful distributions require keeping raw `Floating` observations and skipping the pre-sum.
- Entry timestamps are dropped: OTel readers stamp measurements with their own clock and there is no descriptor system yet for surfacing the entry timestamp as an attribute.
- `with_scope` callers' `String` is leaked once per sink build so the OTel SDK's `meter(&'static str)` API can borrow it. Cost is `O(#sinks_built)` bytes.


metrique-otel/examples/otlp_alongside_emf.rs
```diff
@@ -0,0 +1,144 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Same metric entries delivered to **both** `OtelSink` and a `metrique-writer-format-emf`
+//! pipeline. Useful while migrating from EMF to OTel, you can run them in
+//! parallel and diff the output before flipping over.
+//!
+//! ## Why this needs a small inline helper
```
jlizen: This is a bit unfortunate, not using an EntryIoStream is much worse for composability, sampling support, etc. it also is harder to change down the road than moving off of the otel libraries as a backend (if needed).

Is there a blocker for instead modeling it as a Format/EntryIoStream from the start and eg managing our own current thread runtime if we need something like that? Lacking context on what interfered.

metrique-otel/examples/otlp_aggregated.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Summed across all requests sharing the same `operation`; flushed as a
+    /// single `add()` on an OTel counter. `OtelCounter` bundles the `Sum`
+    /// aggregation with the `Counter` instrument-kind flag the sink reads.
+    #[aggregate(strategy = OtelCounter)]
+    request_count: u64,
+
+    /// Each `add_value` is preserved exactly; on flush, the merged
+    /// distribution is recorded on an OTel histogram. `OtelHistogram<Millisecond>`
+    /// bundles the histogram aggregation with the wire unit.
+    #[aggregate(strategy = OtelHistogram<Millisecond>)]
```
jlizen: Why are these special types needed compared to a regular histogram + tag decoration? This would be much more awkward for somebody that wants to generalize across formats, which is one of the big value adds for using metrique instead of plain otel.

Ideally the tag would be added to specify otel usage, but the regular type is still there for other usage.

metrique-otel/examples/otlp_views_and_temporality.rs
```diff
@@ -0,0 +1,166 @@
...
+/// visible overflow point in the collector. In production you'd set this
+/// to the largest expected legitimate cardinality.
+const MAX_REQUEST_COUNT_DATAPOINTS: usize = 3;
+
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    #[metrics(timestamp)]
+    timestamp: SystemTime,
+    operation: String,
+    /// High-cardinality attribute that drives up data-point count if
```
jlizen: It would be great to have commented in examples of output for these examples in json representation or something like that. We describe textually how these metric structs map to otel but examples always help

metrique-otel/src/flags.rs
```diff
@@ -0,0 +1,110 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Flag markers consumed by [`OtelSink`](crate::OtelSink).
+//!
+//! Apply via `#[metrics(flags(...))]` to declare the OTel instrument kind the
+//! sink should record observations against. At write time the sink downcasts
+//! [`MetricFlags`] to [`OtelOptions`] to pick the kind.
+//!
+//! ```ignore
```
jlizen: Is there a reason the various doc example are ignore instead of eg no_run? Concerned about drift from api

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,589 @@
...
+    /// Errors from `force_flush` are logged at `warn` level but not surfaced —
+    /// the [`EntrySink`] trait has no way to report them.
+    ///
+    /// Internally this uses `tokio::task::spawn_blocking` so it must be
+    /// awaited on a tokio runtime. Callers of [`with_otlp_default`] already
+    /// require tokio (the OTLP/gRPC exporters use it transitively), so this
+    /// is not an additional constraint in practice.
+    ///
+    /// [`with_otlp_default`]: Self::with_otlp_default
+    pub fn flush_async(&self) -> FlushWait {
```
jlizen, line 177: Needs a comment on panicking if not in a tokio runtime (should we have a sync option too?)

metrique-otel/src/flags.rs
```diff
@@ -0,0 +1,110 @@
...
+            InstrumentKind::UpDownCounter => &Self::UP_DOWN_COUNTER,
+            InstrumentKind::Histogram => &Self::HISTOGRAM,
+            InstrumentKind::Gauge => &Self::GAUGE,
+        }
+    }
+}
+
+impl MetricOptions for OtelOptions {
+    fn try_merge(&self, other: &dyn MetricOptions) -> Option<MetricFlags<'static>> {
+        // Only OTEL options merge with OTEL options, and only when they agree
```
jlizen: I think this might be stale since I see no `merge_assert_none` usage.  But, I assume we would only want to panic on debug builds and else have a rate limited warn or something on release, and have a rule like first wins? I do think that this will panic on `MetricFlags::try_merge` currently (not merge_assert_none).

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,589 @@
...
+impl OtelSink {
+    pub fn builder() -> OtelSinkBuilder {
+        OtelSinkBuilder::default()
+    }
+
+    /// Drive `force_flush` on the meter provider and resolve once it's done.
+    /// Errors from `force_flush` are logged at `warn` level but not surfaced —
+    /// the [`EntrySink`] trait has no way to report them.
+    ///
+    /// Internally this uses `tokio::task::spawn_blocking` so it must be
```
jlizen: I think the "not a constraint in practice" is noise, I would just phrase this as a comment block.

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,589 @@
...
+}
+
+impl OtelSink {
+    pub fn builder() -> OtelSinkBuilder {
+        OtelSinkBuilder::default()
+    }
+
+    /// Drive `force_flush` on the meter provider and resolve once it's done.
+    /// Errors from `force_flush` are logged at `warn` level but not surfaced —
+    /// the [`EntrySink`] trait has no way to report them.
```
jlizen: The no way to report them thing is noise in the doc comment

metrique-otel/src/aggregate.rs
```diff
@@ -0,0 +1,212 @@
...
+        self.last.map(|v| ForceFlag::from(WithUnit::from(v)))
+    }
+}
+
+// --- Histogram -------------------------------------------------------------
+
+/// Collect observations into a distribution; emit as an OTel histogram with
+/// unit `U`.
+///
+/// The underlying [`HistogramClosed`] already emits a `Distribution` flag,
```
jlizen: The second part of this doesn't seem like it belongs in customer docs

metrique-otel/src/aggregate.rs
```diff
@@ -0,0 +1,212 @@
+//! OTel-aware aggregation strategies that bundle merge behavior, unit, and
+//! OTel instrument kind into a single `#[aggregate(strategy = …)]` annotation.
+//!
+//! Each strategy here is a thin layer over a `metrique-aggregation` primitive
+//! plus a wrapping at close time:
+//!
+//! - [`OtelCounter<U>`] — `Sum`, closes as `ForceFlag<WithUnit<T, U>, flags::Counter>`.
```
jlizen: Any reason not to provide doc links to the forceflat and flags::counter types? (Same for others)

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,589 @@
...
+                    tracing::warn!(error = %e, "metrique-otel: meter provider force_flush failed");
+                }
+            })
+            .await;
+        })
+    }
+
+    /// Build a sink whose meter provider is wired to an OTLP/gRPC exporter
+    /// using the standard `OTEL_*` environment variables.
+    pub fn with_otlp_default() -> Result<Self, OtelSinkError> {
```
jlizen, line 226: This needs a panics comment if used outside of tokio context (AIUI)

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,589 @@
...
+    /// service component or per logical concern). A single sink always
+    /// records under one scope — looking through wrapper types like
+    /// `BoxEntry` or aggregator entries to recover the originating entry
+    /// type would require the (not-yet-merged) entry-descriptor system.
+    pub fn with_scope(mut self, name: impl Into<String>) -> Self {
+        self.scope = Some(name.into());
+        self
+    }
+
+    pub fn build(self) -> OtelSink {
```
jlizen, line 337: This need a panics comment if called outside tokio scope, if a custom sync friendly meter provider is not provided, correct?

jlizen (commented):
Some drive by comments. Please forgive repetition compared to the design, I haven't looked at the latest iterations. Defer to Russell's opinion on these as I might not be responsive to future cycles.

Fluzko:
Did some benchmarking by reusing `divan` as `metrique-aggregation` does as per our conversation with @rcoh. 
Will address @jlizen comments in a few, since those are pretty much wrong or stale docs and some structural changes, that (I think) are not related to this outcome.
Findings below:

### Bench Results
Based on this commit [dc0be03](https://github.com/awslabs/metrique/pull/291/commits/dc0be03013d5e0fb1cb79170d85de98094e66abd)
Both benches invoked with `--sample-count 500 --sample-size 1000` under the `bench` profile.

```
cargo bench -p metrique-otel --bench otel_append   -- --sample-count 500 --sample-size 1000
cargo bench -p metrique-otel --bench otel_vs_emf   -- --sample-count 500 --sample-size 1000
```

## otel_append
### Single-shot benches

| Bench | Time (fast / med / mean / slow) | Alloc | Throughput (med) |
|---|---|---|---|
| `append_single_counter` | 124.2 / 126.8 / 129.5 / 341.4 ns | 4 / 454 B | 7.88 Mitem/s |
| `append_wide_entry (1, 0)` | 127.8 / 130.1 / 133.6 / 238.5 ns | 4 / 456 B | 7.68 Mitem/s |
| `append_wide_entry (4, 2)` | 1.043 / 1.094 / 1.107 / 1.383 µs | 38 / 1.816 KB | 3.65 Mitem/s |
| `append_wide_entry (16, 8)` | 9.296 / 9.748 / 9.882 / 21.03 µs | 338 / 10.09 KB | 1.64 Mitem/s |
| `record_repeated_same_key` | 129.2 / 140.6 / 141.2 / 235.8 ns | 4 / 452 B | 7.11 Mitem/s |

(n, m) is (metrics, strings), the shape of the WideEntry for that run:
  - n counters (`ForceFlag<u64, Counter>` fields, named `Metric_0..Metric_{n-1}`)
  - m string dimension fields (named `Dim_0..Dim_{m-1}` with values `value_0..value_{m-1})`

## otel_vs_emf

Identical `MixedEntry` (1 Counter, 1 UpDownCounter, 1 Histogram, 1 Gauge, 2 string fields).

| Bench | fastest | median | mean | slowest | allocs (median) | bytes (median) | throughput (median) |
|---|---|---|---|---|---|---|---|
| `format_comparison_emf` | 287.5 ns | 300.7 ns | 309.0 ns | 491.5 ns | 0 | 0 B | 980.9 MB/s |
| `format_comparison_otel` | 1.112 µs | 1.145 µs | 1.167 µs | 1.411 µs | 38 | 1.857 KB | 872.6 Kitem/s |

## Summary 
- Floor cost is ~127 ns / 4 allocs per append, even for a 1-field entry. The cache-hit path (record_repeated_same_key) doesn't reduce alloc count, as seen, it's still 4/iter, same as 
- Per-field cost grows superlinearly: 1 >> 6 >> 24 fields takes 130 ns >> 1.09 µs >> 9.75 µs (~75x for 24x more fields), and allocs go 4 >> 38 >> 338 (~85x). Allocs-per-field rises from 4 to ~14.
- Contention is thread-count-bound, not instrument-bound: same-name vs disjoint-name differ by <2% at every thread count (943 vs 925 ns at t=8). Throughput per thread drops ~7-8x from t=1 to t=8 in both cases.
- OTel vs EMF on the same entry: OTel is ~3.8x slower (1.14 us vs 300 ns) and allocates 1.86 KB/append vs 0 B steady state for EMF.

I'll be taking a look into these to see where we can improve. Suggestions welcome

metrique-otel/examples/otlp_aggregated.rs
Fluzko: This approach comes from this old comment in a temporary PR [here](https://github.com/Fluzko/metrique/pull/3#discussion_r3244155133). Basically, it's about improving the api experience and make it as narrow as possible. On the other hand, you're right, it affects generalization. I feel that reusing our existing API is better, since it's _our way of doing it_. @rcoh what do you think?

rcoh: I'm fine w/ OTelHistogram personally. I was assuming that OtelHistogram is basically a shorthand for Histogram + OtelTag?

it makes it very explicit what is going on. It is _not_ our histogram that is being used, but actually the otel one presumably?

In any case, this is a two way door and I don't mind the explicitness for the moment.

metrique-otel/src/aggregate.rs
```diff
@@ -0,0 +1,212 @@
...
+
+// --- Histogram -------------------------------------------------------------
+
+/// Collect observations into a distribution; emit as an OTel histogram with
+/// unit `U`.
+///
+/// The underlying [`HistogramClosed`] already emits a `Distribution` flag,
+/// which the OTel sink maps to a histogram instrument — so no explicit flag
+/// wrap is needed here; we only override the wire unit.
+pub struct OtelHistogram<U = Dimensionless>(PhantomData<U>);
```
rcoh: I was imagining that we would actually create an Otel histogram backing this...this is fine too. 

I guess I figured we would use the aggregate strategy to initialize the otel primitives, then flush them.

This seems ok too (basically just amortizing the otel initialization to when flushing occurs) which creates a unified codepath

Fluzko:
Did small improvements, numbers are better now:
## `otel_append`

### Single-shot benches: Time (mean)
| Bench | Before | After | Delta |
|---|---|---|---|
| `append_single_counter`     | 129.5 ns | **77.93 ns** | -40% |
| `append_wide_entry (1, 0)`  | 133.6 ns | **82.63 ns** | -38% |
| `append_wide_entry (4, 2)`  | 1.107 µs | **575.2 ns** | -48% |
| `append_wide_entry (16, 8)` | 9.882 µs | **4.253 µs** | -57% |
| `record_repeated_same_key`  | 141.2 ns | **71.44 ns** | -49% |

### Single-shot benches: Allocs

| Bench | Before | After | Delta |
|---|---|---|---|
| `append_single_counter`     | 4 / 454 B      | **0 / 0 B**   | -100% |
| `append_wide_entry (1, 0)`  | 4 / 456 B      | **0 / 0 B**   | -100% |
| `append_wide_entry (4, 2)`  | 38 / 1.816 KB  | **4 / 24 B**  |  -89% |
| `append_wide_entry (16, 8)` | 338 / 10.09 KB | **16 / 96 B** |  -95% |
| `record_repeated_same_key`  | 4 / 452 B      | **0 / 0 B**   | -100% |

### Single-shot benches: Throughput (mean)

| Bench | Before | After | Delta |
|---|---|---|---|
| `append_single_counter`     |  7.72 Mitem/s | **12.83 Mitem/s** |  +66% |
| `append_wide_entry (1, 0)`  |  7.48 Mitem/s | **12.10 Mitem/s** |  +62% |
| `append_wide_entry (4, 2)`  |  3.61 Mitem/s |  **6.95 Mitem/s** |  +93% |
| `append_wide_entry (16, 8)` |  1.62 Mitem/s |  **3.76 Mitem/s** | +132% |
| `record_repeated_same_key`  |  7.08 Mitem/s | **13.99 Mitem/s** |  +98% |


### Contention sweep: per-thread (mean)

| Bench | t | Time before | Time after | Delta |
|---|---|---|---|---|
| `contention_disjoint_names` | 1 | 125.8 ns | **72.63 ns** | -42% |
|                             | 2 | 244.6 ns | **184.9 ns** | -24% |
|                             | 4 | 347.6 ns | **267.7 ns** | -23% |
|                             | 8 | 905.6 ns |   905.3 ns   |   0% |
| `contention_same_name`      | 1 | 144.9 ns | **69.73 ns** | -52% |
|                             | 2 | 227.3 ns | **165.2 ns** | -27% |
|                             | 4 | 363.9 ns | **261.3 ns** | -28% |
|                             | 8 | 930.1 ns | **884.8 ns** |  -5% |

## `otel_vs_emf`

| Bench | Time before | Time after | Delta time | Allocs before | Allocs after | Throughput before | Throughput after |
|---|---|---|---|---|---|---|---|
| `format_comparison_emf`  (control) | 309.0 ns | 305.7 ns    |  -1% | 0 / 0 B       | 0 / 0 B      | 954.5 MB/s     | 964.7 MB/s     |
| `format_comparison_otel`           | 1.167 µs | **553.6 ns** | -53% | 38 / 1.857 KB | **4 / 31 B** | 856.1 Kitem/s  | **1.806 Mitem/s** |


## Summary 
- **Cache-hit floor is alloc-free.** `record_repeated_same_key` 141.2 ns / 4 allocs >> 71.44 ns / 0 allocs (-49% time, -100% allocs). Same for `append_single_counter` and `append_wide_entry (1, 0)`.
- **Wide entries scale linearly in *string fields only*.** `append_wide_entry (16, 8)`: 9.882 µs / 338 allocs / 10.09 KB >> 4.253 µs / 16 allocs / 96 B (-57% time, -95% allocs). The 16 remaining allocs are `2 allocs × 8 string fields`, required by OTel's `KeyValue::new(key, value)` API.
- **OTel vs EMF gap closed from 3.8x to 1.81x.** `format_comparison_otel`: 1.167 µs / 38 / 1.857 KB >> 553.6 ns / 4 / 31 B (-53% time, -89% allocs). EMF unchanged at ~306 ns / 0 allocs.
- **Single-thread contention floor ~50% faster** (t=1 disjoint 125.8 >> 72.63 ns; t=1 same 144.9 >> 69.73 ns). At t=8 the gap shrinks because the bottleneck shifts to OTel SDK `Counter::add`.

metrique-otel/examples/otlp_aggregated.rs
jlizen: Knowing that it is indeed our histogram being used - [ref](https://github.com/awslabs/metrique/pull/291#discussion_r3275156285), does that change your feeling Russell? Meaning that non-otel formats would ignore the tag and treat this as a regular histogram with a weird name?

Don't have super strong feelings, I'm fine with the current if it is both of your preferences.

metrique-otel/src/aggregate.rs
Fluzko: I did prioritize the unified codepath. All four strategies (`OtelCounter`, `OtelUpDownCounter`, `OtelGauge`, `OtelHistogram`) follow the same shape: a plain accumulator type implements `AggregateValue<T>` for insert, and the `CloseValue` wraps the accumulated state into the OTel-flagged form at flush. This keeps OtelHistogram consistent, and with that pattern means the macro plumbing, lifetimes, and trait bounds are identical across strategies, which makes the code easier to extend. Also, we avoid SDK coupling in the hot path.

metrique-otel/examples/otlp_alongside_emf.rs
Fluzko: Agreed, the composability loss is real, and I don't think there's a hard blocker. Quick take:
- `Format` doesn't fit, because it writes bytes to `io::Write`, but the OTel SDK owns its own encode/batch/export pipeline. No byte surface to plug into unless we bypass the SDK or we start chopping down dependencies, and we go with our home-made solution.
- `EntryIoStream` does fit. `next(&mut self, &impl Entry)` is exactly what our translator already does, and flush maps to `SdkMeterProvider::force_flush`. Wrapping with `FlushImmediately` gives us all the handiness you mention.

What I was working around, none of it actually blocking:
1. We would need Tokio at build time since `OTLP/gRPC` + `PeriodicReader` spawn on the current runtime. Your suggestion (current-thread runtime on a background thread, `Runtime::enter` while building the provider) works. `with_otlp_http_default` already avoids it with a blocking client.
2. `SdkMeterProvider` already aggregates and batches, so `BackgroundQueue` would queue twice. `FlushImmediately` (just a `Mutex<Stream>`) seems like the correct wrapper, so actually... the SDK is the queue.
3. Thread-local scratch buffers don't carry cleanly once a `Mutex<Stream>` serializes calls. Keep them as
  thread-locals (correct, lower hit rate) or move into the stream.

So, as a summary, I would do `OtelStream: EntryIoStream` + a thin sink via `FlushImmediately`. Public API stays the same. Thoughts?

jlizen: Recommending `FlushImmediately` for an otel-only case, and then background queue for composed, seems fine. Anyway the extra cost of double queuing is not too bad, it pales in comparison to otel costs. The big thing to look out for is that you aren't causing per-entry tokio wakes if using FlushImmediately, but I assume not if you are using the aggregation API.

For tokio runtime management, I'd suggest also including an optional way of passing in an explicit runtime handle or pick up ambient tokio runtime context. Ideally you could also allow passing in an hyper-style userspace spawn implementation for non-tokio runtime support (or for customization of spawn behavior that is more granular than global runtime hooks) but I imagine you are limited by otel sdk there if it doesn't support similar that we could encapsulate?

Anyway I like the sound of the design you are discussing. It might anyway turn out that a special otel sink is the simplest "batteries included" API for only-otel usage. In which case it's probably fine to start with that if it is easiest or keeps our API more opaque for v0.

Fluzko: check [1128b0](https://github.com/awslabs/metrique/pull/291/commits/1128b001bd23f9f3fb015388e3b6d7060f8ac1d2) and let me know what you think. 
Also comes with 3x improvement with t8 at contention with different names, since we removed the path that was going through `AnyFlushImmediately` and was using the mutex, so this leaves Papaya totally unchained (🤣).

metrique-otel/examples/otlp_aggregated.rs
rcoh: yeah if its using our histogram its nicer to use tags. I don't need to block on it, I'd rather merge as is then iterate on this detail personally. I agree the ability to easily set tags on the value that is emitted would be ideal

Fluzko:
Made another bench, by just basically changing the sintax to use our aggregation crate:
```diff
- struct MyMetrics {
-     latency: ForceFlag<f64, Histogram>,
-     requests: ForceFlag<u64, Counter>,
-     operation: String,
- }
- sink.append(metrics);
+ #[aggregate]
+ #[metrics(rename_all = "PascalCase")]
+ struct MyMetrics {
+     #[aggregate(strategy = OtelHistogram)] latency: f64,
+     #[aggregate(strategy = OtelCounter)] requests: u64,
+     #[aggregate(key)] operation: String,
+ }
+ metrics.close_and_merge(worker.clone());
```
This also comes with an improvement IMO, which is that the cardinality becomes explicit at the struct definition (`#[aggregate(key)]`), the current path silently creates a new attribute set per distinct string combination.

Some numbers to see improvements:

| Bench | Median | Allocs/iter |
|---|---|---|
| `append_otel_direct` | 603 ns | 4 |
| `append_aggregation_single_key` | 21 ns | ~0 |

*~29× speedup per ingest, single-threaded. Aggregation drops the user-thread cost to roughly an mpsc send.*


| Bench | Median | Allocs/iter |
|---|---|---|
| `append_aggregation_keyed[1]` | 42 ns | ~0 |
| `append_aggregation_keyed[10]` | 24 ns | ~0 |
| `append_aggregation_keyed[100]` | 21 ns | ~0 |
| `append_aggregation_keyed[1000]` | 21 ns | ~0 |

*Cardinality is essentially free up to 1000 keys (~20-24 ns/iter, flat). HashMap lookup + merge does not scale visibly.*


| Bench | Median | Allocs/iter |
|---|---|---|
| `contention_aggregation_single_key[t=1]` | 46 ns | 2 |
| `contention_aggregation_single_key[t=2]` | 166 ns | 2 |
| `contention_aggregation_single_key[t=4]` | 496 ns | 2 |
| `contention_aggregation_single_key[t=8]` | 536 ns | 2.5 |
| `contention_aggregation_disjoint_keys[t=1]` | 58 ns | 2 |
| `contention_aggregation_disjoint_keys[t=2]` | 118 ns | 2 |
| `contention_aggregation_disjoint_keys[t=4]` | 3.4 µs | 2 |
| `contention_aggregation_disjoint_keys[t=8]` | 672 ns | 2.25 |

*All producer threads share one channel into a single worker thread. As more threads ingest, they queue up against that one channel, and the per-send cost rises. Since the keys they use don't matter, everyone is stuck behind the same pipe, so "disjoint keys" don't help. By 8 threads, each ingest costs about as much as a single-threaded direct OtelSink call: the win has been spent on contention. The fix, when this matters, is to stop sharing a single WorkerSink.*

Fluzko:
Benching the worker pace at [2d834b](https://github.com/awslabs/metrique/pull/291/commits/2d834b271104c79ad5a61926519322071c59752d), this is the summary:

| Workload | Burst | Drain |
|---|---|---|
| 1 thread, 10k | 10k | 0.4 ms |
| 1 thread, 100k | 100k | 3.9 ms |
| 4 threads, 10k each | 40k | 0.8 ms |
| 4 threads, 100k each | 400k | 7.7 ms |
| 8 threads, 10k each | 80k | 0.8 ms |
| 8 threads, 100k each | 800k | 1.0 ms |

### Summary

With 1 producer & big burst: drain scales 10x with 10x more entries. That's the one case where the producer outruns the worker. The worker's true ceiling shows at **~25 million merges/sec**.

Many producers: drain stays small even at 800k entries. The `mpsc` channel itself bottlenecks the producers, so they never push faster than the worker can drain. The queue stays nearly empty.

Seems that only way to overwhelm the `worker` is one producer pushing >25M entries/sec. Pretty sure real workloads are nowhere near that. Add more producers and they slow each other down at the channel before the worker ever falls behind.

metrique-otel/examples/otlp_aggregated.rs
Fluzko: implemented at [cd231bcc](https://github.com/awslabs/metrique/pull/291/commits/cd231bcc083bf4c19829d3247b26aef97bd31402)

metrique-otel/src/metrics.rs
```diff
@@ -0,0 +1,241 @@
...
+                        Observation::Unsigned(v) => v as f64,
+                        Observation::Floating(v) => v,
+                        // Repeated has already collapsed the distribution to
+                        // (total, occurrences); we can't recover individual
+                        // samples. Record the mean once — bucketing is lossy
+                        // but count and sum stay sensible. Users that need
+                        // faithful distributions should keep raw `Floating`
+                        // observations and avoid pre-summing.
+                        Observation::Repeated { total, occurrences } if occurrences > 0 => {
+                            total / occurrences as f64
```
rcoh: we need to document this since it means that the count field is going to be wrong

metrique-otel/src/metrics.rs
```diff
@@ -0,0 +1,241 @@
...
+                    };
+                    c.add(v, attributes);
+                }
+            }
+            CachedInstrument::UpDownCounter(c) => {
+                for obs in observations {
+                    let v = match obs {
+                        Observation::Unsigned(v) => v as i64,
+                        Observation::Floating(v) => v as i64,
+                        Observation::Repeated { total, .. } => total as i64,
```
rcoh: this should use the `mean`, not the `total`

metrique-otel/src/translator.rs
```diff
@@ -0,0 +1,254 @@
...
+            } else {
+                // Uncommon path: combine per-metric dims (first, so they
+                // take precedence on the wire) with the entry-level
+                // attrs. Reuses a single scratch Vec across metrics. The
+                // OTEL SDK does not de-dup attribute keys, so any
+                // collision is left visible — that's a user-data problem,
+                // not something to paper over here.
+                bufs.scratch.clear();
+                bufs.scratch.extend(m.per_metric_dimensions.iter().cloned());
+                bufs.scratch.extend(bufs.entry_attributes.iter().cloned());
```
rcoh: is `cloned` actually required here? we clear them just below so can we drain them instead?

metrique-otel/src/translator.rs
```diff
@@ -0,0 +1,254 @@
...
+            cache.record(
+                scope,
+                &m.name,
+                m.kind,
+                m.observations.iter().copied(),
+                m.unit,
+                attrs,
+            );
+            m.name.clear();
+            m.observations.clear();
```
rcoh: again, copy followed by clear, it seems like we could avoid this potentially. possilby just worth a todo

metrique-otel/src/translator.rs
```diff
@@ -0,0 +1,254 @@
...
+            m.per_metric_dimensions.clear();
+            bufs.metric_pool.push(m);
+        }
+    }
+}
+
+impl<'a, 'b> EntryWriter<'a> for OtelEntryWriter<'b> {
+    fn timestamp(&mut self, _timestamp: std::time::SystemTime) {
```
rcoh: worth documenting if not already done

metrique-otel/src/metrics.rs
```diff
@@ -0,0 +1,241 @@
...
+
+        match &instrument {
+            CachedInstrument::Counter(c) => {
+                for obs in observations {
+                    let v = match obs {
+                        Observation::Unsigned(v) => v,
+                        // Counters are non-negative; clamp at 0 rather than
+                        // emitting a panic for an out-of-spec observation.
+                        Observation::Floating(v) => v.max(0.0) as u64,
+                        Observation::Repeated { total, .. } => total.max(0.0) as u64,
```
rcoh: I think emitting a rate limited warning rather than pushing bad data is probably more useful. this would also happen for NaN.

rcoh (approved):
Great work! Excited for folks to try this. left a few small notes

metrique-otel/src/metrics.rs
```diff
@@ -0,0 +1,485 @@
...
+                                continue;
+                            }
+                            v
+                        }
+                        // Repeated has already collapsed the distribution to
+                        // (total, occurrences); we can't recover individual
+                        // samples. Record the mean once: bucketing is lossy
+                        // and, critically, the OTel histogram's `count` will
+                        // report 1 here rather than `occurrences`, so any
+                        // count-based aggregation downstream will undercount.
```
jlizen: Is there a reason not to call `histogram.record(mean)` in a loop `n` times instead to preserve the proper count? We could add a cap to avoid an explosively large loop.

metrique-otel/src/metrics.rs
```diff
@@ -0,0 +1,485 @@
...
+                        // The `sum` (mean x 1) is also no longer the true sum
+                        // (which would be `total`). Users that need faithful
```
jlizen: This belongs in user facing docs (the default behavior, what to do for faithful counts with raw `Floating`, etc)

metrique-otel/src/translator.rs
```diff
@@ -0,0 +1,251 @@
...
+                attrs,
+            );
+            m.name.clear();
+            bufs.metric_pool.push(m);
+        }
+    }
+}
+
+impl<'a, 'b> EntryWriter<'a> for OtelEntryWriter<'b> {
+    /// Dropped. OTel meter readers stamp measurements with their own clock,
```
jlizen: This should go in user-facing module docs or something too, similar to the repeated stuff

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,827 @@
...
+    ///
+    /// The returned [`FlushWait`] must be awaited on a tokio runtime; it uses
+    /// `tokio::task::spawn_blocking` internally and will panic if polled
+    /// outside of one.
+    pub fn flush_async(&self) -> FlushWait {
+        let meter = self.meter_provider.clone();
+        FlushWait::from_future(async move {
+            let _ = tokio::task::spawn_blocking(move || {
+                if let Err(e) = meter.force_flush() {
+                    tracing::warn!(error = %e, "metrique-otel: meter provider force_flush failed");
```
jlizen: should this be rate-limited?

metrique-otel/src/lib.rs
```diff
@@ -0,0 +1,827 @@
...
+    /// flush completes. Errors are logged at `warn` level.
+    ///
+    /// `SdkMeterProvider::force_flush` is itself synchronous; `flush_async`
+    /// wraps it in `spawn_blocking` only to remain well-behaved when called
+    /// from inside a tokio executor. From a non-tokio context, prefer this
+    /// method; from an async context, prefer [`Self::flush_async`] so the
+    /// tokio runtime doesn't get blocked.
+    pub fn flush(&self) {
+        if let Err(e) = self.meter_provider.force_flush() {
+            tracing::warn!(error = %e, "metrique-otel: meter provider force_flush failed");
```
jlizen: should this be rate-limited?

metrique-otel/src/flags.rs
```diff
@@ -0,0 +1,246 @@
...
+/// Last time a conflict warning was emitted, used to rate-limit the
+/// `tracing::warn!` in [`warn_conflict`] to at most once per
+/// [`CONFLICT_WARN_INTERVAL`] process-wide.
+static LAST_CONFLICT_WARN: OnceLock<Mutex<Instant>> = OnceLock::new();
+
+fn warn_conflict(kept: InstrumentKind, dropped: InstrumentKind) {
+    let mu = LAST_CONFLICT_WARN.get_or_init(|| Mutex::new(Instant::now() - CONFLICT_WARN_INTERVAL));
+    let mut last = mu.lock().expect("OTel conflict warn mutex poisoned");
+    if last.elapsed() >= CONFLICT_WARN_INTERVAL {
+        tracing::warn!(
```
jlizen, line 41: This seems like it might be reimplementing the `rate_limited!`  machinery we already have elsewhere?

metrique-otel/src/metrics.rs
```diff
@@ -0,0 +1,485 @@
...
+const BAD_VALUE_WARN_INTERVAL: Duration = Duration::from_secs(60);
+
+static LAST_BAD_VALUE_WARN: OnceLock<Mutex<Instant>> = OnceLock::new();
+
+#[cold]
+fn warn_bad_value(name: &str, kind: InstrumentKind, value: f64, reason: &'static str) {
+    let mu =
+        LAST_BAD_VALUE_WARN.get_or_init(|| Mutex::new(Instant::now() - BAD_VALUE_WARN_INTERVAL));
+    let mut last = mu.lock().expect("OTel bad-value warn mutex poisoned");
+    if last.elapsed() >= BAD_VALUE_WARN_INTERVAL {
```
jlizen: This seems like it might be reimplementing the `rate_limited!`  machinery we already have elsewhere?

jlizen (approved):
Looks good, approving with some small comments.

I'd be ok with it landing without them but might be worth considering a follow-up issue if you agree with them.

metrique-otel/src/flags.rs
Fluzko: Oh, didn't know that we had that helper, thanks!

metrique-otel/examples/otlp_alongside_emf.rs
```diff
@@ -0,0 +1,146 @@
...
+    let otel_sink = OtelSink::builder()
+        .with_meter_provider(meter_provider)
+        .with_scope("metrique/otlp_alongside_emf")
+        .build();
+    let otel_worker = WorkerSink::new(
+        KeyedAggregator::<OtelRequest, _>::new(otel_sink.clone()),
+        Duration::from_secs(1),
+    );
+
+    // --- EMF pipeline (direct) ---
```
jlizen, line 113: Is there a reason not to use `AppendOnDrop` in this example?

PR #293, opened 2026-05-19
Support folding RuntimeMetrics (tokio-metrics) into other structs 

### LINKED ISSUE #269: feat: Support folding RuntimeMetrics (tokio-metrics) into other structs
In https://github.com/awslabs/metrique/pull/256 we added support for a tokio-metrics's RuntimeMetrics subscription, which starts a reporter task to poll for the runtime metrics and emit them as a separate `RuntimeMetrics`.

Perhaps we can implement a new `EmbeddedTokioMetrics` struct, that can be embedded into a metrics struct directly, e.g.:
```rust
#[metrics]
struct MyMetrics {
    tokio_metrics: EmbeddedTokioMetrics,
    operation: Operation,
    // ...
}
```
`EmbeddedTokioMetrics` could then have a default impl. that does the interval polling. 
See the original comment: https://github.com/awslabs/metrique/pull/256#discussion_r3065042660

Mentioned on https://github.com/awslabs/metrique/issues/278 by @jlizen:

> I'd like to directly fold those metrics into my existing entries. Technically I can piece it together, but it takes annoying extra querying.
> 
> What I would like is, instead store a "snapshot" that updates on an interval, and then each entry automatically folds it in when I use some special macro annotation.
> 
> We could do it with the [metrique_util::State](https://docs.rs/metrique-util/latest/metrique_util/struct.State.html) API, probably, and some macro magic.

> It seems like that this would be a good fit for wrapping in the task metrics as well, which we don't currently include.

fmzbl:
closes #269 

Adds `EmbeddedTokioMetrics::spawn` returning `State<EmbeddedTokioMetrics>` to fold the latest Tokio runtime sample into per-request entries via `#[metrics(flatten)`


metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -1,13 +1,92 @@
...
+        <RtClosed as InflectableEntry<NS>>::sample_group(self.0.as_ref())
+    }
+}
+
+impl EmbeddedTokioMetrics {
+    /// Spawn the runtime-metrics sampler and return a [`State`] that
+    /// resolves to the most recent sample each time an entry closes.
+    ///
+    /// The sampler runs for the lifetime of the Tokio runtime.
+    pub fn spawn(config: TokioRuntimeMetricsConfig) -> State<Self> {
```
jv1i: Should this really live for the whole Tokio runtime? 
Could we tie it to the sink's AttachHandle drop, same as `subscribe_tokio_runtime_metrics` does? (calling `register_shutdown_fn` with the sampler's abort)

nit: I'd name this method something like `start`/`start_sampling` instead.

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -1,13 +1,92 @@
...
+        let initial = Self(Arc::new(RuntimeMetrics::default().close()));
+        let state = State::new(initial);
```
jv1i: Could we seed `initial` from `monitor.intervals().next()` instead of Default? Avoids emitting all-zeros on cold start and lets the example drop its sleep(500ms) workaround.

metrique-util/src/tokio_metrics_reporter.rs
jv1i: Thinking of it, perhaps you can even abstract some helpers to share between `subscribe_tokio_runtime_metrics` and `EmbeddedTokioMetrics`, like spawning, building the monitor, etc. (if you see it fit)

yulnr (commented):
Nice direction! leaving some notes from a quick first pass

jv1i:
We should make this as discoverable (if not more) than `subscribe_tokio_runtime_metrics` so worth a mention on the same section in `/sinks/docs`, and perhaps a cross-link on `subscribe_tokio_runtime_metrics` (on the trait docs iirc) mentioning this option as an alternative if users want to embed into other structs.

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -1,13 +1,92 @@
...
+        self
+    }
+}
+
+impl<NS: NameStyle> InflectableEntry<NS> for EmbeddedTokioMetrics
+where
+    RtClosed: InflectableEntry<NS>,
+{
+    fn write<'a>(&'a self, w: &mut impl EntryWriter<'a>) {
+        <RtClosed as InflectableEntry<NS>>::write(self.0.as_ref(), w);
```
rcoh: I think this is equivalent to `#[metrics(flatten)]`

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -99,40 +148,83 @@ pub trait AttachGlobalEntrySinkTokioMetricsExt: AttachGlobalEntrySink + 'static
...
-        let handle = Handle::current();
-        let monitor = RuntimeMonitor::new(&handle);
-        for snapshot in monitor.intervals() {
-            sink.append(DynamicInflectionEntry {
-                entry: snapshot.close(),
-                name_style,
-            });
+        for snapshot in intervals {
             tokio::time::sleep(interval).await;
+            on_sample(snapshot);
```
jv1i, line 280: slight regression: I think here we want to append, then sleep, as it was doing before instead of sleep and then append (swap `tokio::time:sleep()` and `on_sample()`). We want the first sample to be pulled after sleeping so that it actually has a meaningful delta from the initial sample pulled at the top of the function.

yulnr (approved):
LGTM overall. Bit more explicit than what I had in mind (some type, or macro support, that polls without an explicit `embed_tokio...()` call), but explicit isn't necessarily  worse :) and it lines up well with @jlizen's `State` idea. @jlizen is this what you had in mind?

An improvement for a more-automatic version, similar to your first iteration @fmzbl (fcb1be2): perhaps we could use a bare-field/zst type (e.g. `EmbeddedTokioMetrics`, without users explicitly writing `State<..>`) whose `close()` folds in the latest sample (background sampler still polls on an interval, the type just reads the current snapshot), so using the type wires it up, no `embed_...()` call. Could still use `State` under the hood.

What do we think, worth layering that on, or is this API good as-is?
(or alternatively some macro support like `#[metrics(embed_tokio_metrics)]`?)

metrique-util/examples/tokio-metrics-folded.rs
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+    for op in ["Read", "Write", "Read"] {
+        let _m = RequestMetrics {
+            operation: op,
+            success: true,
+            runtime: runtime.clone(),
+        }
+        .append_on_drop(ServiceMetrics::sink());
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
rcoh, line 54: an example of the metrics this produces a a comment would be helpful

metrique-util/examples/tokio-metrics-folded.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let _attach_handle = ServiceMetrics::attach_to_stream(
+        Emf::all_validations("FoldedTokioMetricsExample".to_string(), vec![vec![]])
+            .output_to(std::io::stderr()),
+    );
+
+    let runtime = ServiceMetrics::embed_tokio_runtime_metrics(
```
rcoh, line 38: is this safe to call twice?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -1,13 +1,53 @@
...
+///
+/// Cloning the [`State`] (per request) and closing the entry are both
+/// cheap reference-count operations.
+#[derive(Clone)]
+pub struct EmbeddedTokioMetrics(Arc<RtClosed>);
+
+impl CloseValue for EmbeddedTokioMetrics {
+    type Closed = Self;
+    fn close(self) -> Self {
+        self
```
rcoh: if this was `self.0` and it closed to `RtClosed` I don't think you would need the imp below

rcoh: I think it could also be `impl CloseValue for &'_ EmbeddedTokioMetrics` which makes it useable in more places

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -1,13 +1,53 @@
...
+/// Obtain a `State<EmbeddedTokioMetrics>` by calling
+/// [`AttachGlobalEntrySinkTokioMetricsExt::embed_tokio_runtime_metrics`] on
+/// your global entry sink. The sampler is aborted when the sink's
+/// [`AttachHandle`](metrique::writer::sink::AttachHandle) is dropped.
+/// Embed the [`State`] in your entry with `#[metrics(flatten)]`.
+///
+/// Cloning the [`State`] (per request) and closing the entry are both
+/// cheap reference-count operations.
+#[derive(Clone)]
+pub struct EmbeddedTokioMetrics(Arc<RtClosed>);
```
rcoh: I'm not sure about the `Embedded` prefix. What do you think about `TokioRuntimeMetrics`?

metrique-util/src/tokio_metrics_reporter.rs
fmzbl: i did some digging and from what i understand doing this would add on every entry close clone the size of the inner struct i would have to add, right now there are only atomic refcount bumps with the manual impl

metrique-util/examples/tokio-metrics-folded.rs
fmzbl: i believe so, u would have 2 separate parallel samplers with their own state and such

metrique-util/src/tokio_metrics_reporter.rs
fmzbl: hm, it may be confused with the `RuntimeMetrics` struct, what about `TokioRuntimeSnapshot` maybe?

metrique-util/examples/tokio-metrics-folded.rs
```diff
@@ -0,0 +1,58 @@
...
+    // Each loop iteration emits one EMF record like:
+    //   {"Operation":"Read", "Success":1,
+    //    "WorkersCount":12, "TotalParkCount":4, "TotalBusyDuration":0.135,
+    //    "GlobalQueueDepth":0, "Elapsed":500.7, ...}
+    // — user fields and folded runtime fields on the same line.
+    for op in ["Read", "Write", "Read"] {
+        let _m = RequestMetrics {
+            operation: op,
+            success: true,
+            runtime: runtime.clone(),
```
jlizen, line 51: Would we want to demonstrate how to make this accessible via global static? That seems like a more likely use case

metrique-util/examples/tokio-metrics-folded.rs
```diff
@@ -0,0 +1,58 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jlizen, line 1: Any reason not to wrap this into the extras example?

fmzbl: it would get a bit messy in my opinion leaving everything on extras. also now i added an static example about previous comment. but i can do it if you think is better

metrique/examples/tokio-metrics-folded.rs
```diff
@@ -0,0 +1,58 @@
...
+};
+use metrique_util::{
+    AttachGlobalEntrySinkTokioMetricsExt, State, TokioRuntimeMetricsConfig, TokioRuntimeSnapshot,
+};
+
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    operation: &'static str,
+    success: bool,
+    #[metrics(flatten)]
```
rcoh: can you add `prefix = "tokio."`? (and verify everything works). I think that would be an improvement and avoid any possible name collisions

rcoh (approved):
lgtm, one comment re: prefixing

metrique/examples/tokio-metrics-folded.rs
fmzbl: it doesnt work, but i could do `tokio_` maybe instead

jlizen: You can use `prefix_exact` but it might look weird since it won't respect inflection

jlizen: My suggestion would be to instead have a doc example showing flatten + prefixing

The names are specific enough that I don't think unconditional prefixing is warranted necessarily...

jlizen: (you do have it in the example already, but i'd also move it into somewhere visible in docs.rs)

jlizen (approved):

PR #295, opened 2026-05-27
feat(aggregation): support WithDimensions in #[aggregate]

### LINKED ISSUE #286: feat: Support WithDimensions<V, N> in #[aggregate]
There is no way to add dynamic Dimensions using WithDimension with Aggregated Metrics struct. To achieve this we would need something like
```
#[aggregate]
pub struct MyMetrics {
    #[aggregate(strategy = Sum)]
    pub get_count: u64,
    #[aggregate(strategy = Sum)]
    pub post_count: u64,
    #[aggregate(strategy = Sum)]
    pub put_count: u64,
}
```
Which requires 1 field per possible dimension value, and does not support dynamic/runtime dimension values.

Request:
given an aggregate metric struct
```
#[aggregate]
pub struct MyMetrics {
    #[aggregate(strategy = Sum)]
    pub count: WithDimension<u64>,
}
```
and in my code I run
```
let mut gets_count = MyMetrics {
   count: 1u64.with_dimension("event", "GET")
}; // x30 times
let mut posts_count = MyMetrics {
   count: 1u64.with_dimension("event", "POST")
}; // x50 times
```
It will produce 2 separate metric entries
`count=30, event="GET"` and `count=50, event="POST"`

rcoh:
Closes #286.

## Summary

A field of type `WithDimension<V>` / `WithDimensions<V, N>` / `WithVecDimensions<V>` inside an `#[aggregate]` struct now contributes its dimensions to the merge key. Aggregating multiple entries produces one merged row per distinct dimension tuple, with the dimensions hoisted to entry-level (emitted as both ordinary fields and an `EntryDimensions` config so EMF treats them as proper dimensions).

```rust
#[aggregate]
struct MyMetrics {
    #[aggregate(strategy = Sum)]
    count: WithDimension<u64>,
}

// 30 inserts of count=1 with event=GET, 50 with event=POST
// -> 2 entries: { Count: 30, Event: "GET" } and { Count: 50, Event: "POST" }
```

## Approach

**Structural collection, not a runtime write-probe.** The earlier revision of this PR discovered an entry's `(class, instance)` dimension pairs by running a full no-op `Entry::write` traversal over the closed entry — a "double write": once through a probe writer to harvest dimensions, then again to actually emit the merged row.

Since then, entry descriptors (#289/#331) landed on `main` and formalize each entry's field structure at compile time. The `#[aggregate]` macro already has that structural knowledge, so this PR now merges `main` and replaces the runtime probe with **targeted, recursive extraction** that reads only the fields that can carry dimensions:

- A new `CollectDimensions` trait (`metrique_aggregation::dimensions`) with a blanket impl for `WithDimensions<V, N>` that pushes its own `.dimensions()`.
- `#[aggregate]` generates `impl CollectDimensions` for every aggregate struct: dimensional fields contribute their pairs; `Flatten`-strategy fields recurse into the nested aggregate struct's own generated impl (so flattened/nested dimensions are picked up structurally rather than by traversal); everything else is skipped. Method-call syntax auto-derefs through any `WithUnit` wrapper on the closed field.
- `extract_dimensions` keeps its name and signature but is now backed by the trait (collect → sort → dedup), with **no** `Entry` / `Value` / `EntryWriter` / `ValueWriter` machinery on the key-extraction path.

The old `dimension_probe.rs` is deleted; the module is renamed to `dimensions.rs`.

**Macro detection mirrors `WithUnit`.** The aggregate macro detects `WithDimensions` field types by trailing path segment and generates `into_value()` / `inner()` calls to strip the wrapper before passing to `AggregateValue::insert`. It additionally detects `Flatten`-strategy fields (also by trailing path segment) to drive `CollectDimensions` recursion.

**Conditional codegen.** The dimension-*key* machinery (the extended `Key` struct, `__dyn_keys`, `EntryDimensions`) is still emitted only when at least one `WithDimensions` field is present and `dimensions_as_key` is on. The `CollectDimensions` impl is emitted for every aggregate struct (empty no-op body when nothing can contribute) so that `Flatten` fields can always recurse into their inner type.

## What's new

- `WithDimensions::into_value(self) -> V` — consume the wrapper.
- `WithDimensions::inner(&self) -> &V` — borrow the inner value.
- `WithDimensionsExt` trait — used by macro-generated code.
- `metrique_aggregation::dimensions` module: `CollectDimensions` trait, `DimensionSet` alias, and `extract_dimensions(entry)`.
- Container attribute `#[aggregate(dimensions_as_key = false)]` opts out (default `true`).

## Compatibility

Purely additive. The obvious shape (`#[aggregate(strategy = Sum)] count: WithDimension<u64>`) did not compile before this change, since no `AggregateValue<WithDimension<u64>>` impl existed. No existing code is affected; observable behavior (buckets, `EntryDimensions` emission) is identical to the probe-based revision.

`Aggregate<T>` (embedded single-bucket aggregator) refuses to compile for structs with non-empty dimension keys, via the existing `Key = NoKey` constraint. A UI fail test confirms a clean error message.

## Tests

- Direct unit tests of `CollectDimensions` / `extract_dimensions`: empty entry, single dimension, multiple classes, dedup, same-class-different-instance, sorting/canonicalization, runtime-set dimensions via `add_dimension`, nested via flatten.
- Integration tests covering: the issue's example verbatim, mixed static + dynamic key, multiple `WithDimension` fields with different classes, dedup of identical pairs, empty dimensions (single bucket, no `EntryDimensions` emitted), `Histogram<Duration, SortAndMerge>` strategy on `WithDimension<Duration>`, flatten with nested dimensions, and a `WithDimension` field with an explicit `unit` (exercising the `WithUnit<WithDimensions<..>>` closed-type deref path).
- Macro snapshots for the generated `CollectDimensions` impls (dimensional and empty cases).
- UI fail test for `Aggregate<T>` embedding a dimensional struct.
- Full workspace suite passes (671 tests). `cargo +nightly fmt --check` and `cargo clippy --workspace --all-features -- -D warnings` are clean.

## Out of scope

- Field-level "exclude from dimension key" attribute.
- Embeddable keyed aggregator (multi-bucket analogue of `Aggregate<T>`).


jlizen:
Interested to see how this evolves based on #289

PR #296, opened 2026-05-29
Add observer hooks for BackgroundQueue and FlushImmediately sinks

### LINKED ISSUE #205: Improve reporting of the BackgroundQueueBuilder metrics

Fluzko:
closes #205 

## Summary
Sink diagnostics were hardwired to the private `MetricRecorder` trait, which in practice meant they could only go to a `metrics.rs` backend. The only public entry point (`metric_recorder`) was deprecated and uncallable because `MetricRecorder` impls are private, so there was no supported way to observe queue health from any other backend.

The observer trait makes these events first-class and backend-agnostic, so `metrics.rs` becomes just one adapter, and callers using a different observability stack (or no global recorder at all) can capture the same signals directly.

The PR introduces an observer abstraction for the diagnostics that `BackgroundQueue` and `FlushImmediately` emit about themselves (queue overflows, per-flush emitted/error counts, idle/length samples, flush timing).

- New `BackgroundQueueObserver` / `FlushImmediatelyObserver` traits, with `#[non_exhaustive]` event enums (BackgroundQueueEvent) and a blanket impl so any closure of the right shape is an observer.
- New observer(..) builder method on both sinks.
- The existing `metrics_recorder_global` / `metrics_recorder_local` builders are unchanged for callers, so basically they now route through a small internal adapter `MetricsRsObserver` over the same observer trait.
- Removes the previously-deprecated, user-uncallable `metric_recorder(Option<Box<dyn MetricRecorder>>)` methods.


metrique-writer/src/sink/observer.rs
```diff
@@ -0,0 +1,103 @@
...
+/// name as `sink` and the flush duration in milliseconds.
+///
+/// Any `Fn(&str, u32)` is an observer via the blanket impl, so a closure works
+/// wherever a `FlushImmediatelyObserver` is expected.
+///
+/// [`FlushImmediately`]: super::FlushImmediately
+/// [`FlushImmediatelyBuilder::observer`]: super::FlushImmediatelyBuilder::observer
+pub trait FlushImmediatelyObserver: Send + Sync {
+    /// A flush has completed in `flush_time_ms` milliseconds.
+    fn on_flush(&self, sink: &str, flush_time_ms: u32);
```
rcoh: does flush immediately ever have io errors or anything else worth recording?

rcoh: can the flush time be a duration here? ms is probably not the right unit anyway.

metrique-writer/src/sink/observer.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+impl<F: Fn(&str, u32) + Send + Sync> FlushImmediatelyObserver for F {
```
rcoh: I would get rid of this, it is not scalable to ever having more than one callback

rcoh (commented):
Looks good! some small comments

metrique-writer/src/sink/observer.rs
Fluzko: > does flush immediately ever have io errors or anything else worth recording?

Yes, it slipped my mind. I've replicated the approach at BackgroundQueue.


> can the flush time be a duration here? ms is probably not the right unit anyway.

Switched it to micro seconds and also added `Duration`

metrique-writer/src/sink/observer.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Delivered to a [`BackgroundQueueObserver`] alongside the queue's configured
+/// metric name. The queue constructs these events; an observer only ever matches
+/// on them.
+///
+/// [`BackgroundQueue`]: super::BackgroundQueue
+#[cfg(feature = "background-queue")]
+#[non_exhaustive]
+pub enum BackgroundQueueEvent {
+    /// An entry was dropped because the queue was full.
+    QueueOverflow,
```
rcoh, line 26: add non_exhaustive on this as well. its actually pretty plausible we would add more fields here

rcoh (approved):
lgtm, just one missing non_exhaustive

Issue #298, opened 2026-06-01
feat: Support folding Sysinfo metrics into other structs


fmzbl:
#269 added support for folding RuntimeMetrics (tokio-metrics) into other structs.
implement the same idea for the sysinfo metrics.

jlizen:
Closed by #301

PR #301, opened 2026-06-04
feat: Support folding sysinfo metrics into other structs

fmzbl:
Addresses #298
 
Adds `AttachGlobalEntrySinkSysinfoExt::embed_sysinfo_metrics` returning `State` to fold the latest sysinfo sample into per-request entries via `#[metrics(flatten)]`


metrique-util/src/sysinfo_reporter.rs
```diff
@@ -118,6 +120,7 @@ impl SysinfoMetricsConfig {
 ///
 /// [sysinfo docs]: https://docs.rs/sysinfo
 #[metrics(subfield_owned)]
+#[derive(Default)]
```
rcoh: I'm not sure default is semantically accurate here. e.g. 0cpu is not a sensible default. 

I'd avoid adding this derive if possible

metrique-util/src/sysinfo_reporter.rs
```diff
@@ -839,4 +922,63 @@ mod tests {
...
+
+        check!(entry.values["Operation"] == "Read");
+        check!(entry.metrics["TotalMemory"] > 0);
+        check!(entry.metrics["Uptime"] > 0);
+    }
+
+    #[tokio::test]
+    async fn embed_aborted_on_handle_drop() {
+        metrique_writer::sink::global_entry_sink! { Sink }
+        let TestEntrySink { sink, .. } = test_entry_sink();
```
rcoh, line 1019: Does this (and other tests) actually need to sleep?

It's best to avoid this during tests because it is inherently racy. Can tokio::time::advance be used instead (combined with start_paused)

metrique-util/src/sysinfo_reporter.rs
rcoh: can you spawn this thread with a name while we're here?

metrique-util/examples/sysinfo-folded.rs
```diff
@@ -0,0 +1,62 @@
...
+            .output_to(std::io::stderr()),
+    );
+
+    let system = ServiceMetrics::embed_sysinfo_metrics(
+        SysinfoMetricsConfig::default().with_interval(Duration::from_millis(500)),
+    );
+
+    // The worker primes CPU counters and sleeps ~200ms before its first sample.
+    // Without this wait, the first request would fold in the default zeros.
+    tokio::time::sleep(Duration::from_millis(500)).await;
```
rcoh, line 44: same as the RuntimeMetrics pr, lets put a `SysInfo.` prefix on these

jlizen: ++, I agree with this @fmzbl

metrique-util/src/sysinfo_reporter.rs
fmzbl: ill remove the default, but the the state has to have an initial value still.

metrique-util/src/sysinfo_reporter.rs
fmzbl: the reporter doesnt run on tokios timer, but let me see if i can refactor it without the sleep

rcoh (approved):

Issue #226, opened 2026-03-12
feat: CloudWatch `PutLogEvents` (EMF) `EntryIoStream` backend


jlizen:
# CloudWatch `PutLogEvents` (EMF) `EntryIoStream` backend

## Context

We'd like a metrique `EntryIoStream` implementation that submits metrics directly to CloudWatch via `PutLogEvents` using the [Embedded Metric Format](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html) (EMF). This would be a concrete consumer of the generalized async `EntryIoStream` backend proposed in [#225].

Today, metrique supports emitting metrics as EMF JSON to stdout/files, which CloudWatch Logs (via the CloudWatch Agent or container stdout routing) can pick up. A direct `PutLogEvents` backend would be useful for cases where:
* You want to publish metrics to CloudWatch without depending on a log ingestion agent or sidecar
* You need to publish to CloudWatch from environments where EMF log routing isn't available (e.g. bare EC2 without the CW Agent, or custom runtimes)
* You want programmatic control over the target log group/stream

### Why `PutLogEvents` with EMF, not `PutMetricData`?

Metrique is built around structured, unit-of-work metric entries that carry rich metadata: properties, high-cardinality dimensions, timestamps, and contextual fields alongside the numeric metric values. EMF preserves all of this. When CloudWatch Logs ingests an EMF log event, it extracts the declared metrics into CloudWatch Metrics while retaining the full log event (with all its properties) for querying via CloudWatch Logs Insights.

`PutMetricData`, by contrast, only accepts numeric metric values and dimensions. All the non-metric properties, string fields, and contextual data that metrique entries carry would be lost. Since this library is specifically optimized for emitting rich, structured metric entries, `PutLogEvents` with EMF is the natural fit.

A `PutMetricData` backend could also be supported in the future for use cases that only need raw metric values, but it would be a strictly less capable option. The main benefit I see would be some potential cost optimizations, though a
tight lifecycle policy on log entries would be probably comparable.

## How it would work

Using the async `EntryIoStream` framework from [#225], the CloudWatch backend would provide:
1. A **collector** that serializes metric entries as EMF JSON (reusing `metrique-writer-format-emf`) and accumulates them into `PutLogEvents` request payloads, respecting CloudWatch Logs' batch limits.
2. A **submitter** that calls the CloudWatch Logs `PutLogEvents` API using the AWS SDK.

[...]

PR #302, opened 2026-06-08
feat: add metrique-writer-cloudwatch crate

OliverGavin:
📬 *Issue #, if available:* #226

✍️ *Description of changes:* CloudWatch Logs PutLogEvents (EMF) EntryIoStream backend for metrique.

Implements EntryIoStream using a channel-based async architecture: sync next() → EMF serialize → batch → try_send over bounded mpsc → spawned tokio task calls PutLogEvents asynchronously.

Features:
- Respects CW Logs limits (1MB payload, 10k events per request)
- Backpressure: drops batch when channel full, reports via CwLogsSelfMetrics
- Graceful shutdown via CwLogsStreamHandle::shutdown().await
- Auto-creates log group/stream on startup (configurable)
- Logs rejected events from PutLogEvents response
- Builder API via bon

- [X] I confirm that I've made a best effort attempt to update all relevant documentation.
- [X] I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/metrique-writer-cloudwatch/src/lib.rs
+++ b/metrique-writer-cloudwatch/src/lib.rs
@@ -1,0 +1,1075 @@
+//! CloudWatch Logs `PutLogEvents` (EMF) backend for metrique.
+//!
+//! This crate provides [`CwLogsStream`], an [`EntryIoStream`] implementation that serializes
+//! metric entries as EMF JSON and submits them directly to CloudWatch Logs via `PutLogEvents`.
+//!
+//! # Architecture
+//!
+//! ```text
+//! App threads → BackgroundQueue (metrique) → std thread calls CwLogsStream::next(&entry)
+//!   → EMF serialize (reuses metrique-writer-format-emf)
+//!   → batch until size/count limit hit: try_send(batch) over tokio mpsc
+//!   → single tokio task: recv batch → client.put_log_events (async HTTP)
+//! ```
+//!
+//! # Backpressure
+//!
+//! When the submission channel is full (CloudWatch Logs is slow or down), batches are
+//! dropped at submission time with a rate-limited warning. Drops are reported through
+//! [`CwLogsStreamEvent::BatchDropped`]. Once the channel drains, normal submission resumes.
+//!
+//! # Shutdown
+//!
+//! [`CwLogsStream::builder()`] returns `(CwLogsStream, CwLogsStreamHandle)`. Call
+//! [`CwLogsStreamHandle::shutdown()`] to await drain of in-flight batches.
+//!
+//! # Example
+//!
+//! ```rust,ignore
+//! # async fn example() {
+//! use metrique_writer_cloudwatch::{CwLogsStream, CwLogsStreamConfig};
+//! use metrique_writer_core::EntryIoStream;
+//!
+//! let sdk_config = aws_config::load_from_env().await;
+//! let client = aws_sdk_cloudwatchlogs::Client::new(&sdk_config);
+//!
+//! let (stream, handle) = CwLogsStream::builder()
+//!     .client(client)
+//!     .log_group_name("/my-app/metrics".to_string())
+//!     .log_stream_name("host-1".to_string())
+//!     .namespace("MyApp".to_string())
+//!     .default_dimensions(vec![vec![]])
+//!     .build();
+//!
+//! // Pass `stream` to metrique's BackgroundQueue or ServiceMetrics::attach_to_stream()
+//! // ...
+//!
+//! // Graceful shutdown:
+//! handle.shutdown().await;
+//! # }
+//! ```
+
+use std::future::Future;
+use std::io;
+use std::num::NonZeroUsize;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use aws_sdk_cloudwatchlogs::Client as CloudWatchLogsClient;
+use aws_sdk_cloudwatchlogs::types::InputLogEvent;
+use bon::bon;
+use metrique_writer_core::format::Format;
+use metrique_writer_core::{Entry, EntryIoStream, IoStreamError};
+use metrique_writer_format_emf::Emf;
+use tokio::sync::{mpsc, oneshot};
+use tokio::task::JoinHandle;
+use tracing::{debug, warn};
+
+/// Maximum payload size for a single PutLogEvents request (1 MB).
+const MAX_BATCH_BYTES: usize = 1_048_576;
+/// Maximum size of a single log event message (1 MB).
+const MAX_EVENT_BYTES: usize = 1_048_576;
+/// Overhead per log event (timestamp + framing).
+const EVENT_OVERHEAD_BYTES: usize = 26;
+/// Maximum number of log events per PutLogEvents request.
+const MAX_BATCH_EVENTS: usize = 10_000;
+
+/// Configuration for [`CwLogsStream`].
+#[derive(Debug, Clone, bon::Builder)]
+pub struct CwLogsStreamConfig {
+    /// Capacity of the tokio mpsc channel between the collection thread and the
+    /// submission task. Default: 5.
+    #[builder(default = NonZeroUsize::new(5).unwrap())]
+    pub channel_capacity: NonZeroUsize,
+
+    /// Whether to auto-create the log group and stream if they don't exist. Default: true.
+    #[builder(default = true)]
+    pub auto_create: bool,
+}
+
+impl Default for CwLogsStreamConfig {
+    fn default() -> Self {
+        Self {
+            channel_capacity: NonZeroUsize::new(5).unwrap(),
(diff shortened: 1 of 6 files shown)


metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+            channel_capacity: 5,
+            auto_create: true,
+        }
+    }
+}
+
+/// Trait for reporting internal metrics about the CW Logs stream.
+pub trait CwLogsSelfMetrics: Send + Sync + std::fmt::Debug + 'static {
+    /// Called when a batch is dropped due to backpressure (submission channel full).
+    fn batch_dropped(&self, _event_count: usize) {}
```
rcoh: we recently used a different design in `BackgroundQueue` (#205)  where instead of calling a trait method, we pass an enum. That has much nicer back-compat properties because you can add fields to existing values (by putting `#[non_exhaustive]` on the variants, can you switch to that?

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+            self.tx.take();
+            self.submit_task.take();
+        }
+    }
+}
+
+impl EntryIoStream for CwLogsStream {
+    fn next(&mut self, entry: &impl Entry) -> Result<(), IoStreamError> {
+        if self.tx.is_none() {
+            warn!("CwLogsStream::next() called after shutdown, entry dropped");
```
rcoh, line 352: may be worth emitting a self-metric for this

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+        // newline-separated JSON objects (one per dimension set).
+        let mut buf = Vec::new();
+        self.emf.format(entry, &mut buf)?;
+
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_millis() as i64;
+
+        let output = String::from_utf8(buf).map_err(|e| IoStreamError::Io(io::Error::other(e)))?;
```
rcoh: this should basically never happen; it would be a major bug in the emf format. reporting this error as a metric would be better.

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+impl EntryIoStream for CwLogsStream {
+    fn next(&mut self, entry: &impl Entry) -> Result<(), IoStreamError> {
+        if self.tx.is_none() {
+            warn!("CwLogsStream::next() called after shutdown, entry dropped");
+            return Ok(());
+        }
+
+        // Serialize the entry to EMF JSON. The formatter may produce multiple
+        // newline-separated JSON objects (one per dimension set).
+        let mut buf = Vec::new();
```
rcoh: preallocate 1024 here?

rcoh (commented):
looks great! two small notes

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+            }
+            Err(mpsc::error::TrySendError::Closed(_)) => {
+                // TODO: clean up shutdown signaling when upstreaming to metrique —
+                // ideally BackgroundQueue would notify the stream directly.
+                self.tx.take();
+            }
+        }
+    }
+
+    fn batch_would_exceed_limits(&self, event_bytes: usize) -> bool {
```
jlizen, line 326: There is also a per-event limit (currently 1MB) that we need to check:
https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+        namespace: String,
+        #[builder(default = vec![vec![]])] default_dimensions: Vec<Vec<String>>,
+        #[builder(default)] config: CwLogsStreamConfig,
+        metric_reporter: Option<Box<dyn CwLogsSelfMetrics>>,
+    ) -> (Self, CwLogsStreamHandle) {
+        let (tx, rx) = mpsc::channel(config.channel_capacity);
+        let metrics: Arc<dyn CwLogsSelfMetrics> =
+            Arc::from(metric_reporter.unwrap_or_else(|| Box::new(NoOpMetrics)));
+
+        let submit_task = tokio::spawn(submit_loop(
```
jlizen: I'm fine to forego the full trait interface for v0, but we should at least support a custom spawner (though i'm fine with tokio as a default feature/behavior). This matters since we will want to support [dial9::spawn](https://docs.rs/dial9-tokio-telemetry/latest/dial9_tokio_telemetry/telemetry/fn.spawn.html) and other instrumented runtimes.

The spawner also lets us support secondary/non-ambient tokio runtime, and non-tokio runtime. (Ideally we could have `--no-default-features` not enable tokio)

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+            }
+        } else {
+            tracing::warn!("CwLogsStream already shut down when shutdown() was called");
+        }
+    }
+}
+
+#[bon]
+impl CwLogsStream {
+    /// Create a new [`CwLogsStream`].
```
jlizen, line 238: this needs a panics warning if used with tokio spawn in non-tokio context

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+/// Maximum number of log events per PutLogEvents request.
+const MAX_BATCH_EVENTS: usize = 10_000;
+
+/// Configuration for [`CwLogsStream`].
+#[derive(Debug, Clone, bon::Builder)]
+pub struct CwLogsStreamConfig {
+    /// Capacity of the tokio mpsc channel between the collection thread and the
+    /// submission task. Default: 5.
+    #[builder(default = 5)]
+    pub channel_capacity: usize,
```
jlizen: should be `NonZeroUsize` or some other form of validation/clamping

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+    }
+}
+
+async fn create_log_resources(
+    client: &CloudWatchLogsClient,
+    log_group_name: &str,
+    log_stream_name: &str,
+) {
+    // Best-effort creation — ignore "already exists" errors.
+    let _ = client
```
jlizen: Should at least log non-already-exists at warn/error (and i'd suggest logging already exists at debug level). Things like access denied would be pretty confusing here.

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+
+    fn enqueue_batch(&mut self) {
+        let Some(tx) = self.tx.as_ref() else { return };
+        let batch = std::mem::take(&mut self.batch);
+        let count = batch.len();
+        self.batch_bytes = 0;
+
+        match tx.try_send(WorkerCommand::PutLogEvents(batch)) {
+            Ok(()) => {}
+            Err(mpsc::error::TrySendError::Full(_)) => {
```
jlizen, line 312: Can we have a TODO for dropping oldest first

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,944 @@
...
+                self.metrics.batch_dropped(count);
+                warn!("CloudWatch Logs submission channel full, dropping batch of {count} events");
+            }
+            Err(mpsc::error::TrySendError::Closed(_)) => {
+                // TODO: clean up shutdown signaling when upstreaming to metrique —
```
jlizen, line 319: Can we mention in the TODO that we want to make sure that the worker flushes all in-flight batches during attach handle drop / explicit shutdown? Currently you have to explicitly flush it.

jlizen (commented):
Thanks for picking this up! I haven't read the PR in detail (or Russell's comments) but a few initial thoughts.

metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+        // Each line is a separate EMF JSON record → separate log event.
+        for line in output.lines() {
+            if line.is_empty() {
+                continue;
+            }
+            let event_bytes = line.len();
+
+            // Skip events that exceed the per-event size limit (1 MB).
+            if event_bytes > MAX_EVENT_BYTES {
+                warn!(
```
rcoh, line 386: I wonder if we should rate limit this log line

rcoh (approved):
LGTM!

jlizen:
(don't wait on my approval @rcoh , I have a wall of meetings, I'm on board with the general shape and it looks like my comments are addressed)

PR #310, opened 2026-06-09
feat: allow non-default MergeConfig in KeyedAggregator

### LINKED ISSUE #309: feat: Enable providing a non-default `MergeConfig` in `KeyedAggregator`
Right now https://docs.rs/metrique-aggregation/latest/metrique_aggregation/aggregator/struct.KeyedAggregator.html only works when `MergeConfig` impls default; we need to make it settable.

rcoh:
📬 *Issue #, if available:* Closes #309

✍️ *Description of changes:* Add `new_with_config(sink, merge_config)` constructor to `KeyedAggregator` that accepts an arbitrary `MergeConfig` without requiring it to implement `Default`.

The existing `new(sink)` constructor is preserved for backward compatibility when `MergeConfig: Default`.


metrique-aggregation/tests/keyed_aggregator_with_config.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+impl AggregateStrategy for Event {
+    type Source = Event;
+    type Key = EventByEndpoint;
+}
+
+#[test]
+fn test_keyed_aggregator_with_non_default_config() {
```
jlizen, line 93: Nit: might be nice to exercise `merge_ref()` as well

jlizen (approved):

metrique-aggregation/tests/keyed_aggregator_with_config.rs
rcoh: Good call — added `test_keyed_aggregator_with_non_default_config_merge_ref` (and a `MergeRef` impl for `Event`) that drives the same scenario through the `merge_ref()` path. Both tests assert the config-seeded total of 115. Done in a16e7da.

PR #315, opened 2026-06-18
fix(cloudwatch): use proper SdkError pattern matching for ResourceAlreadyExistsException

OliverGavin:
📬 *Issue #, if available:* Follow-up to #302

✍️ *Description of changes:*
The string-based check msg.contains("ResourceAlreadyExistsException") didn't match the actual SDK error format, causing spurious WARN logs on every startup when the log group/stream already exists.

Use SdkError::ServiceError pattern matching with the typed CreateLogGroupError::ResourceAlreadyExistsException / CreateLogStreamError::ResourceAlreadyExistsException variants instead. This correctly suppresses the expected "already exists" case to debug! level while still logging unexpected errors (e.g. AccessDenied) at warn!.

### DIFF
--- a/metrique-writer-cloudwatch/src/lib.rs
+++ b/metrique-writer-cloudwatch/src/lib.rs
@@ -69,6 +69,9 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use aws_sdk_cloudwatchlogs::Client as CloudWatchLogsClient;
+use aws_sdk_cloudwatchlogs::error::SdkError;
+use aws_sdk_cloudwatchlogs::operation::create_log_group::CreateLogGroupError;
+use aws_sdk_cloudwatchlogs::operation::create_log_stream::CreateLogStreamError;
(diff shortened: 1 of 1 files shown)


metrique-writer-cloudwatch/src/lib.rs
```diff
@@ -69,7 +69,10 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use aws_sdk_cloudwatchlogs::Client as CloudWatchLogsClient;
+use aws_sdk_cloudwatchlogs::operation::create_log_group::CreateLogGroupError;
+use aws_sdk_cloudwatchlogs::operation::create_log_stream::CreateLogStreamError;
 use aws_sdk_cloudwatchlogs::types::InputLogEvent;
+use aws_smithy_runtime_api::client::result::SdkError;
```
rcoh: This is re exported by the sdk, no need for a direct dep

rcoh (approved):

PR #316, opened 2026-06-21
feat: add TaskTiming field for per-request task metrics

fmzbl:
#317 

depends on [tokio-metrics pr](https://github.com/tokio-rs/tokio-metrics/pull/127)

Adds TaskTiming, a foldable field that attaches one request's tokio task metrics onto its own record. time spent idle, active, and in scheduling delay, plus poll and schedule counts. instead of crate-wide aggregates.


metrique-util/examples/individual-task-metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+        .instrument(async {
+            // Each loop iteration emits one EMF record carrying that request's
+            // own task metrics, e.g.:
+            //   {"Operation":"Read", "Success":1,
+            //    "TaskPollCount":3, "TaskTotalPollDuration":0.0001,
+            //    "TaskTotalIdleDuration":0.02, "TaskFirstPollDelay":0.0,
+            //    "TaskTotalScheduledDuration":0.0001, ...}
+            for op in ["Read", "Write", "Read"] {
+                let timing = TaskTiming::new();
+                let success = timing.instrument(handle_request(op)).await;
```
rcoh: there is no way in this API to control that this _was_ called only once, right? is that on purpose? what if you called instrument multiple times on the same `TaskTiming`?

metrique-util/examples/individual-task-metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    // The worker task is instrumented once. `record_request_scheduling` opts the
+    // monitor into publishing scheduling delay so each request's `TaskTiming`
+    // can observe it; without it, the scheduling fields would be zero.
+    let mut builder = TaskMonitor::builder();
+    builder.record_request_scheduling();
+    let worker_monitor = builder.build();
+
+    worker_monitor
+        .instrument(async {
```
rcoh: generally, instrumentation is only going to be accurate when a spawned task is instrumented so we should demonstrate it as such

metrique-util/examples/individual-task-metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let _attach_handle = ServiceMetrics::attach_to_stream(
+        Emf::all_validations("IndividualTaskMetricsExample".to_string(), vec![vec![]])
+            .output_to(std::io::stderr()),
+    );
+
+    // The worker task is instrumented once. `record_request_scheduling` opts the
+    // monitor into publishing scheduling delay so each request's `TaskTiming`
+    // can observe it; without it, the scheduling fields would be zero.
+    let mut builder = TaskMonitor::builder();
```
rcoh, line 48: could this be `TaskMonitor::builder().record_request_scheduling().build()`

metrique-util/examples/individual-task-metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+    builder.record_request_scheduling();
```
rcoh: from this context, its clear that `record_request_scheduling` is an overly specific name; we need to think of a name that accurately describes what this is doing at a Tokio/future level; it doesn't have anything specifically tied to the concept of a request.

metrique-util/src/individual_task_metrics.rs
```diff
@@ -0,0 +1,165 @@
...
+                    .await;
+
+                let entry = test_metric(RequestMetrics {
+                    operation: "Read",
+                    timing,
+                });
+
+                check!(entry.values["Operation"] == "Read");
+                check!(entry.metrics["TaskPollCount"] >= 2);
+                check!(entry.metrics["TaskTotalIdleDuration"] >= 1.0);
```
rcoh: this is deterministic so these assertions can be tightened

rcoh: I would also prefer that we test all fields to ensure they are all present.

metrique-util/src/lib.rs
```diff
@@ -27,6 +27,11 @@ pub use tokio_metrics_reporter::{
     AttachGlobalEntrySinkTokioMetricsExt, TokioRuntimeMetricsConfig, TokioRuntimeSnapshot,
 };
 
+#[cfg(feature = "tokio-metrics-bridge")]
+mod individual_task_metrics;
+#[cfg(feature = "tokio-metrics-bridge")]
+pub use individual_task_metrics::TaskTiming;
```
rcoh: `tokio_task_metrics`? not sure the best name. It isn't technically tokio specific, so maybe `future_metrics`?

metrique-util/src/individual_task_metrics.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn folds_per_request_metrics_into_entry() {
+        let mut builder = TaskMonitor::builder();
+        builder.record_request_scheduling();
+        let task_monitor = builder.build();
+        task_monitor
+            .instrument(async {
+                let timing = TaskTiming::new();
```
rcoh: there is some nuance on when this is accurate; it will be accurate for poll count always. The only place where the data is going to be inaccurate is in scheduling delay; in that case, it is generally only accurate if the root future (task) is instrumented. that's something we should document clearly

metrique-util/examples/individual-task-metrics.rs
fmzbl: i can make it so it fails on compile time when called more than once

metrique-util/examples/individual-task-metrics.rs
fmzbl: publish_scheduling_delay?

metrique-util/src/future_metrics.rs
```diff
@@ -0,0 +1,177 @@
...
+//! flattened onto that request's own metric record, the same way
+//! [`TokioRuntimeSnapshot`](crate::TokioRuntimeSnapshot) folds in runtime
+//! metrics.
+
+use std::future::Future;
+
+use metrique::CloseValue;
+use tokio_metrics::{FutureMonitor, FutureMetrics};
+
+/// A metrique field holding the captured Tokio task metrics of a single request.
```
rcoh:
```suggestion
/// A metrique field holding the captured Tokio task metrics of a single future.
```

metrique-util/src/future_metrics.rs
```diff
@@ -0,0 +1,177 @@
...
+///
+/// Wrap the request's future with [`TaskTiming::instrument`]; awaiting the
```
rcoh:
```suggestion
/// Wrap the future with [`TaskTiming::instrument`]; awaiting the
```

metrique-util/src/future_metrics.rs
```diff
@@ -0,0 +1,177 @@
...
+/// returned future yields the future's output together with a `TaskTiming`. Fold
+/// that into your metric struct with `#[metrics(flatten)]` and, on close, it
+/// emits the request's poll, idle, first-poll, and scheduling metrics (see
+/// [`FutureMetrics`](tokio_metrics::FutureMetrics) for the full list).
+///
+/// `instrument` is a one-shot associated function that takes ownership of the
+/// future, so a `TaskTiming` always describes exactly one request — there is no
+/// way to accidentally reuse it across futures.
```
rcoh:
```suggestion
```

delete this

metrique-util/src/future_metrics.rs
```diff
@@ -0,0 +1,177 @@
...
+/// **Scheduling delay is the one exception.** The time a task spends queued
+/// between being woken and being polled can only be observed by the *root*
+/// future the runtime actually schedules — a nested future can't see it. So
+/// `scheduled_count` / `total_scheduled_duration` / `long_delay_count` are only
+/// populated when the surrounding (root) task is instrumented with a monitor
+/// built via
+/// [`TaskMonitorBuilder::publish_scheduling_delay`](tokio_metrics::TaskMonitorBuilder::publish_scheduling_delay),
+/// **and** that instrumented future is the root of a spawned task (see the
+/// example). Without that, those three fields read zero while everything else
+/// stays accurate.
```
rcoh: Can we say this same thing in fewer words and directly get to the heart of the matter?

Measuring scheduling delay accurately requires that the root task has been instrumented with `TaskMonitorBuilder` with `publish_scheduling_delay` enabled. If you don't, `scheduled_count`, <...> will be unset. (Is that correct?). Poll time, idle time and first poll metrics are locally measured and will always be accurate.

metrique-util/src/lib.rs
```diff
@@ -27,6 +27,11 @@ pub use tokio_metrics_reporter::{
     AttachGlobalEntrySinkTokioMetricsExt, TokioRuntimeMetricsConfig, TokioRuntimeSnapshot,
 };
 
+#[cfg(feature = "tokio-metrics-bridge")]
+mod future_metrics;
+#[cfg(feature = "tokio-metrics-bridge")]
+pub use future_metrics::TaskTiming;
```
rcoh: we can drop this re-export

metrique-util/src/future_metrics.rs
fmzbl: yes correct

rcoh (approved):

jasgin:
Looks like there are some clippy issues to resolve

fmzbl:
it depends on a release of tokio-metrics with [https://github.com/tokio-rs/tokio-metrics/pull/127#event-27922265461](https://github.com/tokio-rs/tokio-metrics/pull/127#event-27922265461)

fmzbl:
not a lint issue

jv1i:
This can now depend on tokio-metrics v0.5.2

jasgin (approved):

rcoh (approved):

PR #329, opened 2026-07-02
Add SCREAMING_SNAKE_CASE support to metrique-macro NameStyle

### LINKED ISSUE #328: rename_all on value(string) enums generates Entry type instead of Value type
## Bug

When `rename_all` is used on a `#[metrics(value(string))]` enum, the macro takes the entry enum codegen path instead of the value enum path.

## Repro

```rust
#[metrics(value(string), rename_all = "SCREAMING_SNAKE_CASE")]
#[derive(Debug, Clone, Copy)]
pub enum Operation {
    Token,
    ResourceToken,
}
```

Using this enum as a field inside another `#[metrics]` struct:

```rust
#[metrics(rename_all = "PascalCase", exact_prefix = "MyApp:")]
pub struct MyMetrics {
    pub operation: Option<Operation>,
}
```

## Expected

The macro generates an `OperationValue` type with `impl Value`, allowing the enum to be used as a nested field inside another metrics struct (same behavior as using explicit `#[metrics(name = "...")]` per variant without `rename_all`).

## Actual

The macro generates an `OperationEntry` type (entry enum codegen path) with `append_on_drop` on the enum itself. No `impl Value` is generated for the entry type. Compilation fails with:

```
error[E0277]: `OperationEntry` is not a metric value
```

## Additional issue

Single-word variants don't get the case transformation applied. For example, `Token` produces `"Token"` instead of `"TOKEN"` in the `From<&Operation> for &'static str` impl, while multi-word variants like `ResourceToken` correctly produce `"RESOURCE_TOKEN"`.

## Workaround

Use explicit `#[metrics(name = "...")]` on each variant without `rename_all`:

```rust
#[metrics(value(string))]
pub enum Operation {
    #[metrics(name = "TOKEN")]
    Token,
    #[metrics(name = "RESOURCE_TOKEN")]
    ResourceToken,
}
```

## Version

metrique-writer-macro 0.1.9

fbergero:
## Summary

Fixes #328

The `metrique-writer-macro` crate supports `SCREAMING_SNAKE_CASE` in its `NameStyle` enum, but `metrique-macro` did not. This caused `#[metrics(value(string), rename_all = "SCREAMING_SNAKE_CASE")]` to silently fail — darling couldn't parse the attribute, so the macro recovered with a default `RootEntry` mode instead of `ValueString`, generating an `*Entry` type (missing `impl Value`) instead of the correct `*Value` type.

## Changes

- Add `ScreamingSnakeCase` variant to `metrique-macro/src/inflect.rs` `NameStyle` enum with `apply()`, `apply_prefix()`, and `to_word()` implementations
- Handle new variant in `entry_impl.rs` `make_ns()` (maps to `NS::SnakeCase` as closest runtime equivalent)
- Add snapshot test for the macro codegen
- Add integration test verifying value enums compile and rename correctly when used as fields inside a metrics struct

## Note

For entry enums/structs, `ScreamingSnakeCase` maps to the runtime `NS::SnakeCase` type since there's no `ScreamingSnakeCase` type in `metrique-core`. Full runtime type-level support would require a new type there — left for a follow-up if needed.

## Test plan

- [x] New snapshot test `test_value_enum_screaming_snake_case` passes
- [x] New integration test `value_enum_screaming_snake_case.rs` verifies:
  - Multi-word variants renamed correctly (`ReadData` → `"READ_DATA"`)
  - Explicit `#[metrics(name = "...")]` overrides still work
  - Value enum usable as a field inside another `#[metrics]` struct
- [x] All existing tests continue to pass

### DIFF
--- a/metrique-macro/src/entry_impl.rs
+++ b/metrique-macro/src/entry_impl.rs
@@ -38,7 +38,9 @@ pub(crate) fn mixed_site_self() -> Ident {
 fn make_ns(ns: NameStyle, span: proc_macro2::Span) -> Ts2 {
     match ns {
         NameStyle::PascalCase => quote_spanned! {span=> NS::PascalCase },
-        NameStyle::SnakeCase => quote_spanned! {span=> NS::SnakeCase },
+        NameStyle::SnakeCase | NameStyle::ScreamingSnakeCase => {
+            quote_spanned! {span=> NS::SnakeCase }
+        }
         NameStyle::KebabCase => quote_spanned! {span=> NS::KebabCase },
         NameStyle::Preserve => quote_spanned! {span=> NS },
     }
--- a/metrique-macro/src/inflect.rs
+++ b/metrique-macro/src/inflect.rs
@@ -24,6 +24,8 @@ pub(crate) enum NameStyle {
     PascalCase,
     #[darling(rename = "snake_case")]
     SnakeCase,
+    #[darling(rename = "SCREAMING_SNAKE_CASE")]
+    ScreamingSnakeCase,
     #[darling(rename = "kebab-case")]
     KebabCase,
     #[default]
@@ -36,6 +38,7 @@ impl NameStyle {
         match self {
             NameStyle::PascalCase => name.to_pascal_case(),
             NameStyle::SnakeCase => name.to_snake_case(),
+            NameStyle::ScreamingSnakeCase => name.to_screaming_snake_case(),
             NameStyle::Preserve => name.to_string(),
             NameStyle::KebabCase => name.to_kebab_case(),
         }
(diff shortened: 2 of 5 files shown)


jlizen (commented):
This looks good but can we also handle structs and entry enums? Should be a similar change.

I would also love a test that validates we don't omit n+1 styles, though I could sweep it up in https://github.com/awslabs/metrique/pull/289 if you don't get to it or it is missing public types. Just drop me a note there maybe if so?

Ref: [descriptor.rs](https://github.com/jlizen/metrique/blob/e430c6b0c25a75b297bbaebc73da2c2744f7a3e0/metrique-writer-core/src/descriptor.rs#L61)

jlizen:
Oh, also we should have a snapshot test for the entry struct path which currently is broken too

jlizen (approved):
Looks like I need to add screaming snake case in #289 in many more places.

We can just merge this as is and i'll fix the rest on top :)

Thanks for the fix!

PR #331, opened 2026-07-03
[metrique-writer-core + metrique-macro] Resolve FieldShape and Unit from Value trait

jlizen:
📬 *Issue #, if available:*

✍️ *Description of changes:*

Follows #289 (entry descriptors). Descriptors previously reported `Opaque` for all field shapes and only surfaced units from explicit `#[metrics(unit = X)]`.

### Summary

Adds `const SHAPE` and `const UNIT` as defaulted associated consts on the `Value` trait. The macro resolves both through the `CloseValue::Closed` chain at compile time, so descriptors now carry type-accurate shapes and inherent units for all built-in types with zero user effort and zero runtime cost.

The approach avoids breakage: existing `Value` impls compile unchanged (they inherit `Opaque` and `Unit::None`). Custom types that close to a known type (e.g. a counter closing to `u64`) get the correct shape and unit from the closed type's impl. Users who want richer metadata for custom `Value` types override the consts.

`ValueFormatter` also gets a `const SHAPE` with the same pattern. Formatter-aware sinks can query it separately, though the descriptor's field shape reflects the closed type (not the formatter's rendering).

Lifetime parameters in field types are handled by anonymizing them to `'_` in the macro-generated static expressions, which the compiler resolves correctly since `SHAPE`/`UNIT` are `'static` regardless of the input lifetime.

### Testing

Shape tests cover the full type matrix: all scalar types, Duration/Timer (through CloseValue), Optional, List, nested composition (Option<Vec<T>>), lifetimed types (Cow<'a, str>, Option<Cow<'a, str>>), custom CloseValue resolution, no_close fields falling to Opaque, custom SHAPE overrides, `#[metrics(value)]` struct/enum forwarding, Arc wrapper passthrough, and formatted fields.

A new e2e test (`descriptor_shape_sink.rs`) demonstrates the intended consumption pattern: a sink builds a typed schema from descriptor shapes, units, and flags, then caches it by DescriptorId.

### DIFF
--- a/metrique-util/src/tokio_metrics_reporter.rs
+++ b/metrique-util/src/tokio_metrics_reporter.rs
@@ -323,9 +323,39 @@ mod tests {
         check!(entry.metrics["workers_count"] == 1);
         check!(entry.metrics["elapsed"] > 0.0);
         check!(entry.metrics["total_park_count"] > 0);
+    }
 
-        #[cfg(tokio_unstable)]
-        check!(entry.metrics["poll_time_histogram"].num_observations() > 0);
+    #[cfg(tokio_unstable)]
+    #[test]
+    fn subscribe_appends_metrics_identity_histogram() {
+        let rt = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(2)
+            .enable_all()
+            .enable_metrics_poll_count_histogram()
+            .build()
+            .unwrap();
+        rt.block_on(async {
+            metrique_writer::sink::global_entry_sink! { Sink }
+            let TestEntrySink { inspector, sink } = test_entry_sink();
+            let _handle = Sink::attach((sink, ()));
+
+            Sink::subscribe_tokio_runtime_metrics(
+                TokioRuntimeMetricsConfig::default().with_interval(Duration::from_millis(50)),
+            );
+
+            // Spawn work and wait so the runtime records poll times.
+            tokio::spawn(async {
+                std::thread::sleep(std::time::Duration::from_millis(1));
+            })
+            .await
+            .unwrap();
+            tokio::time::sleep(Duration::from_millis(200)).await;
+
+            let entries = inspector.entries();
+            check!(!entries.is_empty());
+            let entry = entries.last().unwrap();
+            check!(entry.metrics["poll_time_histogram"].num_observations() > 0);
+        });
     }
 
     #[tokio::test(start_paused = true)]
@@ -349,9 +379,6 @@ mod tests {
         check!(entry.metrics["WorkersCount"] == 1);
         check!(entry.metrics["Elapsed"] > 0.0);
         check!(entry.metrics["TotalParkCount"] > 0);
-
-        #[cfg(tokio_unstable)]
-        check!(entry.metrics["PollTimeHistogram"].num_observations() > 0);
     }
 
     #[tokio::test(start_paused = true)]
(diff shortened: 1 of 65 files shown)


metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -323,9 +323,39 @@ mod tests {
         check!(entry.metrics["workers_count"] == 1);
         check!(entry.metrics["elapsed"] > 0.0);
         check!(entry.metrics["total_park_count"] > 0);
+    }
 
-        #[cfg(tokio_unstable)]
-        check!(entry.metrics["poll_time_histogram"].num_observations() > 0);
+    #[cfg(tokio_unstable)]
+    #[test]
+    fn subscribe_appends_metrics_identity_histogram() {
```
rcoh: is this test related to the rest of the PR? this test seems almost guaranteed to be flaky. can it be changed to `start_paused = true`?

metrique-util/src/tokio_metrics_reporter.rs
```diff
@@ -323,9 +323,39 @@ mod tests {
...
+                TokioRuntimeMetricsConfig::default().with_interval(Duration::from_millis(50)),
+            );
+
+            // Spawn work and wait so the runtime records poll times.
+            tokio::spawn(async {
+                std::thread::sleep(std::time::Duration::from_millis(1));
+            })
+            .await
+            .unwrap();
+            tokio::time::sleep(Duration::from_millis(200)).await;
```
rcoh: seems flaky

metrique/docs/extending.md
```diff
@@ -203,6 +205,20 @@ struct MyMetric {
 }
 ```
 
+Formatters that always produce a known wire type should override `ValueFormatter::SHAPE` so descriptor-aware sinks can pre-register the field correctly. `AsUtcDate` always writes a string, so:
```
rcoh: doesn't every formatter _inherently_ write a string? like...it literally cannot do anything else

rcoh:
overall looks good. We are missing `UNIT` passthroughs on wrapper impls.

I suggest removing the default impls behind a `cfg` gate that we test in our CI matrix

metrique/docs/extending.md
jlizen: You can call `writer.metric()` from inside a value formatter, but I can document it more clearly.

metrique-util/src/tokio_metrics_reporter.rs
jlizen: This is nothing new, I'll just ditch it.

jlizen:
@rcoh ready for fresh review. I added a similar cfg-backed ci gate to entry descriptors, good call.

metrique-writer-core/src/entry/mod.rs
```diff
@@ -190,9 +190,13 @@ pub trait Entry {
     /// Each descriptor covers a contiguous segment of the `Entry::write` output.
     /// Simple entries yield one descriptor. Composed entries (like aggregation results)
     /// yield multiple. Hand-written entries return an empty iterator by default.
+    #[cfg(not(metrique_require_explicit_impls))]
```
rcoh, line 193: we can do it another PR, but there are probably other methods we should add this to

rcoh (approved):

metrique-writer-core/src/entry/mod.rs
jlizen: https://github.com/awslabs/metrique/issues/336

Issue #230, opened 2026-03-12
feat: Add convenience API for `FlushImmediately` setup


jlizen:
## Add convenience API for `FlushImmediately` setup

The current `FlushImmediately` setup requires `FlushImmediately::new_boxed(...)` + `ServiceMetrics::attach((sink, ()))`, which is hard to discover and verbose compared to `ServiceMetrics::attach_to_stream(...)` for `BackgroundQueue`.

Consider adding something like `ServiceMetrics::write_directly_to(...)` (or similar) that mirrors `attach_to_stream` but uses `FlushImmediately` internally.

From PR #219 review feedback (rcoh): "this is gross enough I think its worth adding `ServiceMetrics::write_directly_to` or something. I routinely see people get stuck trying to find `new_boxed`."

Once this lands, update the `FlushImmediately` example in `metrique/docs/sinks.md` (the "Immediate Flushing for ephemeral environments" section) to use the new API.

jv1i:
Solved in #333

PR #333, opened 2026-07-06
feat: Add convenience API for FlushImmediately setup

tiny-ac:
📬 *Issue #, if available:*
https://github.com/awslabs/metrique/issues/230

✍️ *Description of changes:*
- `metrique-writer/src/sink/mod.rs`
  -  Add `write_directly_to` to `AttachGlobalEntrySinkExt` as a one-line convenience method for attaching a `FlushImmediately` sink to a global sink - mirroring the existing `attach_to_stream` for `BackgroundQueue` + tests.
- `metrique/docs/sinks.md`
  - Update "Immediate Flushing for ephemeral environments" example to demonstrate the new API usage

### DIFF
--- a/metrique/docs/sinks.md
+++ b/metrique/docs/sinks.md
@@ -68,10 +68,9 @@ struct MyMetrics {
 
 fn main() {
     let _handle = ServiceMetrics::write_directly_to(
-        Emf::no_validations(
-            "MyNS".to_string(),
-            vec![vec![/*your dimensions here*/]],
-        ).output_to(std::io::stdout()),
+        Emf::builder("MyNS".to_string(), vec![vec![/*your dimensions here*/]])
+            .build()
+            .output_to(std::io::stdout()),
     );
     handle_request();
 }
--- a/metrique/docs/sinks.md
+++ b/metrique/docs/sinks.md
@@ -58,8 +58,7 @@ and requires a custom graceful shutdown to drain. For these cases, consider usin
 ```rust
 use metrique::emf::Emf;
 use metrique::ServiceMetrics;
-use metrique::writer::{AttachGlobalEntrySink, FormatExt, GlobalEntrySink};
-use metrique::writer::sink::FlushImmediately;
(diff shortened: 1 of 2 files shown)


metrique/docs/sinks.md
```diff
@@ -68,14 +67,12 @@ struct MyMetrics {
 }
 
 fn main() {
-    let sink = FlushImmediately::new_boxed(
+    let _handle = ServiceMetrics::write_directly_to(
         Emf::no_validations(
```
tiny-ac: Noticed these two examples use different construction styles, which got me curious - why does this example use `Emf::no_validations(...)` here whereas the `Background Queue` [example above](https://github.com/awslabs/metrique/blob/main/metrique/docs/sinks.md?plain=1#L45-L49) uses `Emf::builder(...).build()`? Is there a specific reason to intentionally skip validations here?

tiny-ac: I'd lean towards switching this example to `Emf::builder().build()` too, It seems safe enough by default (validates in debug builds, skips in release) and keeps the examples consistent.

The [doc comment on no_validations](https://github.com/awslabs/metrique/blob/4fd22fa8aa10b2c7aeb25f9c661164baffee5b3a/metrique-writer-format-emf/src/emf.rs#L389-L392) itself frames it as a narrow performance opt-in rather than a default, so it may not be suitable for a getting-started example.

For reference, here's a real-world use of the `Emf::builder().build()` constructor from the dial9 repo (for attach_to_stream):
https://github.com/dial9-rs/dial9/blob/88f2bdde4e969b78abcb9c0d62f52c37865e6efc/dial9-viewer/src/server/metrics.rs#L320C9-L323

I couldn't find anything similar for FlushImmediately in that repo, plus it uses FlushImmediatelyBuilder instead
https://github.com/dial9-rs/dial9/blob/88f2bdde4e969b78abcb9c0d62f52c37865e6efc/examples/metrics-service/src/main.rs#L224-L225

rcoh (commented):
looks good! Please add a test as well

rcoh (approved):
LGTM!

PR #338, opened 2026-07-13
test: add Shuttle-based concurrency tests

### LINKED ISSUE #281: use shuttle to test background queue and other primitives

tiny-ac:
📬 *Issue #, if available:*

closes https://github.com/awslabs/metrique/issues/281

✍️ *Description of changes:*

Added Shuttle for concurrency testing, as well as some regular tests to improve coverage.

Historical bugs this was checked against:
  1. WorkerSink never exits on channel disconnect (f5baeaeda3f4b111f39ddd9adda21b9a291f9a62)
     - Full suite: cargo test -p metrique-aggregation --no-fail-fast (~65 tests)
       - 1 failure (worker_flushes_and_exits_when_all_senders_dropped)
     - shuttle suite: all 6 of 6 shuttle tests fail.
  2. WorkerSink periodic flush starves under continuous load (e44968edc7ec99669bf673d14fb227646c00088f)
     - Full suite: cargo test -p metrique-aggregation --no-fail-fast
         - 1 failure (flushes_periodically_under_continuous_load)
      - shuttle suite (6 tests) → 6/6 pass, unaffected.
  3. MutexSink::close panics with outstanding Arc refs (c407a74ec26980215b255c3703f969ca4bc6a209)
     - Full suite: cargo test -p metrique-aggregation --no-fail-fast (13 test binaries)
       - 1 failure (test_mutex_sink_close_with_outstanding_references)
     - shuttle suite → 6/6 unaffected (file isn't shuttle-tested).
       -  this bug is a sequential ownership-check mistake, not a race, so shuttle coverage wouldn't add anything here even if it existed
  4. BackgroundQueue lost a flush_async() wakeup while parked (pre-open-source diff, no commit object in this repo)
     - Full suite: shuttle suite
      - 6 passed, 2 failed (flush_waits_for_prior_pushes{,_determinism})
        - non-shuttle sweep additionally found flush_without_waiting failing (aborts after its own 60s internal
  timeout)
  5. BackgroundQueue periodic flush starves under continuous load (pre-open-source diff, no commit hash at all)
     - Full suite: non-shuttle sweep
       - 3 failures (next_loop_action_table, flushes_periodically_when_writing, flush_never_empty)
     - shuttle suite (8 tests): unaffected.
    
Also checked these while doing a wider search, no concurrency gaps found here so no action was taken
  - Two !Send/!Sync thread-local guard bugs (9a73c8ee5a8c9adfab8aa40fc517c7cc12035dec, 1e58030f61f15cae741d0365c85bee9a816ebcd9)
  - Timer::stop wasn't idempotent (0e44c8bd470b76123a9ea7c7cf0507a4b2a839bb)


rcoh:
overall looks good; there have been a couple of bugs in BackgroundQueue that have been fixed over time; it would be good to back those fixes out and make sure we can catch them with the shuttle tests

metrique-writer/src/sink/background.rs
```diff
@@ -695,20 +693,19 @@ impl<S: EntryIoStream, E: Entry> Receiver<S, E> {
...
-                    break; // shut down, break out of loop to have a chance to flush stream
-                }
-
-                // if the waker tracker can make progress observing an empty queue, let it
-                if !waker_tracker.will_progress_on_drained_queue() {
-                    let park_start = Instant::now();
-                    self.parker.park_deadline(next_flush);
-                    if self.inner.observer.is_some() {
-                        idle_duration += park_start.elapsed();
+                match next_loop_action(
```
tiny-ac, line 695: Refactored for testing purposes. There's a historical bug caused by a missing branch in an if chain (fixed in an internal, pre-open-source diff). Extracting it allows table-testing every input combination (no threads/clock/Shuttle needed).

This touches a hot, widely-used loop, so I checked the behavior by running a test that re-implemented the old code verbatim and diffed its output against the new function (temporarily added derive(Debug) to DrainResult for this).

  ```rust
  #[test]
  fn next_loop_action_matches_pre_refactor_inline_logic() {
      fn old_inline_logic(
          status: DrainResult,
          shutdown_requested: bool,
          can_progress_on_drained_queue: bool,
      ) -> LoopAction {
          if status == DrainResult::HitDeadline { return LoopAction::Flush; }
          if shutdown_requested { return LoopAction::Flush; }
          if can_progress_on_drained_queue { LoopAction::Continue } else { LoopAction::Park }
      }
      for status in [DrainResult::Drained, DrainResult::HitDeadline] {
          for shutdown_requested in [false, true] {
              for can_progress in [false, true] {
                  assert_eq!(
                      next_loop_action(status, shutdown_requested, can_progress),
                      old_inline_logic(status, shutdown_requested, can_progress),
                      "mismatch for status={status:?}, shutdown_requested={shutdown_requested}, \
                       can_progress_on_drained_queue={can_progress}"
                  );
              }
          }
      }
  }
  ```

Checked that it'd actually have caught the historical bug: removed the `HitDeadline` branch from next_loop_action and reran, which caused a failure:

  ```
  assertion `left == right` failed: mismatch for status=HitDeadline, shutdown_requested=false, can_progress_on_drained_queue=false
    left: Park
   right: Flush
  ```

metrique-writer-core/src/global.rs
```diff
@@ -364,7 +381,7 @@ impl Drop for TokioRuntimeTestSinkGuard {
 impl Drop for AttachHandle {
     fn drop(&mut self) {
         if let Some(arc) = self.shutdown_registry.take() {
-            // The macro holds only a Weak reference, so this is the sole strong ref.
+            // KNOWN BUG (see issue https://github.com/awslabs/metrique/issues/340)
```
tiny-ac, line 384: While Shuttle-testing `AttachHandle`/`ShutdownRegistry`, found that `drop()`'s `Arc::try_unwrap` can observe two strong references (racing a concurrent `register_shutdown_fn`) and panic on what the code assumes is unreachable. Issue #340 was filed.

Reproduced it with the existing (`#[ignore]`d, exactly for this reason) Shuttle test:
```
RUSTFLAGS="--cfg shuttle --cfg tokio_unstable" cargo test -p metrique-writer-core --lib --features _shuttle -- --ignored concurrent_register_and_drop
```
Both variants fail on the very first run:
```
thread 'global::shuttle_tests::concurrent_register_and_drop_determinism' panicked at metrique-writer-core/src/global.rs:387:27:
internal error: entered unreachable code: ShutdownRegistry should have no other strong references

thread 'global::shuttle_tests::concurrent_register_and_drop_pct' panicked at metrique-writer-core/src/global.rs:387:27:
internal error: entered unreachable code: ShutdownRegistry should have no other strong references

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 59 filtered out; finished in 0.01s
```

metrique-util/src/pending_sink.rs
```diff
@@ -5,54 +5,186 @@
...
+use crossbeam_queue::ArrayQueue;
+#[cfg(all(shuttle, feature = "_shuttle"))]
+use shuttle_impl::ArrayQueue;
+
+use once_slot::OnceSlot;
+
 struct Inner {
     buffer: ArrayQueue<BoxEntry>,
-    sink: OnceLock<BoxEntrySink>,
+    sink: OnceSlot<BoxEntrySink>,
```
tiny-ac, line 42: `OnceLock::get()` is a plain atomic load — invisible to Shuttle's scheduler (it only instruments its own re-implemented primitives). Without a shuttle-visible substitute, a Shuttle test here would explore zero interleavings of the sink slot itself.

The shuttle-only path clones the value and releases its `Mutex` guard before calling `f`, so arbitrary sink code never runs while holding the shim's internal lock.

Worth noting: while rewriting this, a bug was introduced in `PendingSink::append` (`metrique-util/src/pending_sink.rs`) which was then caught by one of the shuttle tests added:

```rust
// The regression (this is what actually shipped briefly during the rewrite):
let entry = forward_or_return(&self.0.sink, entry);
self.0.buffering.fetch_sub(1, Ordering::AcqRel);   // decrement moved here...
if let Some(entry) = entry {
    self.0.buffer.force_push(entry);               // ...ahead of the push
}
```
Caught by the existing Shuttle test on the very next run:
```
RUSTFLAGS="--cfg shuttle --cfg tokio_unstable" cargo test -p metrique-util --lib --features _shuttle -- concurrent_appends_racing_resolve_lose_nothing
```
```
assertion `left == right` failed
  left: 3
 right: 4
```

metrique/src/keep_alive.rs
```diff
@@ -266,4 +280,150 @@ mod test {
...
+            self.count.fetch_add(1, Ordering::SeqCst);
+        }
+    }
+
+    /// The value must be released exactly once: not zero (a leak)
+    /// and not more than once (a double drop, which given `Parent`'s
+    /// `unsafe impl Send`/`Sync` and `UnsafeCell` would risk real
+    /// undefined behavior, not just a logic bug).
+    #[test]
+    fn stress_concurrent_guard_and_drop_all_race_releases_value_exactly_once() {
```
tiny-ac, line 356: Implemented this with real OS threads since the Shuttle version of this test didn't add much value. `Arc`/`Weak` aren't Shuttle-instrumented (confirmed directly from the `shuttle` 0.9.1 source: `shuttle::sync::{Arc, Weak}` are bare re-exports of `std::sync`, not shuttle-native types — only `Mutex` is), so real OS thread scheduling can reach about as well as Shuttle can.

metrique/src/keep_alive.rs
```diff
@@ -266,4 +280,150 @@ mod test {
...
+        fn drop(&mut self) {
+            self.count.fetch_add(1, Ordering::SeqCst);
+        }
+    }
+
+    /// Two `DropAll`s dropped concurrently, racing each other (and
+    /// `Parent`'s own drop) on the same slot. Unlike a `DropAll` racing
+    /// ordinary `Guard`s (which races inside `Arc`'s own atomics, outside
+    /// what Shuttle instruments), this race sits entirely between two
+    /// `Mutex` operations Shuttle does control.
```
tiny-ac: Injected bug to check if this test would catch it.

Temporarily changed Drop implementation
```rust
impl Drop for DropAll {
    fn drop(&mut self) {
        if let Some(guard) = self.0.upgrade() {
            if let Some(f) = guard.lock().unwrap().take() {
                (f)()
            }
        }
    }
}
```
to
```rust
impl Drop for DropAll {
    fn drop(&mut self) {
        if let Some(guard) = self.0.upgrade() {
            let has_fn = guard.lock().unwrap().is_some();
            if has_fn {
                let f = guard.lock().unwrap().take().unwrap();
                (f)()
            }
        }
    }
}
```
Ran the new Shuttle test against this mutation:
```
RUSTFLAGS="--cfg shuttle" cargo test -p metrique --lib --features _shuttle -- concurrent_drop_alls_race_releases_value_exactly_once
```
Failed on the first run, every time:
```
thread '...' panicked at metrique/src/keep_alive.rs:90:54:
called `Option::unwrap()` on a `None` value
```

Ran the following to check if this bug was invisible to real-thread testing:
```rust
// THROWAWAY: not part of the permanent suite.
#[test]
fn temp_real_thread_two_drop_alls_race() {
    use std::sync::Barrier;

    for _ in 0..50_000 {
        let (tester, drop_count) = DropCounter::new();
        let primary = Parent::new(tester);

        let drop_all_1 = primary.force_drop_guard();
        let drop_all_2 = primary.force_drop_guard();

        let barrier = Arc::new(Barrier::new(3));

        let h1 = {
            let barrier = barrier.clone();
            std::thread::spawn(move || {
                barrier.wait();
                drop(drop_all_1);
            })
        };
        let h2 = {
            let barrier = barrier.clone();
            std::thread::spawn(move || {
                barrier.wait();
                drop(drop_all_2);
            })
        };

        barrier.wait();
        drop(primary);

        h1.join().unwrap();
        h2.join().unwrap();

        assert_eq!(drop_count.load(Ordering::SeqCst), 1);
    }
}
```

```
cargo test -p metrique --lib --release temp_real_thread_two_drop_alls_race -- --nocapture
```
```
test keep_alive::test::temp_real_thread_two_drop_alls_race ... ok

test result: ok. 1 passed; 0 failed
```
Zero failures at 50,000 iterations, against a mutation Shuttle catches on
its first run.

metrique-aggregation/src/sink/worker.rs
```diff
@@ -147,4 +167,247 @@ mod tests {
...
+    /// comment: `recv_timeout` never actually times out), so any value
+    /// works; 60s just documents "this is not what's under test."
+    fn flush_interval() -> Duration {
+        Duration::from_secs(60)
+    }
+
+    /// Entries sent concurrently from several cloned handles are all merged
+    /// by the time `flush()`'s returned future resolves, for every
+    /// interleaving shuttle explores.
+    fn concurrent_sends_all_merged_before_flush_returns() {
```
tiny-ac: Reverting [commit f5baeae](https://github.com/awslabs/metrique/commit/f5baeaeda3f4b111f39ddd9adda21b9a291f9a62) and rerunning both this test and `concurrent_drops_exit_cleanly_and_flush_once` together, Shuttle didn't just fail an assertion, it hit its own infinite-loop safety bound (`exceeded max_steps bound 1000000 ... this might be caused by an unfair schedule (e.g., a spin loop)?`)

Also stress-tested at 50,000 adversarial iterations (`check_uncontrolled_nondeterminism`) with no failures on clean code.

metrique-aggregation/src/sink/worker.rs
```diff
@@ -147,4 +167,247 @@ mod tests {
...
+            2_000,
+        );
+    }
+
+    /// The historical bug, reproduced directly: several cloned handles send
+    /// an entry and drop concurrently. The background thread must still
+    /// exit (this used to hang forever -- see the module doc comment),
+    /// flush exactly once at shutdown, and lose no entries, no matter how
+    /// the drops and the final disconnect interleave.
+    fn concurrent_drops_exit_cleanly_and_flush_once() {
```
tiny-ac: Reverted [commit f5baeae](https://github.com/awslabs/metrique/commit/f5baeaeda3f4b111f39ddd9adda21b9a291f9a62) 
```rust
// replace:
Err(RecvTimeoutError::Timeout) => { inner.flush(); last_flush = Instant::now(); }
Err(RecvTimeoutError::Disconnected) => { inner.flush(); return; }
// with:
Err(_) => { inner.flush(); last_flush = Instant::now(); }
```
All 6 of 6 shuttle tests fail (every shuttle test in this file eventually joins or drops the background thread, so none can complete if it never exits — not exclusive to this one test, but this is the one written
specifically for this bug shape). Reverted; `diff` clean; 6/6 shuttle + full non-shuttle suite pass again.

metrique-aggregation/src/sink/worker.rs
```diff
@@ -147,4 +167,247 @@ mod tests {
...
+    fn concurrent_drops_exit_cleanly_and_flush_once_determinism() {
+        shuttle::check_uncontrolled_nondeterminism(
+            concurrent_drops_exit_cleanly_and_flush_once,
+            2_000,
+        );
+    }
+
+    /// The shared mpsc channel preserves each sender's own order, so this thread's
+    /// entries must be merged by the time its own `flush()` resolves.
+    fn flush_resolves_after_own_prior_sends() {
```
tiny-ac: Added because `concurrent_sends_all_merged_before_flush_returns` joins its sender threads *before* calling `flush()`, which serializes away the interesting race — whether `flush()` can resolve while a `send()` from another thread is still genuinely in flight, unjoined.

Mutation-tested for exclusivity, no exclusive bug found (surfaced a gap instead).
`WorkerSink`'s receiver loop:
```rust
Ok(QueueMessage::Flush(sender)) => {
    let _ = sender.send(());   // WRONG: replies before the sink is actually flushed
    inner.flush();
    last_flush = Instant::now();
}
```

`cargo test -p metrique-aggregation --no-fail-fast` (~65 tests, lib + all integration binaries + doctests) plus `RUSTFLAGS="--cfg shuttle" cargo test -p metrique-aggregation --lib --features _shuttle -- shuttle`

(6 tests) → 0 failures anywhere. Entries are merged in the `Entry` match arm, unconditionally, strictly before a later-sent `Flush` message (same sender, same channel) is ever processed.

metrique-util/src/pending_sink.rs
```diff
@@ -333,3 +478,190 @@ mod tests {
...
+        }
+        fn flush_async(&self) -> FlushWait {
+            FlushWait::ready()
+        }
+    }
+
+    /// Entries appended concurrently with `resolve()` must never be lost, no
+    /// matter which side of the race each append lands on (buffered then
+    /// drained, or forwarded directly).
+    fn concurrent_appends_racing_resolve_lose_nothing() {
```
tiny-ac:
```rust
// The regression (shipped briefly during the OnceSlot rewrite):
let entry = forward_or_return(&self.0.sink, entry);
self.0.buffering.fetch_sub(1, Ordering::AcqRel);   // decrement moved here...
if let Some(entry) = entry {
    self.0.buffer.force_push(entry);               // ...ahead of the push
}
```
`RUSTFLAGS="--cfg shuttle --cfg tokio_unstable" cargo test -p metrique-util --lib --features _shuttle -- concurrent_appends_racing_resolve_lose_nothing`
→ both `_pct` and `_determinism` fail immediately:
```
assertion `left == right` failed
  left: 3
 right: 4
```
2 of 4 concurrently-appended entries silently dropped. Fixed by restoring push-before-decrement; existing test passes again.

`append()` also relies on incrementing `buffering` *before* the re-check that follows it (a re-check needed because the sink could have been set between the first check and the increment):
```rust
// was:
self.0.buffering.fetch_add(1, Ordering::AcqRel);
let entry = forward_or_return(&self.0.sink, entry);
// reordered to:
let entry = forward_or_return(&self.0.sink, entry);
self.0.buffering.fetch_add(1, Ordering::AcqRel);
```
This reopens a different window: between the re-check and the increment, `resolve()` can see `buffering == 0`, drain, and finish, before this `append()` call increments and force-pushes its own entry into a buffer nobody will ever drain again. Same failure signature, both variants:
```
assertion `left == right` failed
  left: 3
 right: 4
```

metrique-util/src/pending_sink.rs
```diff
@@ -333,3 +478,190 @@ mod tests {
...
+    /// The resolver can be dropped without ever calling `resolve()`
+    /// (documented behavior: cancels the sink, discarding buffered entries)
+    /// concurrently with in-flight `append()` calls. Unlike the resolve()
+    /// race above, there's no "must not lose entries" guarantee here --
+    /// discarding is the documented outcome -- so the invariant this checks
+    /// is narrower but still real: this race must never panic or hang, for
+    /// every interleaving shuttle explores, regardless of whether each
+    /// append's `cancelled` check lands before or after the drop takes
+    /// effect.
+    fn concurrent_append_racing_resolver_drop() {
```
tiny-ac: Dropping the resolver without calling `resolve()` concurrently with in-flight appends must never panic or hang. Not independently mutation-tested in this pass beyond confirming it stays green through the two `concurrent_appends_racing_resolve_lose_nothing` mutations above and their reverts (full 6-test suite reruns).

metrique-util/src/pending_sink.rs
```diff
@@ -333,3 +478,190 @@ mod tests {
...
+
+    /// `flush_async()` can be called concurrently with `resolve()` itself,
+    /// not just after it like the test above does with `append()`. Before
+    /// resolution it's documented to be a no-op (see the real-thread test
+    /// `flush_is_noop_while_pending`); this checks that calling it
+    /// concurrently with `resolve()` never panics or hangs whichever side of
+    /// the race it lands on, and that doing so doesn't interfere with the
+    /// core no-entry-loss guarantee entries appended concurrently still
+    /// depend on.
+    fn concurrent_flush_racing_resolve_lose_nothing() {
```
tiny-ac: `flush_async()` racing `resolve()` itself must not interfere with the no-entry-loss guarantee. Same caveat as the resolver-drop test above — not independently mutation-tested in this pass, confirmed to stay green
through the other two mutations and their reverts.

metrique/src/keep_alive.rs
```diff
@@ -266,4 +280,150 @@ mod test {
...
+    fn concurrent_drop_alls_race_releases_value_exactly_once() {
```
tiny-ac: Hypothetical bugs between two racing `DropAll`s (splitting a `Mutex`-guarded check-then-take into two separate lock acquisitions) are never hit by real-thread testing.

metrique-writer-core/src/global.rs
```diff
@@ -1378,6 +1395,102 @@ mod shutdown_registry_tests {
...
+        shuttle::check_pct(concurrent_register_and_drop, 5_000, 2);
+    }
+
+    #[test]
+    #[ignore = "reproduces a real race in AttachHandle::drop"]
+    fn concurrent_register_and_drop_determinism() {
+        shuttle::check_uncontrolled_nondeterminism(concurrent_register_and_drop, 5_000);
+    }
+
+    fn concurrent_register_and_forget() {
```
tiny-ac: Mutation-tested, exclusive. First attempt, a "helpful" but wrong version that runs pending shutdown fns instead of discarding them:
```rust
pub fn forget(mut self) {
    if let Some(arc) = self.shutdown_registry.take() {
        if let Ok(registry) = Arc::try_unwrap(arc) { registry.drain_and_run(); }
    }
}
```
`cargo test -p metrique-writer-core --lib` → caught by 2 pre-existing sequential tests (`forget_prevents_shutdown_fns_from_running`, `forget_keeps_sink_attached`) — not a useful demonstration of this test's own value. Second attempt, copy-pasting the `drop()` bug's exact shape into `forget()`:
```rust
pub fn forget(mut self) {
    if let Some(arc) = self.shutdown_registry.take() {
        match Arc::try_unwrap(arc) {
            Ok(_registry) => {}
            Err(_) => unreachable!("ShutdownRegistry should have no other strong references"),
        }
    }
}
```

Full sweep: `cargo test -p metrique-writer-core --lib` (57 tests) → 0 failed; `cargo test -p metrique-writer-core --doc` (25 tests) → same 3 pre-existing, unrelated failures as on clean code;

`RUSTFLAGS="--cfg shuttle" cargo test -p metrique-writer-core --lib --features _shuttle -- shuttle`
→ `concurrent_register_and_forget{,_determinism}` fail (panicked on the injected `unreachable!()`), nothing else.

Reverted; `diff` clean; 57/57 + 2/2 shuttle pass again.

metrique-writer-core/src/global.rs
```diff
@@ -1378,6 +1395,102 @@ mod shutdown_registry_tests {
...
+// hitting an `unreachable!()`). Left in place so re-running it after a real
+// fix is a single command.
+#[cfg(all(test, shuttle, feature = "_shuttle"))]
+mod shuttle_tests {
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    use super::{AttachHandle, ShutdownFn};
+
+    fn concurrent_register_and_drop() {
```
tiny-ac: Kept `#[ignore]`d specifically because it reproduces the known, open `AttachHandle::drop` race

metrique-writer/src/sink/background.rs
```diff
@@ -1465,3 +1602,270 @@ mod tests {
...
+    fn overflow_accounting_determinism() {
+        shuttle::check_uncontrolled_nondeterminism(overflow_accounting, 2_000);
+    }
+
+    /// Same conservation invariant as `overflow_accounting`, but pushed from
+    /// several concurrent threads instead of a single-threaded loop, so the
+    /// eviction race in `force_push` (checking `len() >= capacity` then
+    /// evicting then pushing) is actually contended by multiple producers,
+    /// not just raced against the receiver's pops.
+    fn overflow_accounting_concurrent_producers() {
```
tiny-ac: The existing overflow test only pushes from a single thread. This checks the same conservation invariant (`overflows + observed == pushed`), but from two threads pushing concurrently into a capacity-2 queue.

Mutation-tested, no exclusive bug found. `force_push`'s shuttle shim (`metrique-writer/src/sink/shuttle_primitives.rs`):
```rust
// Attempt A: TOCTOU split (worked for pending_sink.rs and keep_alive.rs, doesn't here)
pub(crate) fn force_push(&self, value: T) -> Option<T> {
    let is_full = self.inner.lock().unwrap().len() >= self.capacity;
    let mut queue = self.inner.lock().unwrap();
    let evicted = if is_full { queue.pop_front() } else { None };
    queue.push_back(value);
    evicted
}
```
`RUSTFLAGS="--cfg shuttle" cargo test -p metrique-writer --lib --features _shuttle -- overflow_accounting`
→ all 4 tests pass. A stale `is_full` read lets the queue exceed `capacity`, but every push/eviction decision still executes exactly once, so conservation survives even though the capacity bound doesn't.
```rust
// Attempt B: baseline sanity check — evict physically, never report it
pub(crate) fn force_push(&self, value: T) -> Option<T> {
    let mut queue = self.inner.lock().unwrap();
    if queue.len() >= self.capacity { queue.pop_front(); }
    queue.push_back(value);
    None
}
```
Same command → all 4 tests fail (`left: 4/2/4/7, right: 6/8/6/8`), confirming both variants genuinely check conservation, just not exclusively from each other. Reverted both attempts; `diff` clean; suite passes again. Kept for structural coverage of a scenario the single-producer sibling can't construct, with this caveat on record.

metrique-aggregation/src/sink/worker.rs
```diff
@@ -147,4 +167,266 @@ mod tests {
             "worker should flush exactly once before exiting on disconnect",
         );
     }
+
+    #[tokio::test]
+    async fn flush_does_not_resolve_before_inner_flush_runs() {
```
tiny-ac, line 180: Added to close the gap found while checking what kind of issues `flush_resolves_after_own_prior_sends` could catch. 

First attempt was a Shuttle test using the crate's existing `CollectingSink`/`block_on` shuttle-test infra, `check_pct` at 2,000 iterations → passes, does not catch the bug. `flush()`'s reply channel is a plain `tokio::sync::oneshot` There is no shuttle-instrumented operation at all between `inner.flush()` and the reply send, so Shuttle's scheduler has no yield point to hook a context switch into. Not a probability issue.

Real-thread version instead:
```rust
#[tokio::test]
async fn flush_does_not_resolve_before_inner_flush_runs() {
    for _ in 0..N {
        let flushes = Arc::new(AtomicUsize::new(0));
        let sink = WorkerSink::<(), _>::new(CountingSink { flushes: flushes.clone() }, Duration::from_secs(60));
        sink.flush().await;
        assert_eq!(flushes.load(Ordering::SeqCst), 1, "flush() resolved before inner.flush() actually ran");
    }
}
```
At `N = 1_000` (this codebase's usual stress-test convention): `cargo test -p metrique-aggregation --lib --release ...` → caught the bug in only 1 of 3 runs; plain `cargo test ...` (debug) → 0 of 5 runs —
unreliable, would have been worse than no test if committed at this count.

At `N = 20_000`: 5/5 catches against the mutation (debug and release), 5/5 clean passes against the fix, ~0.4s runtime. Kept at 20,000. Full sweep at 20,000 under the mutation (`cargo test -p metrique-aggregation --no-fail-fast`): exactly this one test fails, nothing else in the crate affected (one flake
seen once, `test_split_sink`, reproduced-away on 5/5 reruns — unrelated pre-existing noise, not caused by this mutation). Shuttle suite (6 tests): unaffected. Reverted; `diff` clean; suite passes again.

metrique-aggregation/src/sink/worker.rs
```diff
@@ -3,14 +3,34 @@
...
+// Cfg-gated concurrency primitives (std vs. shuttle). Gated on both
+// `cfg(shuttle)` and `feature = "_shuttle"` so it also reaches
+// builds of this crate that don't have `_shuttle` enabled (e.g. as a
+// dev-dependency with different requested features) and therefore don't
+// have the optional `shuttle` crate linked at all.
+//
+// `RecvTimeoutError` needs no swap -- shuttle re-exports std's type
+// unchanged. Its `recv_timeout` never actually times out though,
+// that's a known gap documented in Shuttle itself:
+// https://github.com/awslabs/shuttle/blob/c8a46d3965048df3207ec920dae066bc9c4d9d89/shuttle-std/src/sync/mpsc.rs#L433
```
rcoh: yes this correct. If we rely on `recv_timeout` in our code, we should open an issue to do something similar to what we do in dial9 to do our own shuttle primitive that sometimes timeouts w/ some probability

metrique-util/src/pending_sink.rs
```diff
@@ -5,54 +5,186 @@
...
+// protocol but not pure Acquire/Release memory-ordering bugs.
+#[cfg(all(shuttle, feature = "_shuttle"))]
+use shuttle::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+#[cfg(not(all(shuttle, feature = "_shuttle")))]
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+
+#[cfg(not(all(shuttle, feature = "_shuttle")))]
+use crossbeam_queue::ArrayQueue;
+#[cfg(all(shuttle, feature = "_shuttle"))]
+use shuttle_impl::ArrayQueue;
```
rcoh: Can these use the `shuttle_primitives` you already defined in `metrique-writer`?

metrique/src/keep_alive.rs
rcoh: the advantage of shuttle is that when it _can_ reach something, because shuttle is deterministic, you can reproduce it directly.

metrique-writer/src/sink/background.rs
```diff
@@ -1038,6 +1139,42 @@ mod tests {
         }
     }
 
+    /// Reproduces KNOWN BUG, see issue https://github.com/awslabs/metrique/issues/341
+    #[cfg(not(all(shuttle, feature = "_shuttle")))]
+    #[test]
+    #[ignore = "reproduces a real bug: the background thread never auto-shuts-down \
```
rcoh: change the `ignore`'s to `should_fail` so we can ensure they are properly fixed when the bug is fixed.

scripts/shuttle-coverage.sh
```diff
@@ -0,0 +1,20 @@
...
+RUSTFLAGS="--cfg shuttle" \
+  cargo llvm-cov --lib --features _shuttle --html -p metrique-writer-core -- shuttle "$@"
+
+RUSTFLAGS="--cfg shuttle" \
+  cargo llvm-cov --lib --features _shuttle --html -p metrique-writer -- shuttle "$@"
+
+RUSTFLAGS="--cfg shuttle" \
+  cargo llvm-cov --lib --features _shuttle --html -p metrique-aggregation -- shuttle "$@"
+
+RUSTFLAGS="--cfg shuttle --cfg tokio_unstable" \
```
rcoh: these will each overwrite the previous one. if you want this to actually work, we need to generate the report separately

metrique/src/keep_alive.rs
```diff
@@ -266,4 +280,150 @@ mod test {
...
+        );
+    }
+
+    #[test]
+    fn concurrent_drop_alls_race_releases_value_exactly_once_determinism() {
+        shuttle::check_uncontrolled_nondeterminism(
+            concurrent_drop_alls_race_releases_value_exactly_once,
+            2_000,
+        );
+    }
```
rcoh: I wonder if we should make a macro that creates two tests; one to run the actual test and one to check uncontrolled nondeterminism

the macro could also read an environment variable for the number of interleavings to test from env var

rcoh (commented):
overall LGTM!

metrique-util/src/pending_sink.rs
```diff
@@ -5,54 +5,134 @@
...
+            None => Some(entry),
+        });
+        if let Some(entry) = leftover {
+            // force_push before decrementing: resolve()'s spin-wait treats
+            // buffering == 0 as a promise that nothing can still write to the
+            // buffer, so the push must be visible before the count drops.
+            self.0.buffer.force_push(entry);
+            self.0.buffering.fetch_sub(1, Ordering::AcqRel);
         }
     }
```
tiny-ac, line 129: 1. Sink present on first check → append, return. Unchanged
2. Cancelled → drop entry. Unchanged
3. Still absent on second check → `force_push` **then** decrement
   `buffering`. Unchanged
4. Sink becomes present between the two checks → decrement **then**
   `sink.append(entry)`

metrique-writer/src/sink/background.rs
tiny-ac: Done

scripts/shuttle-coverage.sh
tiny-ac: Fixed

metrique-aggregation/src/sink/worker.rs
tiny-ac: Now recv_timeout has an 80% chance of a non-blocking try_recv() (returning Timeout if empty) and a 20% chance of a real blocking recv(). It's in shuttle_test_support.rs, shared so any crate needing it pulls from there instead of reimplementing it

metrique-util/src/pending_sink.rs
tiny-ac: I centralized the shuttle-side shims (ArrayQueue, Parker/Unparker, OnceSlot, GuardArc/GuardWeak,...) in one place. They couldn't go in metrique-shuttle-test (it's proc-macro = true). Also, these shims are called from non-test code.

I put them in metrique-writer-core::shuttle_test_support instead, since it's already a real dependency of every crate that needs them (no new dependency edge to introduce).

metrique-writer's shuttle_primitives.rs is background.rs's own local re-export of the same shared shims, and it's private to that crate

metrique/src/keep_alive.rs
tiny-ac: I added a new non-published crate for the macro (it doesn't depend on env vars in its current state, it'd be a small change)

metrique-shuttle-test/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+/// Generates the `<name>_pct` / `<name>_determinism` shuttle test pair for
+/// the function it's attached to, calling `shuttle::check_pct` and
+/// `shuttle::check_uncontrolled_nondeterminism` with the same iteration
+/// count -- the pattern every shuttle test in this workspace follows.
+///
+/// ```ignore
+/// #[shuttle_test(2_000, 3)]
```
rcoh: please name these fields in the macro so it is clearer

`shuttle_test(numer_iters = ..., `

rcoh: does this need to be a proc_macro in its own crate? could this be a macro_rules macro?

rcoh (approved):

PR #339, opened 2026-07-13
Fix cross-crate E0119 for entry-mode #[aggregate]

rcoh:
## Problem

Two entry-mode `#[aggregate]` structs in different crates in the same build graph produce a coherence error, even when neither crate names the other's type:

```
error[E0119]: conflicting implementations of trait `Merge`
  for type `<DepMetrics as CloseValue>::Closed`
```

Reported by a user with `DepM` in a library crate and `CoreM` in a consumer crate that only used an unrelated type from the library. Switching the library struct to `#[aggregate(direct)]` worked around it.

## Root cause

Entry mode emitted the `Merge` (and, with `#[aggregate(ref)]`, `MergeRef`) impl against the associated-type **projection** `<T as CloseValue>::Closed`. Coherence runs in intercrate mode (`TypingMode::Coherence`) where it does **not** normalize projections and must stay complete — so it treats `<DepM>::Closed` and `<CoreM>::Closed` as rigid types that could unify, and rejects the second impl as a potential overlap. This is expected rustc behavior, not a bug (it only reproduces cross-crate; within one crate both `CloseValue` impls are visible and coherence can prove disjointness).

## Fix

Emit the `Merge`/`MergeRef` impls against the **concrete** generated entry struct (e.g. `ApiCallEntry`) instead of the projection. Trait *selection* (unlike coherence) normalizes the projection to that concrete type, so every projection-based call site (`AggregateStrategy::Source`, `Aggregate<T>`, `close_and_merge`) keeps working unchanged.


jlizen (approved):
Fix looks fine. Only nit is, it would be good to also probe `#[aggregate(ref)]` in the cross crate coherence test.

PR #346, opened 2026-07-26
fix: forward ValueWriter::values() through value wrapper writers

jlizen:
(Noticed this bug while working on https://github.com/dial9-rs/dial9/pull/723)

📬 *Issue #, if available:*

✍️ *Description of changes:*

## Problem

`ValueWriter::values()` has a default implementation that comma-joins each element's string representation into a single `string()` call. Every internal wrapper `ValueWriter` (the write-path adapters behind `ForceFlag`, `WithDimensions`, `WithUnit`, and `WithGlobalDimensions`) forwarded `string()`, `metric()`, and `error()` but not `values()`, so any list value passing through them fell back to the comma-joining default. Formats with native array support (EMF, JSON) never saw the elements: a flagged `Vec<String>` rendered as `"a,b"` instead of `["a", "b"]`, and per-element metric attributes (merged flags, added dimensions, unit conversion) were silently dropped.

### Summary

Each wrapper now overrides `values()` and forwards to the underlying writer, wrapping every element in the same adapter the wrapper already uses for scalar values, so element-level `metric()` calls keep their flag merging, dimension chaining, or unit conversion. Elements are collected into a `SmallVec` (inline capacity 8) to satisfy the by-reference iterator contract without heap allocation for typical list sizes, matching the existing pattern in `entry/boxed.rs`.


metrique-writer-core/src/value/force.rs
```diff
@@ -148,6 +149,13 @@ impl<T: Value, FLAGS: FlagConstructor> Value for ForceFlag<T, FLAGS> {
             fn error(self, error: ValidationError) {
                 self.0.error(error)
             }
+
+            fn values<'a, V: Value + 'a>(self, values: impl IntoIterator<Item = &'a V>) {
+                // Wrap each element so `metric()` calls still merge the flag.
+                let wrapped: SmallVec<[ForceFlag<&'a V, FLAGS>; 8]> =
```
rcoh: worth extracting a const or noting how 8 was chosen?

metrique-writer-core/src/value/mod.rs
rcoh, line 122: can we extend our cfg-gate tests to catch a missing `values` impl?

rcoh (approved):

metrique-writer-core/src/value/mod.rs
jlizen: I thought about it but there were some that were intentionally defaulted. I guess I can make them explicit using the default, not sure why I didn't before.

rcoh: or an attribute that basically just allows the test to pass

jlizen: Nah, explicit is good

PR #347, opened 2026-07-26
fix: Yield descriptor segments in Entry::write order for flattened entries

### LINKED ISSUE #345: bug: Entry::write value order does not match descriptors() segment order for flattened entries
(Discovered while working on: https://github.com/dial9-rs/dial9/pull/723)

## Problem

`docs/entry-descriptors.md` ("Entry::write order contract") says the macro emits one `EntryWriter::value` callback per `FieldDescriptor`, in the same order as the fields listed in each descriptor returned by `descriptors()`, with each descriptor covering a contiguous segment of the write output.

That contract breaks whenever a flatten site precedes sibling fields: the generated `Entry::write` visits child fields where they are declared, but `descriptors()` puts the whole parent segment ahead of every child segment.

## Repro

```rust
#[metrics]
struct Child { a: u64 }

#[metrics]
struct Parent {
    #[metrics(flatten)]
    child: Child,
    b: u64,
}
```

Write order: `a`, `b`. Descriptor segment order: `[Parent: b]`, `[Child: a]`.

## Impact

Descriptor-aware sinks cannot consume positionally, even though the design doc presents that as the intended pattern.

## Expected

Make `descriptors()` yields segments in write order (child segments interleaved at declaration position)

jlizen:
📬 *Issue #, if available:*

Closes https://github.com/awslabs/metrique/issues/345 (discovered while working on https://github.com/dial9-rs/dial9/pull/723)

✍️ *Description of changes:*

`docs/entry-descriptors.md` promises that walking `descriptors()` segments in sequence yields fields in exactly the order `Entry::write` emits `EntryWriter::value` callbacks, with each segment covering a contiguous slice of the write output. That contract broke whenever a flatten site preceded sibling fields: `Entry::write` visits child fields at their declaration position, but `descriptors()` put the whole parent segment ahead of every child segment. Descriptor-aware sinks could not consume positionally, even though the design doc presents that as the intended pattern.

### Summary

The macro now splits an entry's own (non-flatten) fields into contiguous runs at flatten boundaries and emits one descriptor segment per run, interleaved with the flattened children's segments in declaration order. For `Parent { a, flatten child, b }`, `descriptors()` yields `[Parent: a]`, `[Child: ...]`, `[Parent: b]` instead of the old `[Parent: a, b]`, `[Child: ...]`.

The first segment always belongs to the parent, even when no own fields precede the first flatten site (it then lists zero fields and consumes zero values). This keeps the parent's canonical name first and gives the timestamp descriptor a stable home: the timestamp flows through `EntryWriter::timestamp` rather than `value()`, so it has no position in the field stream. Later runs are emitted only when non-empty, so entries whose flattens all trail their own fields (the common layout) generate identical output to before. A cfg-disabled flatten site still splits the surrounding runs; two adjacent parent segments still satisfy the contract, which sidesteps needing cfg-conditional statics.

Enum variants get the same treatment, with the tag field anchoring the first run. As part of unifying the flatten chain generation between structs and enums, enums also pick up a fix for a pre-existing gap: flatten-site `default_flags` on enum variant fields were applied on the write path but never surfaced in the child's descriptor (`with_extra_flags` was only wired up for structs).

Review of the write path surfaced three more descriptor/write-path mismatches in the same class, fixed here as well:

- cfg-gated plain fields were listed in the descriptor unconditionally while `write()` only emits them when the cfg is active. They now get their own cfg-gated segment, so the descriptor tracks whatever the write path compiles to.
- `DescriptorRef::with_extra_flags` replaced rather than accumulated, so nested flatten sites that both carry `default_flags` dropped the inner site's flags from grandchild segments, diverging from the write path's nested `ForceFlagEntryWriter`s. Extra flag slices now stack, and `DescriptorId` incorporates them.
- Enum tag descriptor names were style-invariant while the write path inflects the tag through the propagated name style; a subfield enum without its own `rename_all` flattened into a `PascalCase` parent reported `operation_type` in the descriptor but wrote `OperationType`. The descriptor now carries per-style tag names.

### DIFF
--- a/metrique-macro/src/entry_impl/enum_impl.rs
+++ b/metrique-macro/src/entry_impl/enum_impl.rs
@@ -93,10 +93,7 @@ fn generate_write_arms(
     variants: &[MetricsVariant],
     root_attrs: &RootAttributes,
 ) -> Vec<Ts2> {
-    let tag_name = root_attrs
-        .tag
-        .as_ref()
-        .map(|tag| tag.field_name(root_attrs));
+    let tag = root_attrs.tag.as_ref();
     let writer_ident = mixed_site_writer();
 
     variants
@@ -104,11 +101,11 @@ fn generate_write_arms(
         .map(|variant| {
             let variant_ident = &variant.ident;
 
-            let tag_write = tag_name.as_ref().map(|tag_name| {
+            let tag_write = tag.map(|tag| {
                 let (extra, name) = make_inflect(
                     &make_ns(root_attrs.rename_all, variant.ident.span()),
                     variant.ident.span(),
-                    |style| style.apply(tag_name),
+                    |style| tag.styled_name(root_attrs, style),
                 );
                 let value = crate::inflect::inflect_no_prefix(root_attrs, variant);
                 quote! {
@@ -250,21 +247,18 @@ fn generate_sample_group_arms(
     root_attrs: &RootAttributes,
     iter_enum_name: &Ident,
 ) -> Vec<Ts2> {
-    let tag_name = root_attrs
-        .tag
-        .as_ref()
-        .map(|tag| tag.field_name(root_attrs));
+    let tag = root_attrs.tag.as_ref();
     let include_tag_in_sample_group = root_attrs.tag.as_ref().is_some_and(|t| t.sample_group());
 
     variants.iter().enumerate().map(|(idx, variant)| {
         let variant_ident = &variant.ident;
         let iter_variant_name = quote::format_ident!("V{}", idx);
 
-        let tag_sample_group = if let Some(tag_name) = tag_name.as_ref().filter(|_| include_tag_in_sample_group) {
+        let tag_sample_group = if let Some(tag) = tag.filter(|_| include_tag_in_sample_group) {
             let (extra, name) = make_inflect(
                 &make_ns(root_attrs.rename_all, variant.ident.span()),
                 variant.ident.span(),
-                |style| style.apply(tag_name),
+                |style| tag.styled_name(root_attrs, style),
             );
             let value = crate::inflect::inflect_no_prefix(root_attrs, variant);
             Some(quote! {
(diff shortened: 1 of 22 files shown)


metrique-macro/src/entry_impl/enum_impl.rs
```diff
@@ -381,14 +386,18 @@ fn generate_enum_descriptor(
...
+            // Run 0 starts with the tag field, if present: the tag value is
+            // written before any variant fields.
+            let mut runs: Vec<Vec<DescriptorFieldMeta>> = vec![Vec::new()];
             if let Some(tag) = &root_attrs.tag {
+                // The write path inflects the tag through the propagated style.
+                let tag_name = tag.field_name(root_attrs);
                 let names: [String; metrique_core::Styles::COUNT] =
-                    std::array::from_fn(|_| tag.field_name(root_attrs));
-                v_field_metas.push(DescriptorFieldMeta {
+                    std::array::from_fn(|i| styles[i].apply(&tag_name));
```
rcoh: the change from `tag.field_name` to `.apply` causes inflection to be applied to name_exact.

failing test:
```rust
  #[metrics(subfield, tag(name_exact = "raw_name"))]
  pub enum ExactTagStyleEnum {
      DoThing { some_field: u64 },
  }

  struct ExactTagStyleParent {
      #[metrics(flatten)]
      op: ExactTagStyleEnum,
  }

  #[test]
  fn enum_exact_tag_name_ignores_propagated_style() {
      let m = ExactTagStyleParent {
          op: ExactTagStyleEnum::DoThing { some_field: 1 },
      };
      let closed = metrique::CloseValue::close(m);
      let entry = metrique::RootEntry::new(closed);

      let expected = vec!["raw_name".to_owned(), "SomeField".to_owned()];
      assert_eq!(
          (write_order(&entry), descriptor_order(&entry)),
          (expected.clone(), expected)
      );
  }
```

jlizen: Ah, `main` already had a write-path bug, and this PR made the descriptor handling consistent with it. I'll fix both.

rcoh (approved):

PR #356, opened 2026-08-03
fix(metrique-writer): auto-shutdown BackgroundQueue when all clones drop

### LINKED ISSUE #341: `BackgroundQueue` doesn't auto-shut-down when all queue handles are dropped
### Summary

`BackgroundQueue` is normally long-lived: `BackgroundQueue::new()` (or `BackgroundQueueBuilder::build`) is called once, typically at service startup, and the returned `(queue, handle)` pair is held for the service's whole lifetime — `queue` gets reused across many requests/units of work, and `handle` (a `BackgroundQueueJoinHandle`) is dropped (or `shut_down()`'d) exactly once, when the service itself shuts down.

[forget()](https://github.com/awslabs/metrique/blob/432c21e9f40483677394afa78c9a2ebafd5465c3/metrique-writer/src/sink/background.rs#L522) gives up the ability to explicitly `shut_down()` later, but documents that the background thread should still stop on its own once it's no longer needed.

The thread is supposed to keep running only *until no `BackgroundQueue`s exist*. It keeps running indefinitely instead (parking, waking on its flush interval, flushing an empty queue) until the process exits.

[/metrique-writer/src/sink/background.rs](https://github.com/awslabs/metrique/blob/432c21e9f40483677394afa78c9a2ebafd5465c3/metrique-writer/src/sink/background.rs#L678-L751)

```rust
fn run(mut self, flush_queue_receiver: mpsc::Receiver<FlushSignal>) {
    ...
    let inner = self.inner.clone(); // kept alive for the whole function

    loop {
        ...
        loop {
            ...
            waker_tracker.handle_waiting_wakers(
                || inner.queue.capacity(),   // borrows the local `inner`
                || self.flush_stream(),      // needs `&mut self` as a whole
                status,
                entry_count,
            );
            ...
        }
        ...
        if Arc::get_mut(&mut self.inner).is_some() {
            tracing::info!("no appenders left, shutting down background metrics queue");
            return self.shut_down();
        }
    }
}
```

`Arc::get_mut` requires the strong count across *all* clones of `self.inner` (not just `self.inner` itself) to be exactly 1. With `inner` and `self.inner` both alive for the whole function, the count never drops below 2 — so this check can never succeed, no matter how many external `BackgroundQueue` clones are dropped. This seems to have been present since the initial commit.

### Impact
No data lost or panic — it's a dead cleanup path that silently breaks the contract documented on `forget()`.

The same broken check applies just the same if a caller holds the `BackgroundQueueJoinHandle` open without calling `forget()` or `drop()` at all (e.g. as a struct field, meaning to `shut_down()` it explicitly later) — the code doesn't distinguish between the two.

### Reproducible test case
Full integration-level reproduction, as a test in `metrique-writer/src/sink/background.rs`:

```rust
#[test]
fn auto_shutdown_when_no_appenders_left() {
    let output: Arc<Mutex<TestStream>> = Default::default();
    let (queue, handle) = BackgroundQueueBuilder::new()
        .capacity(100)
        .flush_interval(Duration::from_millis(5))
        .build::<TestEntry>(Arc::clone(&output));

    queue.append(TestEntry(0));
    drop(queue); // all BackgroundQueue clones gone; `handle` stays alive

    let start = Instant::now();
    let auto_shut_down = loop {
        if handle.handle.as_ref().unwrap().is_finished() {
            break true;
        }
        if start.elapsed() > Duration::from_secs(3) {
            break false;
        }
        std::thread::sleep(Duration::from_millis(10));
    };

    handle.shut_down(); // clean up regardless of the outcome above
    assert!(auto_shut_down, "background thread never auto-shut-down after all queues were dropped");
}
```

(`handle.handle` is a private field; run this as a test inside the crate's own `mod tests`, or add a `#[doc(hidden)] pub fn is_running(&self) -> bool` accessor if reproducing from outside the crate.) This fails every time, deterministically, with `background thread never auto-shut-down after all queues were dropped` — no scheduling luck involved, no retries needed.

**Minimal repro** — the bug's mechanism stripped down to nothing else, demonstrating the exact same result on every run:

```rust
use std::sync::Arc;

fn main() {
    let outer = Arc::new(1u64);
    let mut self_inner = outer.clone(); // stands in for `self.inner`
    drop(outer);                        // stands in for dropping all `BackgroundQueue` clones

    let inner = self_inner.clone();     // stands in for `run()`'s local `inner`
    assert!(Arc::get_mut(&mut self_inner).is_none()); // always None -- `inner` is still alive
}
```

**Public-API-only demonstration** 
```rust
use metrique_writer::{Entry, EntrySink};
use metrique_writer_core::{EntryIoStream, IoStreamError};
use std::sync::mpsc;
use std::time::{Duration, Instant};

#[derive(Entry)]
struct DemoEntry {
    value: u32,
}

struct CanaryStream {
    name: &'static str,
    dropped_tx: mpsc::Sender<&'static str>,
}

impl EntryIoStream for CanaryStream {
    fn next(&mut self, _entry: &impl metrique_writer_core::Entry) -> Result<(), IoStreamError> {
        Ok(())
    }
    fn flush(&mut self) -> std::io::Result<()> {
        Ok(())
    }
}

impl Drop for CanaryStream {
    fn drop(&mut self) {
        // Only fires inside `Receiver::shut_down()`.
        let _ = self.dropped_tx.send(self.name);
    }
}

fn main() {
    println!("=== Scenario A: working path (explicit shut_down()) ===");
    {
        let (tx, rx) = mpsc::channel();
        let (queue, handle) = metrique_writer::sink::BackgroundQueueBuilder::new()
            .capacity(10)
            .flush_interval(Duration::from_millis(50))
            .build::<DemoEntry>(CanaryStream { name: "scenario-a", dropped_tx: tx });

        queue.append(DemoEntry { value: 1 });
        drop(queue); // every BackgroundQueue clone gone

        let start = Instant::now();
        handle.shut_down(); // explicit, blocking shutdown
        println!(
            "handle.shut_down() returned after {:?} -- thread shut down as expected",
            start.elapsed()
        );
        let shut_down = rx.recv_timeout(Duration::from_millis(100)).is_ok();
        println!("stream was dropped (thread reached shut_down()): {shut_down}");
    }

    println!("\n=== Scenario B: broken path (forget(), then drop every queue clone) ===");
    {
        let (tx, rx) = mpsc::channel();
        let (queue, handle) = metrique_writer::sink::BackgroundQueueBuilder::new()
            .capacity(10)
            .flush_interval(Duration::from_millis(50))
            .build::<DemoEntry>(CanaryStream { name: "scenario-b", dropped_tx: tx });

        queue.append(DemoEntry { value: 1 });
        handle.forget(); // doc comment: thread keeps running "until no BackgroundQueues exist"
        drop(queue); // every BackgroundQueue clone is now gone

        println!(
            "dropped every BackgroundQueue clone after forget(); waiting up to 3s for the \
             thread to notice and shut down on its own..."
        );
        let start = Instant::now();
        let shut_down = rx.recv_timeout(Duration::from_secs(3)).is_ok();
        println!(
            "after {:?}: stream was dropped (thread shut down on its own): {shut_down}",
            start.elapsed()
        );
        if !shut_down {
            println!(
                "BUG CONFIRMED: the background thread never noticed it had no appenders left, \
                 despite forget()'s documented promise -- it's still running, doing nothing, \
                 right now, as this process is about to exit."
            );
        }
    }
}
```

Actual output from running this:

```
=== Scenario A: working path (explicit shut_down()) ===
handle.shut_down() returned after 184.375µs -- thread shut down as expected
stream was dropped (thread reached shut_down()): true

=== Scenario B: broken path (forget(), then drop every queue clone) ===
dropped every BackgroundQueue clone after forget(); waiting up to 3s for the thread to notice and shut down on its own...
after 3.002999709s: stream was dropped (thread shut down on its own): false
BUG CONFIRMED: the background thread never noticed it had no appenders left, despite forget()'s documented promise -- it's still running, doing nothing, right now, as this process is about to exit.
```

Scenario A shuts down in microseconds; Scenario B is still running, unnoticed, a full 3 seconds later — and would keep running indefinitely, since nothing else ever tells it to stop.

### Potential fix

Move `let inner = self.inner.clone();` from once before the outer loop to freshly inside the inner loop, right before the handle_waiting_wakers` call that's the only thing that needs it — so it's dropped at the end of every inner-loop iteration instead of living for the whole function:

```diff
         let mut waker_tracker = WakerTracker::new(flush_queue_receiver);
-        let inner = self.inner.clone();

         loop {
             let next_flush = Instant::now() + self.flush_interval;
             let loop_start: Instant = Instant::now();
             let mut idle_duration = Duration::ZERO;
             loop {
                 let (status, entry_count) = self.drain_until_deadline(next_flush);

+                let inner = self.inner.clone();
                 waker_tracker.handle_waiting_wakers(
                     || inner.queue.capacity(),
                     || self.flush_stream(),
                     status,
                     entry_count,
                 );
```

### Version

- `metrique-writer` 0.1.24 (workspace at `main` commit
  `5d3a198ae24cc1f00d4f94136a65ac939ec1a21e`, "Release: Jul 8th 2026 (#337)").

tiny-ac:
📬 *Issue #, if available:*
closes #341 

✍️ *Description of changes:*

- `Receiver::run`'s `self.inner.clone()` used to live for the whole function, keeping the strong count above 1 forever, so `Arc::get_mut` could never detect "every external `BackgroundQueue` clone has been dropped" and auto-shut-down. Removed the clone entirely (not just relocated) by changing `WakerTracker::handle_waiting_wakers`'s `queue_capacity` param from a closure to a plain `usize`
- Added regression tests

### DIFF
--- a/metrique-writer/src/sink/background.rs
+++ b/metrique-writer/src/sink/background.rs
@@ -676,7 +676,6 @@ impl<S: EntryIoStream, E: Entry> Receiver<S, E> {
         let span = tracing::span!(tracing::Level::TRACE, "metrics background queue", sink=?self.inner.name);
         let _enter = span.enter();
         let mut waker_tracker = WakerTracker::new(flush_queue_receiver);
-        let inner = self.inner.clone();
 
         loop {
             let next_flush = Instant::now() + self.flush_interval;
@@ -684,6 +683,9 @@ impl<S: EntryIoStream, E: Entry> Receiver<S, E> {
             let mut idle_duration = Duration::ZERO;
             loop {
                 let (status, entry_count) = self.drain_until_deadline(next_flush);
+                // Must be dropped before `Arc::get_mut(&mut self.inner)` below to auto-shut-down.
+                // This is a borrow-checker workaround since it's also being borrowed mutably when calling flush_stream().
+                let inner = self.inner.clone();
 
                 waker_tracker.handle_waiting_wakers(
                     || inner.queue.capacity(),
@@ -738,7 +740,6 @@ impl<S: EntryIoStream, E: Entry> Receiver<S, E> {
                 tracing::info!("caught shutdown signal, shutting down background metrics queue");
                 return self.shut_down();
             }
-            // KNOWN BUG (see issue https://github.com/awslabs/metrique/issues/341)
             if Arc::get_mut(&mut self.inner).is_some() {
                 tracing::info!("no appenders left, shutting down background metrics queue");
                 return self.shut_down();
@@ -906,6 +907,8 @@ fn next_loop_action(
 #[cfg(test)]
 #[allow(deprecated)]
 mod tests {
+    #[cfg(not(all(shuttle, feature = "_shuttle")))]
+    use std::sync::Barrier;
     use std::{
         future::Future,
         pin::Pin,
@@ -1138,12 +1141,8 @@ mod tests {
         }
     }
 
-    /// Reproduces KNOWN BUG, see issue https://github.com/awslabs/metrique/issues/341
     #[cfg(not(all(shuttle, feature = "_shuttle")))]
     #[test]
-    #[should_panic(
-        expected = "background thread never auto-shut-down after all queues were dropped"
-    )]
     fn auto_shutdown_when_no_appenders_left() {
         test_all_queues! {
             |builder| builder.capacity(100).flush_interval(Duration::from_millis(5)),
@@ -1170,6 +1169,85 @@ mod tests {
                     auto_shut_down,
                     "background thread never auto-shut-down after all queues were dropped"
                 );
+            }
+        }
+    }
+
+    #[cfg(not(all(shuttle, feature = "_shuttle")))]
+    #[test]
+    fn holding_a_queue_clone_blocks_auto_shutdown() {
+        test_all_queues! {
+            |builder| builder.capacity(100).flush_interval(Duration::from_millis(5)),
+            |_output, queue, handle| {
+                queue.append(TestEntry(0));
+                let kept_alive = queue.clone();
+                drop(queue);
+
+                // give the background thread several flush cycles' worth of time to (incorrectly) shut down
+                std::thread::sleep(Duration::from_millis(200));
+                assert!(
+                    !handle.handle.as_ref().unwrap().is_finished(),
+                    "background thread exited even though a BackgroundQueue clone is still held"
+                );
+
+                drop(kept_alive);
+                handle.shut_down();
+            }
+        }
+    }
+
+    #[cfg(not(all(shuttle, feature = "_shuttle")))]
+    #[test]
+    fn concurrent_clone_and_drop_do_not_race_auto_shutdown() {
+        test_all_queues! {
+            |builder| builder.capacity(100).flush_interval(Duration::from_millis(2)),
+            |_output, queue, handle| {
+                const ITERATIONS: usize = 200;
+                const THREADS: usize = 8;
+
+                queue.append(TestEntry(0));
+
+                for i in 0..ITERATIONS {
+                    // held by this thread for the whole iteration: guarantees the queue
+                    // must NOT auto-shut-down at any point during the race below
+                    let keep_alive = queue.clone();
+                    let barrier = Arc::new(Barrier::new(THREADS + 1));
+
+                    std::thread::scope(|scope| {
+                        for _ in 0..THREADS {
+                            let queue = queue.clone();
+                            let barrier = Arc::clone(&barrier);
+                            scope.spawn(move || {
+                                barrier.wait();
+                                for _ in 0..50 {
+                                    drop(queue.clone());
+                                }
+                            });
+                        }
+                        barrier.wait();
+                    });
+
+                    assert!(
+                        !handle.handle.as_ref().unwrap().is_finished(),
+                        "background thread auto-shut-down while a clone was still held (iteration {i})"
+                    );
+                    drop(keep_alive);
+                }
+
+                // now really drop the last clone and confirm the thread eventually exits
+                drop(queue);
+                let start = Instant::now();
+                loop {
+                    if handle.handle.as_ref().unwrap().is_finished() {
+                        break;
+                    }
+                    assert!(
+                        start.elapsed() < Duration::from_secs(3),
+                        "background thread never auto-shut-down after all queues were dropped"
+                    );
+                    std::thread::sleep(Duration::from_millis(10));
+                }
+                handle.shut_down();
             }
         }
     }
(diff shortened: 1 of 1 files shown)


metrique-writer/src/sink/background.rs
```diff
@@ -676,14 +676,15 @@ impl<S: EntryIoStream, E: Entry> Receiver<S, E> {
...
             let next_flush = Instant::now() + self.flush_interval;
             let loop_start: Instant = Instant::now();
             let mut idle_duration = Duration::ZERO;
             loop {
                 let (status, entry_count) = self.drain_until_deadline(next_flush);
+                // Must be dropped before `Arc::get_mut(&mut self.inner)` below to auto-shut-down.
+                let inner = self.inner.clone();
 
                 waker_tracker.handle_waiting_wakers(
                     || inner.queue.capacity(),
```
rcoh: can this just be `self.inner.capacity()`? Then we don't need to clone self.inner at all?

All the other usages appear to use self.inner, there is no _actual_ need for `self.inner.clone()` as far as I can tell.

tiny-ac: It seems to be a borrow-checker workaround, `self.inner.capacity()` throws this error: 

>   error[E0502]: cannot borrow `self` as mutable because it is also borrowed as immutable
>      --> metrique-writer/src/sink/background.rs:689:21
>       |
>   687 |                 waker_tracker.handle_waiting_wakers(
>       |                               --------------------- immutable borrow later used by call
>   688 |                     || self.inner.queue.capacity(),
>       |                     -- ---------- first borrow occurs due to use of `self.inner` in closure
>       |                     |
>       |                     immutable borrow occurs here
>   689 |                     || self.flush_stream(),
>       |                     ^^ ---- second borrow occurs due to use of `self` in closure
>       |                     |
>       |                     mutable borrow occurs here

rcoh: ah I see. OK. Please explain that in a comment.

metrique-writer/src/sink/background.rs
```diff
@@ -632,7 +632,7 @@ impl WakerTracker {
     // number of pops.
     fn handle_waiting_wakers(
         &mut self,
-        queue_capacity: impl FnOnce() -> usize,
```
rcoh, line 635: hmm, I am guessing this computed it lazily for a reason; you'll note that the queue capacity is computed in a loop.

tiny-ac: It gets the capacity value from [crossbeam's ArrayQueue](https://github.com/crossbeam-rs/crossbeam/blob/v0.8/crossbeam-queue/src/array_queue.rs#L211-L212) (set once when constructed and never mutated afterward) and self.inner seems to remain unchanged, so my understanding is that capacity is the same for as long as the loop runs. I'll look into it further

rcoh (approved):
LGTM!

Issue #159, opened 2025-12-31
Patterns for appending metrics via outer middleware that incorporates inner middleware/handler metrics


jlizen:
## Background

Frequently users are simply creating and flushing metrics entries inside their handler. If middleware is generating metrics, it also tends to be local to the middleware.

But, ideally you would want metrics that gathers together all the metrics prepared by different layers into a single unit of work context without joining on request ID, etc. There even might be cases multiple middleware might want to write to the same metrics struct.

This is complicated by a few things:
- All extensions need to be `Clone` (even if the frameworks themselves don't generally clone extensions and swap out which extensions map is attached to the request)
- The request extensions are not automatically moved to the response extensions
- For axum, the extension extractor implicitly clones the requested type -> this one I think we can ignore and have our own helpers for, someday our own extractor

This issue discusses what patterns we might want to recommend, and some initial metrique-side support that might help.
(thanks to @arielb1 for offline suggestions)

## Vision

It would be great to have a "non-clonable extensions" map in `http::Request` to simplify the story, but probably we need to work around it in the short term.

Our desired state, without changes to `http::Request`, and not considering any sort of metrique-vended middleware, might be something like this:

1. hyper inbound
   a. opens a "parent waits on flush" slotguard-like thing containing an opaque entry struct capturing socket timings, overall timings before/during/after router processing, and other low-level details
   b. holds onto the slotguard and puts the Arc'd parent slot into extension
2. metrics middleware inbound
   a. extracts the hyper-injected slot using .into_inner()
   b. creates request metrics entry with generic information like path, opens timer, includes the hyper-injected slot if Some
   c. open a "super entry" that is append on drop and contains the union of the request metrics entry and (currently unknown) inner entries
   d. opens up some sort of clonable smart pointer that lets it retrieve type erased middleware/handler metric, clones it into extensions
   e. calls inner services
3. inner middlewares/handler
   a. retrieves smart pointer from extensions
   b. opens up its own metric struct, which is is queued for usage in the pointer on drop
4. metrics middleware outbound
   a. further enrich parent metric with results  of inner calls (status code, close timer, etc)
   b. fold loaded middleware/handler entries into a "super" metrics entry
        i. optionally could wait up to X duration for inner middleware/handler to finish dropping their entries, but probably not critical for now
        ii. more likely just discard any pending inner metrics to avoid misbehaving middleware hanging indefinitely
   c. drop the append-on-drop "super entry" (which still waits on hyper to drop its guard when it is done processing)
[...]

PR #357, opened 2026-08-05
RFC: Metrics pools for request-scoped collection

rcoh:
## Summary

- propose `MetricsPool` for dynamically collecting child metrics into a parent entry
- define explicit-handle and poll-scoped current-pool APIs, including prefix, casing, collision, and close semantics
- document the AWS SDK interceptor adaptation, implementation internals, benchmark plan, and changes checklist

This is a proposed primitive to support #159 

[Rendered](https://github.com/awslabs/metrique/blob/metrics-pool-rfc/docs/metrics-pool-rfc.md)

### DIFF
--- a/docs/metrics-pool-rfc.md
+++ b/docs/metrics-pool-rfc.md
@@ -1,0 +1,567 @@
+# RFC: Metrics pools for request-scoped metric collection
+
+> Status: RFC with prototype
+>
+> Applies to: `metrique`, `metrique-util`
+
+For a summarized list of proposed changes, see the
+[Changes checklist](#changes-checklist).
+
+This RFC defines `MetricsPool`, a collector that lets independently-owned
+middleware and libraries contribute fields to one parent metrics entry. A pool
+can be passed explicitly like a regular sink or discovered through a scoped
+current-pool API. The scoped API is intended for synchronous callbacks, such as
+AWS SDK interceptors, that execute while a request future is being polled but
+cannot accept another application parameter.
+
+## Terminology
+
+- **Parent entry**: The metrics struct that owns a flattened `MetricsPool` and
+  is eventually appended to the application's real sink.
+- **Child metric**: A metrics struct appended to a pool and flattened into the
+  parent when the parent closes.
+- **Pool handle**: A cloneable producer, `MetricsPoolHandle`, that appends child
+  metrics without exposing previously appended metrics.
+- **Current pool**: The pool installed while a scoped future is being polled.
+- **Standalone entry**: An entry emitted directly to an application sink when
+  no current pool is available.
+
+## The user experience if this RFC is implemented
+
+This section is written as the public documentation users should be able to
+read once the API is released.
+
+### Capability matrix
+
+Metric contribution has two independent choices: whether the destination is
+strongly or dynamically typed, and how a producer obtains it.
+
+| Destination typing | Explicitly passed | OS thread-local | Async task-local |
+| --- | --- | --- | --- |
+| Strongly typed | Supported today: pass `&mut RequestMetrics` or a typed child guard | Not supported | Not currently supported; reserve for a future typed task-local API |
+| Dynamically typed | This RFC: pass `MetricsPoolHandle` as a regular sink | Deliberately not exposed because async work migrates between threads | This RFC: `MetricsPool::current()` within a poll-scoped future |
+
+Explicit, strongly typed access remains the preferred choice when the producer
+and parent metric can share a concrete type. `MetricsPool` is for integration
+boundaries where that coupling is unavailable. Its current-pool API has
+task-local semantics even though its executor-independent implementation uses
+a thread-local only during each future poll.
+
+### Collect child metrics in a parent entry
+
+Enable the `metrics-pool` feature on `metrique-util`:
+
+```toml
+[dependencies]
+metrique-util = { version = "...", features = ["metrics-pool"] }
+```
+
+Add a flattened `MetricsPool` to the metrics entry that represents the unit of
+work. Normal flatten-site prefixes are supported:
+
+```rust,ignore
+use metrique::unit_of_work::metrics;
+use metrique_util::MetricsPool;
+
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    operation: &'static str,
+
+    #[metrics(flatten, prefix = "child_")]
+    child_metrics: MetricsPool,
+}
+```
+
+Code with direct access to the parent can obtain a cloneable handle for a
+suboperation:
+
+```rust,ignore
+let request = RequestMetrics {
+    operation: "PutObject",
+    child_metrics: MetricsPool::new(),
+};
+
+let suboperation_pool = request
+    .child_metrics
+    .handle()
+    .with_prefix(["sdk", "s3"]);
+
+suboperation_pool.append(SuboperationMetrics {
+    operation: "PutObject",
+    retry_count: 2,
+});
+
+// Closing `request` emits one object containing the parent fields followed by
+// the contributed suboperation fields shown below.
+```
+
+`MetricsPoolHandle` implements the normal entry sink trait. It also has
+`append`, `append_on_drop`, and `append_on_drop_default` convenience methods.
+The handle can be cloned into request extensions, middleware, or library state.
+It cannot inspect or remove entries already contributed by another producer.
+
+When `RequestMetrics` closes, the suboperation fields are part of that same
+entry:
+
+```text
+Operation = PutObject
+ChildSdkS3Operation = PutObject
+ChildSdkS3RetryCount = 2
+```
+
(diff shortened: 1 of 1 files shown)


jlizen:
I haven't yet read this, I will (hopefully today), but can you contextualize this wrt #159 ? Is it roughly the data structure that issue calls for to gather opaque entries, or something different?

rcoh:
this is basically a primitive designed to support #159 and similar use cases although as of the current design, it doesn't support metrics that outlive the current request (basically you have to keep the parent alive long enough).

This could be relaxed later though.

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+pool.with_exact_prefix("sdk.request.").append(metrics);
+```
+
+Static prefix segments allow the fully inflected paths to be interned and used
+by entry descriptors without leaking caller-owned runtime strings.
+
+### Parent close and late producers
+
+Closing the parent takes all child metrics available at that moment. A handle
+may outlive its parent, but appends made after the pool closes are discarded.
```
jlizen, line 195: Should we have any telemetry for this? Eg metrics or a rate limited warn?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+the original descriptor sequence contains fields removed by the overwrite
+pass. This trades a descriptor optimization for correctness only on the error
+path.
+
+Collision handling only compares child fields inside one pool. A flattened
+field cannot observe fields written by other fields in its parent, so the pool
+cannot detect a collision with a normal parent field or another flattened
+component. Producers must use distinct prefixes to avoid those collisions.
+
+A future collision policy could preserve every occurrence by renaming a group
```
jlizen, line 479: Are we comfortable recommending the interceptor pattern as-is given that even with the qualifier, retries will collide?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+winner for that fully-inflected name. Iteration still follows original append
+and field order; only superseded values are skipped.
+
+Descriptors are chained normally when the scan finds no collisions. If a
+collision exists, `descriptors()` returns `Descriptors::Unavailable`, because
+the original descriptor sequence contains fields removed by the overwrite
+pass. This trades a descriptor optimization for correctness only on the error
+path.
+
+Collision handling only compares child fields inside one pool. A flattened
```
jlizen, line 474: Is per-field collisions the right choice compared to hoisting collisions to be per-child? I'd be a bit worried about "frankenentries" where we have interwoven fields that apply to different retries, etc.

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+### Sampling and entry metadata
+
+Pooled children do not contribute sample-group elements. An opaque library or
+middleware must not silently change the parent request's sampling policy.
+Future tail-sampling support may add an explicit opt-in for known child types.
+
+Timestamp and `EntryConfig` calls are currently forwarded like ordinary
+flattened entries. Public examples should use field-only child metrics and put
+entry-level metadata on standalone wrappers. Before stabilization, this RFC
+should decide whether the pool should instead suppress child timestamp and
```
jlizen, line 495: I would assume that suppressing by default makes more sense...? And allow opting in if anybody wants it?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+    operation: "Second",
+    retry_count: 2,
+});
+```
+
+The parent contains `Operation = Second` and `RetryCount = 2`. A later optional
+field set to `None` emits no occurrence, so it does not overwrite an earlier
+value.
+
+`MetricsPoolEntry::write` evaluates closed child entries twice.
```
jlizen, line 446: This does run into some edge cases with interior mutabilty, probably we should call that out and guard with a debug assert.

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+for shared clients and middleware that cannot change every SDK call.
+
+## How to actually implement this RFC
+
+### Collector ownership
+
+`MetricsPool` owns:
+
+```rust,ignore
+Arc<Mutex<Option<Vec<BufferedEntry>>>>
```
jlizen, line 310: Do we want to have any backpressure on misbehaving inners growing the pool endlessly? Maybe some max size and a rate limited warn + drop?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+```rust,ignore
+let pool = cfg
+    .load::<CapturedPool>()
+    .and_then(|captured| captured.0.as_ref());
+
+if let Some(pool) = pool {
+    pool.append(invocation);
+} else {
+    drop(append_and_close(
+        StandaloneSdkInvocationMetrics { timestamp, invocation },
```
jlizen, line 267: Should we have some telemetry on this behavior? Or a "strict" mode that has a debug assert or similar if you don't expect inner spawns?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+- Concurrent futures polled on the same thread do not see each other's pools.
+- Nested scopes use the innermost pool.
+- A separately spawned task does not inherit the pool. Pass a
+  `MetricsPoolHandle` explicitly when detached work should contribute.
+
+This API does not require Tokio and is usable with any executor.
+
+### Prefixes and casing
+
+Prefixes are paths of static string segments:
```
jlizen, line 156: Seems fine to keep static for now, easy enough to evolve to support dynamic later

jlizen (commented):
Seems like this meets our needs, though I'm a bit nervous about some of the collision handling for v0, particularly the franken-entry risk with retries... We wouldn't want to cause alarm noise if a team has an error-like metric on first attempt, that vanishes on a subsequent successful retry.

docs/metrics-pool-rfc.md
jlizen: I guess last-retry-wins is probably right in general? Though do we have a convenient pattern to capture try count in a different layer?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+
+### Parent close and late producers
+
+Closing the parent takes all child metrics available at that moment. A handle
+may outlive its parent, but appends made after the pool closes are discarded.
+Closing does not wait for outstanding handles.
+
+`append_on_drop` guards therefore need to be dropped before the parent entry
+closes. Future work may add an explicit bounded wait policy, but an unbounded
+wait is not part of this proposal.
```
jasgin, line 200: I think this is the right call until we have the bounded wait policy, but to Jess's comment I think at least adding logging would be good to have in the meantime. Also, there might be another sharp edge if there are use cases where append_on_drop is used in a detached context from the parent.

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+such as `with_prefix` operate on an explicit `MetricsPoolHandle`. Library code
+should use `current()`'s `Option` to select its standalone fallback.
+
+The scope follows future polling rather than an operating-system thread:
+
+- A future can migrate between executor threads without losing its pool.
+- Concurrent futures polled on the same thread do not see each other's pools.
+- Nested scopes use the innermost pool.
+- A separately spawned task does not inherit the pool. Pass a
+  `MetricsPoolHandle` explicitly when detached work should contribute.
```
jasgin, line 150: How much of a footgun could this be? Should we think about ways to making it clear in terms of whether current returning None means no pool exists vs you crossed a spawn boundary?

docs/metrics-pool-rfc.md
```diff
@@ -0,0 +1,567 @@
...
+
+```rust,ignore
+if let Some(pool) = MetricsPool::current() {
+    pool.with_prefix(["sdk", "s3"]).append(suboperation_metrics);
+}
+```
+
+`MetricsPool::current()` is the only API that discovers scoped state. Methods
+such as `with_prefix` operate on an explicit `MetricsPoolHandle`. Library code
+should use `current()`'s `Option` to select its standalone fallback.
```
jasgin, line 142: How do we think consumers would handle fallback, generally? A fallback sink to a separate entry? Just thinking in terms of whether there's opportunity to reduce potential fallback boilerplate here.

docs/metrics-pool-rfc.md
jasgin: They can be accumulated separately at the invocation layer, similar to what the sdk example in the prototype shows

jasgin:
Revised and addressed comments in latest commit, also aligned prototype

rcoh:
LGTM!

jasgin (approved):

Issue #359, opened 2026-08-06
Formatters should detect incompatible entries using entry descriptors


popematt:
> Another thing that we can do to mitigate (that I've been meaning to do anyway is use the entry descriptors to make it really easy to validate a given entry against a given format (e.g. emitting a warning if you use an object/property but the format has no way to render it.

 _Originally posted by @rcoh in [#355](https://github.com/awslabs/metrique/issues/355#issuecomment-5191044830)_

I'm opening this issue so that there's somewhere to track this and I can reference it from other issues/RFC, etc.

Things to consider:
* When should the incompatible entry detection happen—on every write, in some sort of registration hook, something else?
* What should happen when an incompatible entry is detected—log a warning, emit a metric about the metrics pipeline, return an error? This will likely depend on when/where the detection takes place.

Issue #355, opened 2026-07-30
feat: Add support for object property values


popematt:
## Problem

Metrique's `#[metrics(value)]` mode maps a struct to a single scalar value. This works for simple metric dimensions and counters, but it can't represent structured contextual data that should be attached to a metric entry as a nested JSON object (e.g., request context, error metadata, or trace annotations).

Today, if you want to attach a multi-field blob of context to a metric entry, you have two options:

1. Flatten every field into a top-level metric (pollutes the metric namespace because each field becomes its own "metric" even when it's not one)
2. Serialize to a string manually and emit it as a string value (loses structure for any format that supports native objects, like EMF, and makes querying via `jq` or CWL Insights more tedious)

Neither option is great. The first changes the semantics of the data; the second throws away structure that downstream systems (CloudWatch, JSON parsers) could otherwise use.

Concrete example: today a field like `context` must either become multiple top-level fields (`ContextRequestId`, `ContextRetries`, ...) or a string field such as `"Context":"{\"request_id\":\"abc\",\"retries\":2}"`.

The proposed feature would allow the same field to be emitted as `"Context":{"request_id":"abc","retries":2}`.

## Proposed Design

Add a new mode: `#[metrics(value(object))]`. A struct annotated with this mode:

1. Implements `ObjectValue` (a visitor trait with a `write_object(&self, writer: &mut impl ObjectWriter)` method)
2. Implements `Value` by calling `writer.object(self)`, delegating to the format's native object representation
3. Generates a `*Value` closed-value struct (same naming convention as `#[metrics(value)]`)
4. Closes fields via `CloseValue` as usual

```rust
#[metrics(value(object))]
struct RequestContext {
    request_id: &'static str,
    retries: u64,
    detail: RequestDetail,                // nested objects compose
    maybe_detail: Option<RequestDetail>,  // None fields are omitted
}
```

When used as a field in a `#[metrics]` struct, the object is emitted as a native JSON object in formats that support it (EMF, JSON) and falls back to a JSON-encoded string for formats that don't.

### New traits

```rust
pub trait ObjectWriter {
    fn field(&mut self, name: &str, value: &(impl Value + ?Sized));
}

pub trait ObjectValue {
    fn write_object(&self, writer: &mut impl ObjectWriter);
}
```

`ObjectWriter::field` receives each named field as a `Value`, so nested objects, arrays (`values()`), scalars, and `Option` (which writes nothing when `None`) all compose without special-casing.

[...]

PR #360, opened 2026-08-06
docs: add object-valued fields RFC

popematt:
📬 *Issue #, if available:*

Proposal to add support for #355 

✍️ *Description of changes:*

 - Adds ValueWriter::object() method (default = drop; EMF/JSON override to native nested JSON)
 - `#[metrics(format = AsObject)]` renders any `#[metrics]` struct as a nested object — no re-annotation on the nested type
 -  `Each<F>` format combinator for arrays of objects (`format = Each<AsObject>` on `Vec<Phase>`)
 - Bare struct-typed field is a compile error; object rendering is always explicit
 - Descriptor-based warning when an object field targets a format without native support (tmp solution pending #359)

I do not feel strongly about any of the type names, and welcome any suggestions if you prefer something different.
I would like your opinion on the `BoxEntry` issue that is called out in the RFC—whether we need to solve it now, best way to deal with it, etc.

[Rendered](https://github.com/popematt/metrique/blob/object-rfc/docs/object-values-rfc.md)

- [ ] I confirm that I've made a best effort attempt to update all relevant documentation.  **<--_Is this relevant for a RFC?_**


docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+existing `#[metrics(format = ...)]` attribute, a nested object reuses the
+`Entry` machinery rather than introducing a parallel one, and any ordinary
+`#[metrics]` struct can be nested with no re-annotation. A bare struct-typed
+field remains a compile error, so nothing becomes an object by accident._
+
+---
+
+## Motivation
+
+`metrique` maps each field of a `#[metrics]` struct to a flat scalar. A field is
```
rcoh, line 33: not technically true, fields can actually contain histograms / repeated observations

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+    {
+      "Type": "eval",
+      "Duration": 22,
+      "Phases": [
+        {
+          "Type": "multiply",
+          "Duration": 33,
+          "Phases": [
+            { "Type": "add", "Duration": 44 },
+            { "Type": "sub", "Duration": 55 }
```
rcoh, line 79: just curious, how would you actually query this with cloudwatch?

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+## The user experience if this RFC is implemented
+
+This section is written as the public documentation users should be able to read
+once the feature is released.
+
+### Nesting an existing metric struct
+
+Any `#[metrics]` struct can be nested inside another entry as an object. The
+nested struct requires no new annotation; the parent field selects object
+rendering with `#[metrics(format = AsObject)]`:
```
rcoh, line 112: cool!

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+When `RequestMetrics` closes, `root_phase` is a nested object:
+
+```json
+{
+  "RequestId": "...",
+  "RootPhase": { "Type": "parse_expr", "Duration": 11 }
+}
+```
+
+The nested object is a fresh naming scope. `Phase` applies its own `rename_all`;
```
rcoh, line 142: is this a requirement of the way the traits work or an intentional feature? Currently renames apply transitively so this may be a bit surprising.

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+   causes `E0283` inference ambiguity at the macro's use site (the compiler
+   cannot choose between `Lifted` and `NotLifted` when both satisfy the bound).
+2. Without a concrete `Lifted` impl, `Option`/`Box`/`Arc` auto-lifting does not
+   work (the lifting blankets are keyed on `L = Lifted`).
+
+The derive generates `impl ValueFormatter<PhaseEntry> for AsObject` per type.
+Hand-implementors write the same impl. Because every impl is concrete and
+`Lifted`, the lifting blankets compose automatically for all types.
+
+### The `Each` format combinator
```
rcoh, line 365: is `Each` actually required for coherence? I guess "yes"? you could imagine that we gracefully handled `Vec<Phase>` via `AsObject`

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+
+impl<V, F> ValueFormatter<Vec<V>> for Each<F>
+where
+    F: ValueFormatter<V>,
+{
+    const SHAPE: FieldShape<'static> =
+        FieldShape::List(ShapeRef::new(&<F as ValueFormatter<V>>::SHAPE));
+
+    fn format_value(writer: impl ValueWriter, value: &Vec<V>) {
+        let wrapped: SmallVec<[FormattedValue<_, F, _>; VALUES_INLINE_CAPACITY]> =
```
rcoh, line 387: since this is iterator in iterator out why do we have the collect it? lifetime issues?

rcoh (commented):
this seems pretty reasonable and also forward looking; once we add `.object`, you could have an out-of-tree adapter to work with `serde`, e.g.

docs/object-values-rfc.md
popematt: I got my DSLs mixed up 🤦‍♂️. CWL Insights doesn't support any sort of multi-level queries. I guess we'll have to query it with something else regardless of whether we use inline json strings or native objects. That being said, this feature is still potentially useful to us because if we can't query with CWL anyway, then we _might_ end up emitting these particular metrics as binary Ion to reduce the size of our logs.

docs/object-values-rfc.md
popematt: I don't think it's required for coherence, but I haven't really looked into that. The original motivation was because I was originally going to have AsObject fallback to writing as a Json string instead of dropping for unsupported formats. Under that design, it was useful (necessary, even) to have a way to distinguish between stringified-array-of-objects and array-of-stringified-objects.

It's no longer needed for that, but I kept it because:
* I think explicit is better so that there's no surprises
* I like the composability of it
* It helps with the `Vec<Vec<T>>` situation that is unsupported right now.
* It can help differentiate `AsSerdeJsonString` from `Each<AsSerdeJsonString>`, etc. sorts of use cases in the future where a user might want to control whether to emit stringified-array-of-objects vs array-of-stringified-objects.

I'm open to dropping it though, if you think it's unnecessary complexity.

docs/object-values-rfc.md
jlizen: I think CWL insights does support multi-level?
Ref: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_AnalyzeLogData-discoverable-fields.html

jlizen: (Anyway even if it didn't, I agree that this is still useful)

docs/object-values-rfc.md
jlizen: ++ not seeing why we wouldn't preserve this to be transitive to nested objects

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+
+## Terminology
+
+- **Object field**: A field of a `#[metrics]` struct that is rendered as a
+  nested object rather than a flat scalar.
+- **Object payload**: The value written as an object. Either an ordinary
+  `#[metrics]` struct or a type that hand-implements `ObjectValue`.
+- **Native object**: A structured object emitted directly in a format that
+  supports it (a JSON object in JSON and EMF; a self-describing value in a
+  future OTel integration).
```
jlizen, line 101: We have otel integration now FYI (`metrique-otel)`, so ideally we could ship all three with v0.

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+impl ObjectValue for EndpointObject {
+    fn write_object<'a>(&'a self, writer: &mut impl EntryWriter<'a>) {
+        writer.value("host", &self.0.host);
+        writer.value("port", &self.0.port);
+    }
+}
+
+impl ValueFormatter<EndpointObject> for AsObject {
+    fn format_value(writer: impl ValueWriter, value: &EndpointObject) {
+        writer.object(value)
```
jlizen, line 208: This feels a bit verbose to have to specify the formatter impl... Would this ever not just be `writer.object()` inline? Is there not something we could do with an intrinsic or default method that wouldn't cause coherence issues?

docs/object-values-rfc.md
```diff
@@ -0,0 +1,636 @@
...
+`L`-inference is unambiguous at the macro's use site.
+
+### What happens on formats without native objects
+
+On formats that support native objects (JSON, EMF), `AsObject` renders the field
+as a structured object. On formats that do not, the field is **dropped** — it
+does not appear in the output. A descriptor-time warning (see below) informs the
+user at setup which fields will not render on which formats.
+
+Dropping is preferable to a string fallback as the default: it avoids surprise
```
jlizen, line 234: Not sure I agree with this logic... I would have assumed that it is nicer to fall back to stringified JSON as a default and maybe let formats opt out of that. Dropping feels pretty surprising...? Ultimately you can recover stringified JSON into processable json if you are so motivated, and somebody thought the fields were relevant enough to include them in the first place....

docs/object-values-rfc.md
popematt: It supports multilevel, but it doesn't support queries like "select log events where any element in `@details.nested_events` matches `kind == foo` and `status == error`" so it's not possible to query the nested tree structure in the way I had proposed. 

I do still think it's useful, but we might push our aggregate metrics as EMF to CWL and our sampled metrics as json to some other destination so it's easy to process with some other query engine.

docs/object-values-rfc.md
popematt: The iterator in has one type, and the iterator out must be references of a different type. The collect is necessary to avoid having references to temporary values.

I'm looking into other options here.

docs/object-values-rfc.md
popematt: This was an oversight. I don't want this to be surprising so I'll follow whatever precedent already exists.

docs/object-values-rfc.md
popematt: I saw the `metrique-otel` crate, but left it as future work because the OTel SDK for Rust doesn't have support for object values yet in [`opentelemetry::Value`](https://docs.rs/opentelemetry/latest/opentelemetry/enum.Value.html).

(If that changes soon, then absolutely we include it in v0.)

docs/object-values-rfc.md
popematt: > you could imagine that we gracefully handled `Vec<Phase>` via `AsObject`

As I've tried to solve the allocation here, I realized that the `Each` combinator appears nice, but it's not strictly needed, even for the `AsSerdeJsonString` case that I mentioned in my other comment. The only thing we lose is support for `Vec<Vec<T>>`, which probably isn't worth fighting too hard for right now. I don't have a need for it, and it seems a lot more niche of a use case than the object support.

I'll update things to drop `Each`, which will remove this allocation.

docs/object-values-rfc.md
popematt: Yeah, now that I've stepped away for a bit and come back to it, I definitely agree with you more than I agree with me from 2 weeks ago.

docs/object-values-rfc.md
popematt: Okay, so after looking at this again, here are a few more thoughts:

For flattening, everything that gets renamed is ending up in the top-level struct/object, so it makes sense to use the parent name strategy. For nested objects, the fields are in a different struct/object, so it seems less clear to me. In the case of a `#[metrics]` struct that is also being used a nested object, do we want it to potentially have different names depending on whether it gets emitted as a top-level object or as a nested object?

I'd like to hear what you think about this. In the meantime, I'm also thinking about way that we could make it support both (based on an annotation or something).

docs/object-values-rfc.md
popematt: This is no longer needed with the updated design.

rcoh (approved):
Awesome LGTM! My only concern is the lack of rename_all propagation but I don't feel that strongly.

PR #361, opened 2026-08-06
fix(metrique-writer-core): resolve AttachHandle::drop panic on concurrent shutdown registration

### LINKED ISSUE #340: `AttachHandle::drop` panics when `register_shutdown_fn` races with detachment
### Summary

Found while working on https://github.com/awslabs/metrique/issues/281. Reliably reproducible with a Shuttle-based test (see below); when run explicitly, it fails deterministically within the first 1-2 iterations.

`AttachHandle::drop` ([metrique-writer-core/src/global.rs](https://github.com/awslabs/metrique/blob/432c21e9f40483677394afa78c9a2ebafd5465c3/metrique-writer-core/src/global.rs#L370)) assumes it always holds the sole strong `Arc<ShutdownRegistry>` reference, and calls `Arc::try_unwrap` on that assumption:

```rust
fn drop(&mut self) {
    if let Some(arc) = self.shutdown_registry.take() {
        // The macro holds only a Weak reference, so this is the sole strong ref.
        match Arc::try_unwrap(arc) {
            Ok(registry) => registry.drain_and_run(),
            Err(_) => unreachable!("ShutdownRegistry should have no other strong references"),
        }
    }
}
```

When [AttachGlobalEntrySink::register_shutdown_fn ](https://github.com/awslabs/metrique/blob/432c21e9f40483677394afa78c9a2ebafd5465c3/metrique-writer-core/src/global.rs#L598) runs (or the equivalent macro-generated code in `global_entry_sink!`), it upgrades a `Weak<ShutdownRegistry>` to a strong `Arc` in order to call `.push(f)`, briefly creating a second strong reference:

```rust
if let Some(registry) = weak.upgrade() {
    registry.push(f);
}
```
If a thread is inside that `upgrade()` → `push()` window (holding a live, temporary strong `Arc`) at the exact moment another thread drops the `AttachHandle`. `Arc::try_unwrap` correctly observes two strong references, returns `Err`, and the `unreachable!()` panics.

### Precondition
Any service calls `register_shutdown_fn` from one thread while another thread concurrently drops/detaches the same global
sink (e.g. during shutdown, if some component registers a cleanup hook at the same moment the main thread starts tearing everything down).

#### Shuttle
```rust
#[cfg(all(test, shuttle, feature = "_shuttle"))]
mod shuttle_repro {
    use std::sync::Arc;
    use std::sync::atomic::{AtomicUsize, Ordering};
    use super::{AttachHandle, ShutdownFn};

    fn concurrent_register_and_drop() {
        let ran = Arc::new(AtomicUsize::new(0));
        let handle = AttachHandle::new(|| {});
        let weak = handle.shutdown_registry_weak();

        let ran2 = ran.clone();
        let registrar = shuttle::thread::spawn(move || {
            if let Some(registry) = weak.upgrade() {
                registry.push(ShutdownFn::new(move || {
                    ran2.fetch_add(1, Ordering::SeqCst);
                }));
            }
        });

        // Without an explicit yield here, shuttle can only schedule
        // `registrar` entirely before or entirely after `drop` below --
        // never *during* it, so the race would never be reachable.
        shuttle::thread::yield_now();

        drop(handle);
        registrar.join().unwrap();

        assert!(ran.load(Ordering::SeqCst) <= 1);
    }

    #[test]
    fn concurrent_register_and_drop_pct() {
        shuttle::check_pct(concurrent_register_and_drop, 5_000, 2);
    }
}
```

Requires adding `shuttle = "0.9"` as an optional/dev-dependency and a `_shuttle` feature (or just running with `--features shuttle` against a manually added dependency). Run with:

```bash
RUSTFLAGS="--cfg shuttle" cargo test -p metrique-writer-core --lib \
  --features _shuttle -- concurrent_register_and_drop_pct
```

<details>
  <summary> Isolated reproduction using OS threads</summary>

- **Unsynchronized attempt** — spawn a thread, then immediately race against it, with nothing coordinating the two. Standalone: no `metrique` or Shuttle dependency, just `std::sync`, and only reuses the *shape* of the real race (`Weak::upgrade` racing `Arc::try_unwrap`) `Registry` here is a stand-in with no relation to `ShutdownRegistry`:

  ```rust
  use std::sync::{Arc, Weak};
  use std::thread;

  struct Registry(Vec<u8>);

  fn attempt() -> bool {
      let arc = Arc::new(Registry(vec![1]));
      let weak: Weak<Registry> = Arc::downgrade(&arc);

      let registrar = thread::spawn(move || {
          if let Some(strong) = weak.upgrade() { std::hint::black_box(&strong); } // race target
      });

      let hit_race = Arc::try_unwrap(arc).is_err(); // fires right away, no waiting
      registrar.join().unwrap();
      hit_race
  }
  ```

- **Synchronized attempt** — use a `Barrier` to force both threads to start the racy part at the same instant, more help than the real code ever gets (which has none at all). Same standalone setup and stand-in `Registry` as the unsynchronized attempt above, just with a `Barrier` added:

  ```rust
  use std::sync::{Arc, Barrier, Weak};
  use std::thread;

  struct Registry(Vec<u8>);

  fn attempt(barrier: &Barrier) -> bool {
      let arc = Arc::new(Registry(vec![1]));
      let weak: Weak<Registry> = Arc::downgrade(&arc);

      thread::scope(|scope| {
          scope.spawn(|| {
              barrier.wait(); // blocks until the main thread also arrives
              if let Some(strong) = weak.upgrade() { std::hint::black_box(&strong); }
          });

          barrier.wait(); // blocks until the spawned thread also arrives
          Arc::try_unwrap(arc).is_err()
      })
  }
  ```
</details>

#### Shuttle

Instrumented `concurrent_register_and_drop_pct` (`check_pct(concurrent_register_and_drop, 5_000, 2)`, i.e. up to 5,000 iterations allowed) with a counter tracking exactly which iteration the panic occurs on:

```rust
static TEMP_ITER_COUNTER: AtomicUsize = AtomicUsize::new(0);

fn concurrent_register_and_drop() {
    TEMP_ITER_COUNTER.fetch_add(1, Ordering::SeqCst); // +1 at the top of every check_pct iteration
    // ...unchanged body...
}

#[test]
fn temp_measure_iters_to_failure() {
    TEMP_ITER_COUNTER.store(0, Ordering::SeqCst);
    let prev_hook = std::panic::take_hook();
    std::panic::set_hook(Box::new(move |info| {
        // check_pct's iterations run sequentially in-process and stop as
        // soon as one panics, so the counter's value at panic time is
        // exactly the 1-based iteration number that failed.
        eprintln!(
            "[measure] panicked after {} iterations",
            TEMP_ITER_COUNTER.load(Ordering::SeqCst)
        );
        prev_hook(info);
    }));
    shuttle::check_pct(concurrent_register_and_drop, 5_000, 2);
}
```

Ran it **28 times** — each run gets a fresh random seed, so each of the 28 runs explored a genuinely different sequence of schedules:

**All 28 out of 28 runs hit the panic within iteration 1 or 2** — i.e. effectively the first schedule PCT's depth-2 change-point selection tries,
regardless of random seed.

#### Isolated reproduction using OS threads

**Unsynchronized attempt:** **0 hits in 200,000 iterations.** Starting a new OS thread just takes longer than the few instructions the main thread runs right after it. So by the time `try_unwrap` runs, the spawned thread usually hasn't even started yet — the two sides never actually overlap.

**Synchronized attempt:** **1 hit in 200,000 iterations**, then **3 hits in 1,000,000** on a repeat   run — roughly **1-in-300,000**. A `Barrier` makes both threads wait for each other and then start at the same moment, so this is about as close to "same time" as two real threads can get. But the actual unsafe window — the moment between `weak.upgrade()` succeeding and that temporary `Arc` going out of scope — only lasts a few nanoseconds. Even starting "at the same time," normal jitter from the OS and CPU means the two threads   usually still miss that tiny window.

### Version

- `metrique-writer-core` 0.1.18, `metrique-writer` 0.1.24 (workspace at  `main` commit `5d3a198ae24cc1f00d4f94136a65ac939ec1a21e`, "Release: Jul 8th 2026 (#337)").

### Environment

Measured on an Apple M2 (8 logical CPUs), macOS (Darwin 25.5.0), rustc 1.95.0, release build (`rustc -O`). Noting this because race-hit-rates like these are sensitive to core count, OS scheduler behavior, and system load — the exact numbers below are this machine's result.

tiny-ac:
📬 *Issue #, if available:*
closes #340 

✍️ *Description of changes:*

`AttachHandle::drop` used Arc::try_unwrap to check it was the sole owner of ShutdownRegistry before draining shutdown functions, panicking via unreachable!() if a concurrent register_shutdown_fn() held a transient extra reference.

Fixed by giving ShutdownRegistry explicit open/closed state instead of inferring it from refcounting. Push now returns whether it was accepted, and AttachHandle::drop no longer needs sole ownership to drain.

Similar kind of gap found at attach()'s own SINK/SHUTDOWN_REGISTRY, leaving a window where try_sink() reported "attached" but a concurrent register_shutdown_fn() still saw the stale registry. Fixed by merging into one ATTACHED slot written atomically.

Another potential issue: AttachHandle::drop could handle a panicking shutdown fn the same way sysinfo_reporter.rs does by wrapping its polling loop in catch_unwind.

Added shuttle tests and regular tests.


tiny-ac:
Tried some other approaches before choosing this one.

The one I went with makes ShutdownRegistry track its own open/closed state explicitly, guarded by its internal mutex, instead of leaning on Arc refcounts.

Some of the other approaches I tried:
1. make the static hold a strong Arc instead of a Weak. Gets rid of the .upgrade() step, but turned out to need more changes overall. `forget()` has to explicitly close the registry now instead of getting that for free from the refcount dropping to zero
  ```
static SHUTDOWN_REGISTRY: RwLock<Option<Arc<ShutdownRegistry>>> = ...; // no Weak
  fn register_shutdown_fn(f: ShutdownFn) {
      SHUTDOWN_REGISTRY.read().unwrap().clone().unwrap().push(f);
  }
```

  2. Swallow the race silently instead of panicking. It does technically stop the panic, but the test showed the shutdown fn (and the sink-detach cleanup) just silently never runs, with no error at all. That felt worse than the original bug, so I dropped it too.
```
  if let Ok(registry) = Arc::try_unwrap(arc) {
      registry.drain_and_run();
  }
  // else: no panic, but shutdown fns silently never run
```

3. Switch the internal lock to an RwLock doesn't even compile without adding an unsafe impl Sync. RwLock<T> needs T: Sync and ShutdownFn isn't, whereas Mutex<T> only needs T: Send.
  `pub struct ShutdownRegistry(RwLock<Option<Vec<ShutdownFn>>>); // push/drain both .write()`

File: metrique-writer-core/src/global.rs [hunk 1]
```diff
@@ -510,12 +527,19 @@ macro_rules! global_entry_sink {
...
-            static SINK: RwLock<Option<(BoxEntrySink, Box<dyn Send + Sync + 'static>)>> = RwLock::new(None);
-            static SHUTDOWN_REGISTRY: RwLock<Option<Weak<ShutdownRegistry>>> = RwLock::new(None);
+
+            // Both fields are set together by `attach()` and cleared together by the
+            // shutdown fn it registers, under one lock. This way, a reader can never observe
+            // one half of "attached" without the other.
+            struct AttachedState {
+                sink: (BoxEntrySink, Box<dyn Send + Sync + 'static>),
+                shutdown_registry: Weak<ShutdownRegistry>,
+            }
```
tiny-ac, line 566: While looking for gaps in coverage across similar locks in global.rs, SINK/SHUTDOWN_REGISTRY stood out. Tracing through attach() by hand to check surfaced the race described above.

metrique-writer-core/src/global.rs
```diff
@@ -381,10 +400,8 @@ impl Drop for TokioRuntimeTestSinkGuard {
...
         if let Some(arc) = self.shutdown_registry.take() {
-            // KNOWN BUG (see issue https://github.com/awslabs/metrique/issues/340)
-            match Arc::try_unwrap(arc) {
-                Ok(registry) => registry.drain_and_run(),
-                Err(_) => unreachable!("ShutdownRegistry should have no other strong references"),
+            for f in arc.drain().into_iter().rev() {
+                f.call();
             }
         }
     }
```
tiny-ac, line 430: Potential issue: `register_shutdown_fn` is public API that accepts any closure a caller wants to register, with nothing preventing it from panicking.

metrique-util's `tokio-metrics` and `sysinfo` integrations both register trivially panic-safe closures, so not a live incident today.

`AttachHandle::drop` could handle panicking the same way sysinfo_reporter.rs's own background worker already wraps its polling loop in catch_unwind. (https://github.com/awslabs/metrique/blob/a3d8e9431a008495f1de69609ea19e05efcb6fca/metrique-util/src/sysinfo_reporter.rs#L586-L637)

metrique-writer-core/src/global.rs
```diff
@@ -381,10 +400,33 @@ impl Drop for TokioRuntimeTestSinkGuard {
...
+                            .copied()
+                            .or_else(|| p.downcast_ref::<String>().map(String::as_str))
+                            .unwrap_or("<non-string panic payload>")
+                    })
+                    .collect();
+                panic!(
+                    "{} of {total} shutdown fn(s) panicked during AttachHandle::drop \
+                     (every shutdown fn still ran to completion, including cleanup): {messages:?}",
+                    panics.len()
+                );
```
tiny-ac: Image

metrique-writer-core/src/global.rs
```diff
@@ -283,28 +283,47 @@ type ShutdownRegistryMutex<T> = std::sync::Mutex<T>;
 #[cfg(all(shuttle, feature = "_shuttle"))]
 type ShutdownRegistryMutex<T> = shuttle::sync::Mutex<T>;
 
+/// Lock type for the macro-generated `ATTACHED` static
+#[doc(hidden)]
+#[cfg(not(all(shuttle, feature = "_shuttle")))]
+pub type AttachedLock<T> = std::sync::RwLock<T>;
```
rcoh: Let's not make pub types; this does not provide any actual value but leaks a public API surface

metrique-writer-core/src/global.rs
```diff
@@ -283,28 +283,47 @@ type ShutdownRegistryMutex<T> = std::sync::Mutex<T>;
...
+#[doc(hidden)]
+#[cfg(all(shuttle, feature = "_shuttle"))]
+pub type AttachedLock<T> = shuttle::sync::RwLock<T>;
```
rcoh: Same here.

We should be using a primitives module and the primitives module should expose RwLock

metrique-writer-core/src/global.rs
```diff
@@ -283,28 +283,47 @@ type ShutdownRegistryMutex<T> = std::sync::Mutex<T>;
...
     fn new(initial: ShutdownFn) -> Self {
-        Self(ShutdownRegistryMutex::new(vec![initial]))
+        Self(ShutdownRegistryMutex::new(Some(vec![initial])))
     }
 
     /// Add a shutdown function. Functions run in LIFO order when the
     /// [`AttachHandle`] is dropped.
+    ///
+    /// Returns `false` (and does not add `f`) if the registry has already been closed by a
+    /// concurrent (or prior) [`ShutdownRegistry::drain`].
```
rcoh, line 317: Any idea why this is doc(hidden)?

metrique-writer-core/src/global.rs
```diff
@@ -1394,6 +1448,202 @@ mod shutdown_registry_tests {
...
+        use std::sync::Barrier;
+
+        const ROUNDS: usize = 200_000;
+        const MAX_PROBES_PER_ROUND: usize = 1_000;
+
+        let bad_panic = Arc::new(AtomicBool::new(false));
+        let start_barrier = Arc::new(Barrier::new(2));
+        let done_barrier = Arc::new(Barrier::new(2));
+
+        std::thread::scope(|scope| {
```
rcoh: If we are writing tests with threads the tests should use shuttle which makes them deterministic and reproducible

metrique-writer-core/src/global.rs
```diff
@@ -381,10 +400,33 @@ impl Drop for TokioRuntimeTestSinkGuard {
 impl Drop for AttachHandle {
     fn drop(&mut self) {
         if let Some(arc) = self.shutdown_registry.take() {
-            // KNOWN BUG (see issue https://github.com/awslabs/metrique/issues/340)
-            match Arc::try_unwrap(arc) {
-                Ok(registry) => registry.drain_and_run(),
-                Err(_) => unreachable!("ShutdownRegistry should have no other strong references"),
+            let fns = arc.drain();
```
rcoh: I'm not positive it's worth the danger of unwind safety to call multiple handlers since we are going to panic anyway

metrique-writer-core/src/global.rs
```diff
@@ -381,10 +400,33 @@ impl Drop for TokioRuntimeTestSinkGuard {
...
+            let total = fns.len();
+            let mut panics = Vec::new();
+            for f in fns.into_iter().rev() {
+                // wrap in catch_unwind so we can get to the detach fn even if a previous fn call panics
```
rcoh: A clean way to handle this is to put the fns in a wrapper that calls them on drop

metrique-writer-core/src/global.rs
tiny-ac: The idea is to isolate each call so every shutdown fn still runs and the sink still gets detached, even if one of them panics.

Left unwrapped, the sink can get stuck "attached". A panic skips every fn registered after it (including the detach fn). Trying to [reattach](https://github.com/awslabs/metrique/blob/a3d8e9431a008495f1de69609ea19e05efcb6fca/metrique-writer-core/src/global.rs#L198) after that will panic.

I'll revert if this is not a concern right now and file an issue if it might become a concern in the future. What do you think?

metrique-writer-core/src/lib.rs
```diff
@@ -48,3 +48,9 @@ pub mod test_stream;
 #[cfg(all(shuttle, feature = "_shuttle"))]
 #[doc(hidden)]
 pub mod shuttle_test_support;
+
+/// Cfg-gated concurrency primitives (std vs. shuttle) used by macro-generated
+/// code. Public only so that expansion works from downstream crates; not
+/// meant to be used directly.
+#[doc(hidden)]
+pub mod shuttle_primitives;
```
tiny-ac: Tried pub(crate) just in case:
Image

metrique-writer-core/src/global.rs
rcoh: I see; does that mean you would have had to catch_unwind to begin with to get into this state?

Generally if your code panics, I'm not _super_ concerned about recovering from it, its generally safer to have the application crash.

I think I don't totally understand the failure mode. Is it reachable from normal code?

tiny-ac: With the current internal implementations for tokio/sysinfo, this wouldn't be much of an issue (they run infallible operations). `register_shutdown_fn` is public API, though, so any callback registered here can, if it panics, interrupt execution of the callbacks still waiting to run.

- Fns run in LIFO order, and the detach function is always the first one registered.
  - If a fn panics, none of the fns registered after it run, because the panic aborts the loop.
  - Thus, the sink remains attached: the detach fn runs last, and execution never reaches it once an earlier panic aborts the loop.

The code in main is a plain loop with no catching.
- The loop aborts if one of the fns panics.
- Remaining fns won't execute.

```
    for f in self.0.into_inner().unwrap().into_iter().rev() {
        f.call();
    }
```

This behavior `predates these changes.` I ran these tests in main (added to `metrique-writer-core/src/global.rs`)

```rust
#[test]
fn repro_panicking_shutdown_fn() {
    // Just an ordinary panicking closure and a drop.
    metrique_writer::sink::global_entry_sink! { Sink }
    let TestEntrySink { sink, .. } = test_entry_sink();

    let handle = Sink::attach((sink, ()));
    Sink::register_shutdown_fn(ShutdownFn::new(|| {
        panic!("ordinary bug in a shutdown fn, e.g. an unwrap() on None");
    }));

    // Catch it only so this test doesn't abort the whole test
    let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| drop(handle)));

    assert!(
        Sink::try_sink().is_some(),
        "BUG: the sink is still reported as attached -- detach never ran"
    );

    // Any future attach() on this sink type is now permanently broken.
    let TestEntrySink { sink, .. } = test_entry_sink();
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        Sink::attach((sink, ()))
    }));
    assert!(
        result.is_err(),
        "BUG: a fresh attach() should have panicked with \"Already installed\" \
         but it succeeded -- the sink was wedged forever after the first panic"
    );
}
```

This is the output
```
thread '...repro_panicking_shutdown_fn_wedges_the_sink' panicked at metrique-writer-core/src/global.rs:1284:13:
ordinary bug in a shutdown fn, e.g. an unwrap() on None

thread '...repro_panicking_shutdown_fn_wedges_the_sink' panicked at metrique-writer-core/src/global.rs:1279:9:
Already installed a global Sink sink, drop the attach handle first if intentionally attaching a new sink

test global::shutdown_registry_tests::repro_panicking_shutdown_fn_wedges_the_sink ... ok
```

metrique-writer-core/src/global.rs
tiny-ac: doc(hidden) as opposed to public and documented API or private?

If you mean private, that's because it's used by the `global_entry_sink!` and compilation fails if it's not public (it compiles in whatever crate calls the macro and they need to be able to reach `push`)
image

If you mean public and documented, `register_shutdown_fn` is the clean entry endpoint

metrique-writer-core/src/global.rs
tiny-ac: I tried this, which double panics and aborts, so not much of an improvement (ran the test `multiple_panicking_shutdown_fns_all_run_without_aborting`)

<details>
  <summary>Test code (multiple_panicking_shutdown_fns_all_run_without_aborting)</summary>

  ```rust
  fn multiple_panicking_shutdown_fns_all_run_without_aborting() {
      metrique_writer::sink::global_entry_sink! { Sink }
      let TestEntrySink { sink, .. } = test_entry_sink();

      let handle = Sink::attach((sink, ()));
      Sink::register_shutdown_fn(ShutdownFn::new(|| panic!("first")));
      Sink::register_shutdown_fn(ShutdownFn::new(|| panic!("second")));

      let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| drop(handle)));

      let payload = result.expect_err("drop() must still re-panic with the aggregated failure");
      let msg = payload
          .downcast_ref::<&str>()
          .copied()
          .or_else(|| payload.downcast_ref::<String>().map(String::as_str))
          .unwrap_or_default();
      assert!(
          msg.contains("one or more shutdown fn(s) panicked"),
          "expected an aggregated message naming both panics, got: {msg}"
      );
      assert!(
          Sink::try_sink().is_none(),
          "sink must still be detached even though two shutdown fns panicked"
      );
  }
  ```
  </details>


```rust
  struct RunOnDrop(Option<ShutdownFn>);
  impl Drop for RunOnDrop {
      fn drop(&mut self) {
          if let Some(f) = self.0.take() {
              f.call();
          }
      }
  }

  impl Drop for AttachHandle {
      fn drop(&mut self) {
          if let Some(arc) = self.shutdown_registry.take() {
              let fns: Vec<RunOnDrop> = arc.drain().into_iter().rev().map(|f| RunOnDrop(Some(f))).collect();
              drop(fns);
          }
      }
  }
```

Output:
image
image


Then this version includes a tweak to catch each fn's panic internally (loses granularity compared to the current committed version)

```rust
  struct RunOnDrop<'a> {
      f: Option<ShutdownFn>,
      any_panicked: &'a std::cell::Cell<bool>,
  }
  impl Drop for RunOnDrop<'_> {
      fn drop(&mut self) {
          if let Some(f) = self.f.take() {
              if std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| f.call())).is_err() {
                  self.any_panicked.set(true);
              }
          }
      }
  }

  impl Drop for AttachHandle {
      fn drop(&mut self) {
          if let Some(arc) = self.shutdown_registry.take() {
              let any_panicked = std::cell::Cell::new(false);
              let fns: Vec<RunOnDrop> = arc.drain().into_iter().rev()
                  .map(|f| RunOnDrop { f: Some(f), any_panicked: &any_panicked })
                  .collect();
              drop(fns);
              if any_panicked.get() && !std::thread::panicking() {
                  panic!("one or more shutdown fn(s) panicked during AttachHandle::drop");
              }
          }
      }
  }
```

Output:
image

metrique-writer-core/src/global.rs
tiny-ac: If this needs more discussion than fits here, I could file it as a separate issue to keep exploring alternatives (to avoid blocking this PR)

metrique-writer-core/src/global.rs
```diff
@@ -268,43 +268,39 @@ impl ShutdownFn {
...
+            None => false,
+        }
     }
 
-    pub(crate) fn drain_and_run(self) {
-        for f in self.0.into_inner().unwrap().into_iter().rev() {
-            f.call();
-        }
+    /// Close the registry and return whatever functions were registered before closure, in
+    /// registration order. Whichever of `push` or `drain` acquires the lock first wins outright.
```
rcoh: does this change the ordering that shutdown functions are called in?

tiny-ac: No, still calling `.into_iter().rev()`, so the order is the same as before

rcoh (approved):

PR #362, opened 2026-08-06
feat: `AppendAndCloseOnDrop` now avoids allocations unless required

popematt:
`AppendAndCloseOnDrop` now skips allocations unless a `.flush_guard()` is created

📬 *Issue #, if available:*

None that I'm aware of. I think this is a simple enough change that one is probably not required, but I can make one if you prefer.

✍️ *Description of changes:*

`AppendAndCloseOnDrop` now lazily allocates

### DIFF
--- a/metrique/src/lib.rs
+++ b/metrique/src/lib.rs
@@ -96,6 +96,11 @@ use metrique_writer_core::EntrySink;
 use std::fmt::Debug;
 use std::sync::Arc;
 
+#[cfg(all(shuttle, feature = "_shuttle"))]
+use shuttle::sync::Mutex;
+#[cfg(not(all(shuttle, feature = "_shuttle")))]
+use std::sync::Mutex;
+
 pub use metrique_core::{
     CloseValue, CloseValueRef, Counter, CounterGuard, InflectableEntry, NameStyle,
     OwnedCounterGuard,
@@ -180,7 +185,9 @@ pub type DefaultSink = metrique_writer_core::sink::BoxEntrySink;
 /// A wrapper that appends and closes an entry when dropped.
 ///
 /// This struct holds a metric entry and a sink. When the struct is dropped,
-/// it closes the entry and appends it to the sink.
+/// it closes the entry and appends it to the sink. Construction is
+/// allocation-free; heap allocation is deferred until [`flush_guard`](Self::flush_guard)
+/// or [`force_flush_guard`](Self::force_flush_guard) is called.
 ///
 /// The [`metrics`] macro generates a type alias to this type
 /// named `<metric struct name>Guard`, you should normally mention that instead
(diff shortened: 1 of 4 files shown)


metrique/src/lib.rs
```diff
@@ -510,6 +514,102 @@ pub fn append_and_close<
...
+        self.sink.append(RootEntry::new(entry.close()));
+    }
+}
+
+impl<E: CloseEntry + Debug, S: EntrySink<RootMetric<E>> + Debug> Debug
+    for NoAllocAppendOnDrop<E, S>
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("NoAllocAppendOnDrop")
+            .field("value", &**self)
```
rcoh: this would be a bit clearer if we either did `self.value` or `self.deref()`

metrique/tests/no-alloc-append.rs
```diff
@@ -0,0 +1,86 @@
...
+    guard.value = 77;
+    guard.emit();
+
+    let entries = inspector.entries();
+    check!(entries.len() == 1);
+    check!(entries[0].metrics["value"] == 77);
+}
+
+#[test]
+fn deref_mut_mutations_reflected() {
```
rcoh: a test of the debug impl would be good

rcoh (commented):
I'm fine with this. But another idea could be to do an optimization like:

```
enum LazilyAllocated<E> {
   Direct(E),
   Parented(Parent<E)
}
```

then inside `AppendAndCloseOnDrop` we would hold `Mutex<LazilyAllocated>`

When you call `flush_guard` (which is very rare anyway!) we would promote it to parent.

Otherwise, we would keep it directly inline.

That saves an allocation for _everyone_ and the cost of the uncontended mutex in deref should be effectively 0.

Thoughts?

popematt:
Your suggestion is a much more elegant solution! (And even better, it avoids the naming problem I was having.) I'll do that instead.

popematt:
I've gone with a slightly different approach because the ownership/borrowing didn't work out when I looked into the `enum LazilyAllocated` approach.

We can't move the entry to the heap on promotion because `Deref<Target=E>` would break. The user holds a borrow into the guard, and relocating `E` would invalidate it. Instead, the entry stays inline for its entire lifetime; only the emission-coordination state (`PendingEmit`) gets lazily heap-allocated behind `Parent`, and the entry transfers into it at `Drop` time when no references can exist.

metrique/src/lib.rs
```diff
@@ -211,20 +218,49 @@ pub type DefaultSink = metrique_writer_core::sink::BoxEntrySink;
 /// # }
 /// ```
 pub struct AppendAndCloseOnDrop<E: CloseEntry, S: EntrySink<RootMetric<E>>> {
-    inner: Parent<AppendAndCloseOnDropInner<E, S>>,
+    inner: Option<(E, S)>,
+    promoted: Mutex<Option<Parent<PendingEmit<E, S>>>>,
```
rcoh: we could potentially constrain this unsafety in something like:

```
struct ContrainedOption<T> {
   inner: Option<T>
}

impl ConstrainedOption<T: Send> { fn set(&mut self, t: ...) }

unsafe impl Send for ContrainedOption<T> {... } // can only be set when T is send. 
```

That way we make this pattern a primitive with a narrowly defined safety contract.

rcoh (commented):
I like the design, looks great. See one comment inline.

metrique/src/lib.rs
```diff
@@ -495,18 +515,71 @@ impl<E: CloseEntry, S: EntrySink<RootMetric<E>>> std::ops::Deref
...
-        inner: Parent::new(AppendAndCloseOnDropInner {
-            entry: Some(base),
-            sink,
-        }),
+        inner: Some((base, sink)),
+        promoted: LazyPromotionSlot::new(),
+    }
+}
+
+/// A `Mutex<Option<T>>` wrapper that is unconditionally `Sync` by
```
rcoh: this code as written is sound but the safety justifications are not quite right.

`!Send` T actually _can_ enter the slot via `get_mut`. Everything is still sound because no `&self` method reveals `T`.

But since that is the key invariant, and its not documented, this code could evolve to easily break that safety invariant by accident.

popematt: Good catch. Fixed it.

rcoh (approved):

PR #363, opened 2026-08-10
feat(metrique-writer-core): add metric quantization

MikeDombo:
## What this adds

Opt-in quantization that reduces the precision of metric values as they are written, trading a precisely bounded amount of accuracy for a much smaller set of distinct emitted values — which makes a metric stream both cheaper to store and substantially more compressible.

A quantizer retains the leading `N` significant bits of each observation and clears the rest, equivalent to rounding to the nearest value representable with an `N`-bit significand. Because the discarded bits are always the least significant ones, the error is bounded *relative* to the value rather than in absolute terms, which suits quantities spanning many orders of magnitude such as latencies and byte sizes.

> [!NOTE]
> Opening as a **draft** because `CONTRIBUTING.md` asks that significant work be discussed in an issue first, and this adds new public API across four crates. Happy to split it up, move things, or rework the API shape — the pieces are independent. Please tell me if you'd prefer an issue first and I'll close this.

## API

The core is dependency-light and lives in `metrique-writer-core` so it can be reused without taking on the value model. `metrique-writer-core` was chosen as the home because it is the lightest crate that could host it (`metrique-core` and `metrique-util` both depend on it) and it already owns `Value`/`ValueWriter`/`Observation`.

```rust
pub struct SignificantBits(u8);          // 1..=52, DEFAULT = 8
pub enum Rounding { Floor, Ceil, Midpoint }   // Default: Midpoint
pub struct Quantizer { /* .. */ }        // quantize_u64, quantize_f64
pub struct Bits<const N: u8, R = rounding::Midpoint>;  // settings as a type
```

Three ways to apply it, in increasing order of blast radius:

```rust
// 1. One field, declared in the struct
struct RequestMetrics {
    #[metrics(quantize(bits = 8))]                      // within 0.390625%
    latency_nanos: u64,
    #[metrics(quantize(bits = 4, rounding = floor))]    // within 6.25%, never over
    payload_bytes: u64,
    request_count: u64,                                 // left exact
}

// 2. One value, when the bit count comes from configuration
let latency = 1_234_567u64.quantized(bits, Rounding::Midpoint);

// 3. A whole stream, without touching field declarations
Emf::all_validations("MyApp".into(), vec![vec![]])
    .output_to(out)
    .with_quantization(QuantizationPolicy::for_metrics(quantizer, ["LatencyNanos"]))
```

## Error bounds

Every bound is a power of two, so these are exact rather than rounded. `SignificantBits::max_relative_error` returns them programmatically.

| significant bits | `Floor` / `Ceil` (`< 2^(1-N)`) | `Midpoint` (`<= 2^-N`) | exact below |
|---|---|---|---|
| 1 | 100% | 50% | 2 |
| 2 | 50% | 25% | 4 |
| 3 | 25% | 12.5% | 8 |
| 4 | 12.5% | 6.25% | 16 |
| 6 | 3.125% | 1.5625% | 64 |
| 8 (default) | 0.78125% | 0.390625% | 256 |
| 11 | 0.09765625% | 0.048828125% | 2048 |
| 16 | 0.0030517578125% | 0.00152587890625% | 65536 |

## Design decisions worth reviewing

**`Midpoint` is the default.** It maps a whole bucket onto a single representative just as `Floor` does, so it produces the same number of distinct values and the same value-histogram entropy, while halving the worst-case error and largely cancelling the directional bias the one-sided modes introduce into aggregates. Measured on a 300k-value corpus of synthetic latencies, sizes, and counters, `Floor` and `Midpoint` produced identical distinct counts and identical entropy at every bit width (at 8 bits: both 3192 distinct, 10.519 bits/value); `Midpoint`'s gzipped output was ~1.6% larger because floor representatives carry more trailing decimal zeros.

**The 52-bit ceiling.** Two of three `Observation` variants carry values as `f64`, whose significand is 53 bits. Allowing more than 52 would silently no-op on those variants while still altering `Unsigned(u64)`, making the documented bound untrue for some values but not others. Out-of-range input is rejected rather than clamped.

**`Midpoint` is applied unconditionally, `Ceil` is not.** Returning a lattice-exact value untouched under `Midpoint` would reduce its error to zero but make the bucket produce two distinct outputs instead of one, working against the goal. `Ceil` must special-case them, both to keep its error tight and to stay idempotent — without it, `8 → 12 → 16` diverges.

**Timestamps are excluded structurally, not by name.** `EntryWriter` exposes `timestamp()`, `value()`, and `config()`; timestamps never flow through `value()`, so a decorator intercepting only values provably cannot reach one. A test asserts this holds even under a policy matching every name.

**`Repeated.occurrences` is always left exact.** Formats divide by it to recover an average, so quantizing it would corrupt that average by an amount unrelated to the error bound. Only `total` is reduced.

**The stream policy has no "quantize everything" constructor.** A stream carries counts that must reconcile and identifiers that happen to be numeric, so the metrics to quantize have to be named.

**Subnormal `f64` values pass through untouched.** A subnormal already carries fewer than 53 significant bits, so masking could erase it outright and report a 100% error. Passing them through keeps the documented bound true of every value the function actually modifies.

## Properties, all covered by tests

- **Idempotent** in every mode: `q(q(v)) == q(v)`.
- **Monotonic**: `a <= b` implies `q(a) <= q(b)`.
- **Never panics**: arithmetic saturates at `u64::MAX` / `f64::MAX`, which still satisfies the `Ceil` guarantee.
- Values below `exact_below()` are returned unchanged, so small counters are never disturbed.
- `NaN`, the infinities, and both signed zeros pass through unchanged.
- On negative values the one-sided modes swap, so each mode's promise holds against the true value rather than its magnitude.
- `quantize_u64` and `quantize_f64` agree exactly at and above `exact_below()`.

## Testing

- 68 unit tests on the quantizer, including a cross-check against a deliberately naive reference implementation (division loop and `u128` arithmetic, sharing no arithmetic with the shift-and-mask path) across every bit width in `1..=52`.
- 27 unit tests on the value wrapper and writer, including that a list-valued field still emits a native array with each element quantized — the regression `ValueWriter`'s trait docs warn about when `values()` isn't forwarded.
- 21 doc tests, plus `compile_fail` tests for `Bits<0>` and `Bits<53>`.
- 14 integration tests across `metrique-writer` and `metrique`.
- A `trybuild` case covering 9 macro diagnostics, including a pointed message when someone reaches for `digits = 2` instead of `bits = 8`.
- Full workspace suite passes (1027 tests). `cargo clippy --workspace --all-features --all-targets` reports zero warnings in any file this PR touches, and `RUSTDOCFLAGS="-D warnings" cargo doc` is clean. `cargo fmt --all` applied.

## Review
- The primary file to review is: https://github.com/MikeDombo/metrique/blob/b9a8ef49a40f45636ecf4270de31b910e7781e19/metrique-writer-core/src/quantize.rs

### DIFF
--- a/metrique-writer/src/entry/quantized.rs
+++ b/metrique-writer/src/entry/quantized.rs
@@ -1,0 +1,213 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Applying a quantizer to a whole stream of entries.
+
+use std::{
+    borrow::Cow,
+    collections::HashSet,
+    fmt,
+    ops::{Deref, DerefMut},
+    sync::Arc,
+    time::SystemTime,
+};
+
+use metrique_writer_core::{
+    Entry, EntryConfig, EntryWriter, Value, quantize::Quantizer, value::Quantized,
+};
+
+use crate::CowStr;
+
+/// Chooses which metrics a [`QuantizationPolicy`] applies to.
+///
+/// Kept private so that new selection strategies can be added without a breaking change; build
+/// one with [`QuantizationPolicy::for_metrics`] or [`QuantizationPolicy::matching`].
+#[derive(Clone)]
+enum MetricFilter {
+    Names(HashSet<CowStr>),
+    Predicate(Arc<dyn Fn(&str) -> bool + Send + Sync>),
+}
+
+impl fmt::Debug for MetricFilter {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            MetricFilter::Names(names) => f.debug_tuple("Names").field(names).finish(),
+            MetricFilter::Predicate(_) => f.write_str("Predicate(..)"),
+        }
+    }
+}
+
+/// Which metrics to quantize on a stream, and how.
+///
+/// A policy pairs a [`Quantizer`] with a filter naming the metrics it applies to. There is
+/// deliberately **no** "quantize everything" constructor: a stream carries counts that must
+/// reconcile exactly, identifiers that happen to be numeric, and other values whose precision
+/// is load-bearing. Reducing those silently would be a bug, so the set of metrics to quantize
+/// has to be stated.
+///
+/// Timestamps need no filtering. An entry's timestamp travels through
+/// [`EntryWriter::timestamp`], never through [`EntryWriter::value`], so a decorator that only
+/// intercepts values cannot reach it.
+///
+/// # Examples
+///
+/// Quantize two named latency metrics and nothing else:
+///
+/// ```
+/// use metrique_writer::quantize::{Quantizer, Rounding, SignificantBits};
+/// use metrique_writer::stream::QuantizationPolicy;
+///
+/// let quantizer = Quantizer::new(SignificantBits::new(8).unwrap(), Rounding::Midpoint);
+/// let policy = QuantizationPolicy::for_metrics(quantizer, ["LatencyNanos", "DownstreamNanos"]);
+///
+/// assert!(policy.applies_to("LatencyNanos"));
+/// assert!(!policy.applies_to("RequestCount"));
+/// ```
+///
+/// Or select by a predicate, when the metric names follow a convention:
+///
+/// ```
+/// use metrique_writer::quantize::Quantizer;
+/// use metrique_writer::stream::QuantizationPolicy;
+///
+/// let policy = QuantizationPolicy::matching(Quantizer::default(), |name| {
+///     name.ends_with("Nanos") || name.ends_with("Bytes")
+/// });
+///
+/// assert!(policy.applies_to("LatencyNanos"));
+/// assert!(policy.applies_to("PayloadBytes"));
+/// assert!(!policy.applies_to("RequestCount"));
+/// ```
+#[derive(Clone, Debug)]
+pub struct QuantizationPolicy {
+    quantizer: Quantizer,
+    filter: MetricFilter,
+}
+
+impl QuantizationPolicy {
+    /// Quantize exactly the named metrics, matched by their emitted name.
+    ///
+    /// Names are compared after any renaming the entry applies, so they should match what
+    /// appears in the output.
+    pub fn for_metrics(
+        quantizer: Quantizer,
+        names: impl IntoIterator<Item = impl Into<CowStr>>,
+    ) -> Self {
+        Self {
+            quantizer,
+            filter: MetricFilter::Names(names.into_iter().map(Into::into).collect()),
+        }
+    }
+
+    /// Quantize the metrics whose emitted name satisfies `predicate`.
+    ///
+    /// The predicate is consulted once per value written, so it should be cheap.
+    pub fn matching(
+        quantizer: Quantizer,
+        predicate: impl Fn(&str) -> bool + Send + Sync + 'static,
+    ) -> Self {
+        Self {
+            quantizer,
+            filter: MetricFilter::Predicate(Arc::new(predicate)),
+        }
+    }
+
+    /// The quantizer this policy applies.
+    pub fn quantizer(&self) -> Quantizer {
+        self.quantizer
+    }
+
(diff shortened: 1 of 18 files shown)


metrique-writer/src/stream.rs
```diff
@@ -227,6 +278,29 @@ impl<S: EntryIoStream, const N: usize> EntryIoStream for MergeGlobalDimensions<S
...
+pub struct Quantize<S> {
+    pub(crate) stream: S,
+    pub(crate) policy: QuantizationPolicy,
+}
+
+impl<S: EntryIoStream> EntryIoStream for Quantize<S> {
+    fn next(&mut self, entry: &impl Entry) -> Result<(), IoStreamError> {
+        self.stream.next(&crate::entry::QuantizedEntry::new(
+            entry,
+            self.policy.clone(),
```
jasgin: Are we concerned about performance cost of this if this is per request? Maybe an Arc would make sense

metrique-writer/src/entry/quantized.rs
```diff
@@ -0,0 +1,213 @@
...
+                // there is no path by which this decorator could do so: timestamps do not go
+                // through `value()`.
+                self.writer.timestamp(timestamp);
+            }
+
+            fn value(&mut self, name: impl Into<Cow<'a, str>>, value: &(impl Value + ?Sized)) {
+                let name: Cow<'a, str> = name.into();
+                if self.policy.applies_to(&name) {
+                    self.writer
+                        .value(name, &Quantized::new(value, self.policy.quantizer()));
```
jasgin, line 192: What happens if a value has already been quantized with #[metrics(quantize(...))]

MikeDombo: Requantizing a value will not change it at all -- as long as it is quantized with the same settings. If the settings are different then it could have precision reduced further. If the second quantization is to more significant digits then the value would not be changed. But if the second quantization is to fewer digits then we'd move it to a different value.

metrique-writer/src/stream.rs
MikeDombo: Yeah that's reasonable, can do an Arc here.

```
  ┌─────────────┬───────────────────────────────┬─────────────────────────┬───────────┐
  │ filter size │ deep clone, &'static str keys │ deep clone, String keys │ Arc clone │
  ├─────────────┼───────────────────────────────┼─────────────────────────┼───────────┤
  │ 2           │ 34.0 ns                       │ 58.7 ns                 │ 4.1 ns    │
  ├─────────────┼───────────────────────────────┼─────────────────────────┼───────────┤
  │ 10          │ 45.9 ns                       │ 236.8 ns                │ 4.1 ns    │
  ├─────────────┼───────────────────────────────┼─────────────────────────┼───────────┤
  │ 100         │ 232.0 ns                      │ 2081.6 ns               │ 4.2 ns    │
  └─────────────┴───────────────────────────────┴─────────────────────────┴───────────┘

```

MikeDombo: Updated in a new commit.

jasgin:
Overall LGTM. I think Midpoint makes sense as a default for the bias cancellation. Though one could argue for Floor to be the default due to the 1.6% larger output of Midpoint, since the reason someone would use this feature is to maximally reduce cost.

> Midpoint is applied unconditionally, Ceil is not.

I don't know if I fully understand what it means for Ceil to have special case / the significance of it, if you'd be able to elaborate a bit.

The decisions for the others seem sound to me though.

MikeDombo:
> Though one could argue for Floor to be the default due to the 1.6% larger output of Midpoint

That output size will differ based on the exact compressor used, zstd may perform a bit better than gzip here. And we're talking about 1% of compressed size difference where the overall compression savings by applying this method at all are 10-20% depending on the log. So overall it is a small effect and I think the better error bars are worth it.

> I don't know if I fully understand what it means for Ceil to have special case / the significance of it, if you'd be able to elaborate a bit.

Ceil is special because if a value lands exactly on a bucket boundary it would be rounded up to the next bucket without it. eg. 8 would become 12 because 8 is the bucket boundary and 12 is the top of the next bucket as an example, so that blows up the error. Instead we will handle that and emit it as exactly the same value with 0 error

jasgin (approved):

Issue #367, opened 2026-08-14
feat: in-memory drainable MakeWriter for pull-based metric delivery


Elvrarin:
## Problem

Every destination output_to_makewriter accepts is push-based (rotating file, stdout, TCP socket). On a host whose only egress is a fetcher that pulls on demand — there's no push target to point at.

## Proposal

Add InMemoryMakeWriter: an unbounded, cheaply-cloneable MakeWriter that buffers formatted records in memory and lets a fetcher drain() them on demand. It bridges metrique's push-only writer pipeline to a pull-based transport while reusing the full formatting stack (Format + BackgroundQueue/FlushImmediately) unchanged.

Because output_to_makewriter calls make_writer() once per entry, holding the buffer lock for the writer's lifetime makes each entry's writes atomic — a concurrent drain() only ever observes whole, newline-terminated records.

Gated behind tracing-subscriber-03, matching output_to_makewriter itself.

## Why MakeWriter and not EntryIoStream

Unlike the async egress backends in #225, this isn't a new egress backend — it's an in-memory destination for the existing one. Sitting at the MakeWriter layer reuses FormattedMakeWriterEntryIoStream as-is and gets the per-entry record-atomicity guarantee for free. Open to moving it to an EntryIoStream if preferred.

## Caveat

Unbounded (nothing evicted until drained); a fetcher that stops pulling grows the buffer without limit. Documented; a bound could be added later.

## Related

- #225 (async EntryIoStream push backends) — this is the pull counterpart
- #240 / #275 (pending_sink) — in-memory entry buffering, but for deferred attachment

I have a implementation + tests already on feat/in-memory-makewriter in my fork.

rcoh:
fixed in #368

PR #368, opened 2026-08-14
feat(metrique-writer): add in-memory MakeWriter for pull-based delivery

Elvrarin:
📬 *Issue #, if available:*
https://github.com/awslabs/metrique/issues/367

✍️ *Description of changes:*
All existing `output_to_makewriter` destinations push (file, stdout, socket). `InMemoryMakeWriter` buffers formatted records in memory for a fetcher to `drain()` on demand, and holds the buffer lock per entry so drains only ever observe whole records. Gated behind `tracing-subscriber-03`.

### DIFF
--- a/metrique-writer/src/in_memory.rs
+++ b/metrique-writer/src/in_memory.rs
@@ -75,10 +75,11 @@
 //!     .unwrap();
 //!
 //! // The fetcher pulls the formatted EMF records on demand.
-//! let records = buffer.drain();
-//! assert!(!records.is_empty());
+//! let drained = buffer.drain();
+//! assert!(!drained.bytes.is_empty());
+//! assert_eq!(drained.records_dropped, 0);
 //! // A second drain returns nothing — the buffer was emptied.
-//! assert!(buffer.drain().is_empty());
+//! assert!(buffer.drain().bytes.is_empty());
 //! ```
 
 use std::{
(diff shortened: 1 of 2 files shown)


metrique-writer/src/in_memory.rs
```diff
@@ -0,0 +1,289 @@
...
+//! calls [`MakeWriter::make_writer`] exactly once per entry and formats the
+//! whole entry into the returned writer (a format may issue *several*
+//! [`io::Write::write`] calls for one entry). [`InMemoryMakeWriter`] holds the
+//! buffer lock for that writer's entire lifetime, so every write for an entry
+//! happens under a single lock hold released only at the entry boundary. A
+//! concurrent [`drain`](InMemoryMakeWriter::drain) therefore can only run
+//! *between* entries and always observes whole records — no partial records, so
+//! no reassembly is needed by the fetcher.
+//!
+//! # Unboundedness
```
jasgin: Is there a reason this is intentionally unbounded? I'm thinking it may make sense to follow the prior art in metrique-writer/src/sink/background.rs, which uses a ring buffer and drop-oldest semantics.

metrique-writer/src/in_memory.rs
```diff
@@ -0,0 +1,289 @@
...
+
+impl<'a> MakeWriter<'a> for InMemoryMakeWriter {
+    type Writer = InMemoryWriter<'a>;
+
+    /// Acquire the buffer lock and hold it for the returned writer's lifetime,
+    /// making all of one entry's writes atomic with respect to
+    /// [`drain`](InMemoryMakeWriter::drain).
+    fn make_writer(&'a self) -> Self::Writer {
+        InMemoryWriter {
+            guard: self.inner.lock().unwrap(),
```
jasgin: I'd be worried about poisoning and performance since we hold the lock during the whole format. Can we use a second local/scratch buffer and only commit the finished record under a brief lock?

metrique-writer/src/in_memory.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// The number of bytes currently buffered across all pending records.
+    pub fn len(&self) -> usize {
+        self.inner.lock().unwrap().len()
+    }
+
+    /// Whether no bytes are currently buffered.
+    pub fn is_empty(&self) -> bool {
+        self.inner.lock().unwrap().is_empty()
+    }
```
jasgin: I think these would cause race conditions

metrique-writer/src/in_memory.rs
```diff
@@ -0,0 +1,494 @@
...
+
+use std::{
+    io::{self, Write},
+    sync::{Arc, Mutex},
+};
+
+use tracing_subscriber::fmt::MakeWriter;
+
+/// Default maximum buffer capacity: 1 MiB.
+const DEFAULT_MAX_CAPACITY: usize = 1024 * 1024;
```
jasgin: Is 1 MiB the right default or is it a bit small. That'd be something like 1k entries at 1kb each, whereas BackgroundQueue has a default of 64 * 1024 entries.

metrique-writer/src/in_memory.rs
```diff
@@ -0,0 +1,494 @@
...
+                if inner.buf.len() - drop_up_to <= target {
+                    break;
+                }
+            }
+        }
+
+        if drop_up_to > 0 {
+            inner.buf.drain(..drop_up_to);
+            inner.records_dropped += dropped;
+        }
```
jasgin, line 287: I think having rate limited error logging for when metrics are dropped probably makes sense

metrique-writer/src/in_memory.rs
Elvrarin: changed to 8MiB

jasgin (approved):

PR #369, opened 2026-08-16
Improve subfield_owned diagnostics with a marker type

### LINKED ISSUE #164: Insert marker type for `#[metrics(subfield)]`entries that need `#[metrics(subfield_owned)]`
Context: https://github.com/awslabs/metrique/pull/158#discussion_r2658420964

One of the more confusing macro situations is a case where you have a subfield that needs to `CloseValue` by value rather than by ref. In this case, you need `#[metrics(subfield_owned)]`.

I inserted at least a small diagnostic that unconditionally recommends this in: https://github.com/awslabs/metrique/pull/162/commits/afbd2cc61affd9377ea8ec471a4c7d76d8b8191c

But, another option is to have a implementation bridge that passes through a dummy type in cases that specifically need `#[metrics(subfield_owned)]`. That can show up in the diagnostic and be named in a way that explicitly states the problem.

@rcoh implemented a similar pattern in his aggregation macro PR via `IfYouSeeThisUseAggregateOwned`.

danielgaskins:
📬 *Issue #, if available:*

Fixes #164

✍️ *Description of changes:*

Some field types can only be closed by value. If one of these fields uses `#[metrics(subfield)]`, the user needs to change it to `#[metrics(subfield_owned)]`.

Previously, Metrique suggested `subfield_owned` for every `CloseValue` error, including unrelated errors. This change limits that clue to errors produced by borrowed subfields. In those cases, the compiler diagnostic now includes `IfYouSeeThisUseSubfieldOwned`.

This does not change the behavior of code that already compiles. It only improves compiler diagnostics.

Tests cover the diagnostic for both structs and enums.

Testing:

- `CARGO_BUILD_JOBS=1 NEXTEST_TEST_THREADS=2 RUST_TEST_THREADS=2 ./scripts/ci-local.sh`

### DIFF
--- a/metrique-core/src/lib.rs
+++ b/metrique-core/src/lib.rs
@@ -147,8 +147,21 @@ pub trait CloseValue {
     fn close(self) -> Self::Closed;
 }
 
-/// Diagnostic bridge for `#[metrics(subfield)]` fields that only close by value.
-#[doc(hidden)]
+/// A diagnostic marker for metric fields that must be closed by value.
+///
+/// `#[metrics(subfield)]` closes fields by reference. When a field implements
+/// [`CloseValue`] only by value, compiler errors mention this type to point to
+/// `#[metrics(subfield_owned)]` as the alternative.
+///
+/// For example, a subfield containing a `String` must close by value:
+///
+/// ```
+/// # use metrique::unit_of_work::metrics;
+/// #[metrics(subfield_owned)]
+/// struct ChildMetrics {
+///     field: String,
+/// }
+/// ```
 pub struct IfYouSeeThisUseSubfieldOwned<T>(pub T);
 
 impl<T: CloseValue> CloseValue for IfYouSeeThisUseSubfieldOwned<T> {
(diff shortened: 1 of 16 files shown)


metrique-core/src/lib.rs
```diff
@@ -148,6 +147,18 @@ pub trait CloseValue {
     fn close(self) -> Self::Closed;
 }
 
+/// Diagnostic bridge for `#[metrics(subfield)]` fields that only close by value.
+#[doc(hidden)]
```
rcoh: making this doc unhidden and putting some docs here with an example might actually be helpful

rcoh (commented):
Clever! One inline comment.

metrique-core/src/lib.rs
danielgaskins: Made the marker visible in rustdoc and added an example showing when to use `#[metrics(subfield_owned)]`.

rcoh (approved):

PR #375, opened 2026-08-20
Warn when append-on-drop guards are unused

### LINKED ISSUE #364: Should `AppendAndCloseOnDrop` have `#[must_use]`?
Constructing an `AppendAndCloseOnDrop` guard without binding it to a variable silently emits a metric with uninitialized data, I think.

```rust
MyMetrics { operation: "foo" }.append_on_drop(ServiceMetrics::sink());
// ^ emits immediately; timers never started, counters never incremented
```

The metric is emitted, but with none of the meaningful fields populated. At best, this performs useless work, and at worst, this produces silent data loss.  Adding `#[must_use = "guard will emit immediately if not bound to a variable"]` to the struct would catch this at compile time.

danielgaskins:
## Summary

- mark `AppendAndCloseOnDrop` as `must_use` so unused guards produce a warning
- make existing immediate emissions explicit with `drop`
- add a compile-fail test for the warning and its guidance

Fixes #364

## Testing

- `cargo +1.91 nextest run`
- `cargo +1.91.0 test -p metrique --test ui`
- `cargo +1.91 clippy --workspace --all-features -- -D warnings`
- `cargo +nightly fmt --all -- --check`

### DIFF
--- a/metrique/src/lib.rs
+++ b/metrique/src/lib.rs
@@ -219,6 +219,7 @@ pub type DefaultSink = metrique_writer_core::sink::BoxEntrySink;
 /// // When `metrics` is dropped, it will be closed and appended to the sink
 /// # }
 /// ```
+#[must_use = "use `drop(metric.append_on_drop(sink))` to emit immediately, or `let _metric = metric.append_on_drop(sink);` to emit at the end of the current scope"]
 pub struct AppendAndCloseOnDrop<E: CloseEntry, S: EntrySink<RootMetric<E>>> {
     inner: Option<(E, S)>,
     promoted: LazyPromotionSlot<Parent<PendingEmit<E, S>>>,
--- a/metrique-aggregation/examples/histogram.rs
+++ b/metrique-aggregation/examples/histogram.rs
@@ -32,7 +32,7 @@ fn main() {
     metrics.size.add_value(2048u32);
     metrics.size.add_value(2048u32);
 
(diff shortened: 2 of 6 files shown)


metrique/src/lib.rs
```diff
@@ -217,6 +217,7 @@ pub type DefaultSink = metrique_writer_core::sink::BoxEntrySink;
 /// // When `metrics` is dropped, it will be closed and appended to the sink
 /// # }
 /// ```
+#[must_use = "guard will emit immediately if not bound to a variable"]
```
rcoh, line 220: I think this PR is reasonable, but we need a clearer error message to explain what to do. The error should say:

"If you want to emit the metric immediately _do X_. If you want to emit the entry at the end of the current scope, do `let _metric = ...."

Or something along those lines. I don't think guard will emit immediately is going to be a helpful error to see.

danielgaskins: Sure thing, I updated the diagnostic to spell out both options: use `drop(metric.append_on_drop(sink))` to emit immediately, or bind it with `let _metric = metric.append_on_drop(sink);` to emit at the end of the current scope. The Rust 1.91.0 UI suite and workspace Clippy both pass.

danielgaskins:
Hi @rcoh, could you take another look when you have a chance? The updated diagnostic now covers both options you suggested.

rcoh (approved):

PR #378, opened 2026-08-24
Migrate inflector to heck

Mark-Simulacrum:
✍️ *Description of changes:*

This migrates from the unmaintained/archived inflector dependencies to heck, which is at least not archived and more widely used (e.g., by clap's derive macro). It is a breaking change for (probably rare?) keys, so we'll want to discuss whether we're OK with that. One alternative is to inline inflector and patch the stale dependencies (dropping regex + lazy-static). My personal view is that the new metric names are better than the old ones, but obviously backwards compatibility matters too.

See individual commits for diffs (snapshot tests added in first commit, updated in the second one) and more discussion.


Mark-Simulacrum:
Discussed offline, we think it makes sense to avoid changes in behavior as an interim step and reproduce the inflector behavior inside metrique directly. Will work on a change to this PR to do that.

PR #380, opened 2026-08-24
fix(metrique-aggregation): implement FlushableSink for MutexSink

popematt:
📬 *Issue #, if available:*

None.

✍️ *Description of changes:*

`MutexSink` implements `RootSink` but not `FlushableSink`, so a `KeyedAggregator` behind a `MutexSink` accepted entries forever but silently emitted nothing, and dropping the sink didn't flush it either.

An `Aggregate` was unaffected, because it gets drained through `CloseValue`, which `MutexSink` already forwards. `KeyedAggregator` has no `CloseValue` impl, so flushing was its only exit path and there wasn't one.

Changes:

- `impl FlushableSink for MutexSink<Inner> where Inner: FlushableSink`, forwarding under the lock. The lock uses `.unwrap()`, matching the neighbouring `RootSink::merge`
- `tests/mutex_sink.rs` covering the flush path and flushing through a clone, since `MutexSink` is an `Arc` handle. Both tests fail with `E0599: no method named flush` without the impl
- `README.md`: the "Choosing between WorkerSink and MutexSink" entry said `MutexSink` "does not support automatic flushing", which reads as though manual flushing was available. It now says there's no timer and you should call `FlushableSink` yourself when pairing it with a `KeyedAggregator`, and the section above explains why an `Aggregate` doesn't need one
- `CHANGELOG.md` entry under `Unreleased`


metrique-aggregation/README.md
```diff
@@ -145,7 +147,7 @@ async fn setup_queue_processor() {
 
 **Choosing between WorkerSink and MutexSink:**
 
-- **[`MutexSink`]** - Use when you have inputs from a smaller number of threads. Great for supporting `close_and_merge` with embedded metrics. Currently does not support automatic flushing.
+- **[`MutexSink`]** - Use when you have inputs from a smaller number of threads. Great for supporting `close_and_merge` with embedded metrics. Does not flush on a timer, so call [`FlushableSink`] yourself when combining it with [`KeyedAggregator`].
```
rcoh: should this say `FlushableSink::flush`?

rcoh (approved):

metrique-aggregation/README.md
popematt: Probably.

popematt:
```suggestion
- **[`MutexSink`]** - Use when you have inputs from a smaller number of threads. Great for supporting `close_and_merge` with embedded metrics. Does not flush on a timer, so call [`FlushableSink::flush`] yourself when combining it with [`KeyedAggregator`].
```

rcoh (approved):

PR #383, opened 2026-08-26
feat(metrique-macro): identity CloseValue for generated entries

### LINKED ISSUE #382: Generate an identity `CloseValue` impl for generated entry types
### Problem

A `#[metrics]` struct `Foo` generates a closed entry `FooEntry`, which has no `CloseValue` impl of its own. So an *already-closed* entry used as a field of another `#[metrics]` struct has to be marked `#[metrics(no_close)]`, because the parent's close step has nothing to call on it. `no_close` has sharp edges (it is invisible to `#[aggregate]`, and it fails to compile on `#[metrics(subfield)]` fields) so relying on it for a routine case is worth avoiding.

### Proposal

Generate an identity `CloseValue` impl for every generated entry type:

```rust
impl CloseValue for FooEntry {
    type Closed = FooEntry;
    fn close(self) -> Self { self }
}
```

Then a closed entry is a valid closing field with no attribute:

```rust
struct Parent {
    child: FooEntry,          // was: #[metrics(no_close)] child: FooEntry
    children: Vec<FooEntry>,  // closes elementwise, also no attribute
}
```

and the closed type can be named without spelling `FooEntry`, via `<Foo as CloseValue>::Closed` or an alias over it.

This is a small change. I was able to prototype this with only a handful of lines of codegen, although there is some test churn in the macro-expansion snapshots needing to be updated. The entry is a local type, so there is no coherence conflict.

Two properties matter beyond the ergonomics:

- **It does not enable sharing.** `Arc<FooEntry>` without `no_close` still fails, because that path needs `&FooEntry: CloseValue`, which identity close does not provide (it would require `Clone`, which would copy the payload). So this change removes `no_close` from the owned case without opening a shared-mutation footgun.
- **It keeps two `no_close` defects off the common path.** The `#[aggregate]`-ignores-`no_close` case and the `subfield` + `no_close` compile error both stop mattering for fields that no longer need `no_close`.

I have this working locally and am willing to open a PR.

Thoughts?

popematt:
📬 *Issue #, if available:*

Closes #382

✍️ *Description of changes:*

Generated `*Entry` types now have an identity implementation of `CloseValue`.

This eliminates a sharp edge that previously required `no_close` to workaround. Now it just works.

In the one build-failure UI test case that I modified, there is a new duplication of an error. I don't think this is a blocker because the same error is already repeated a few times. (What's one more?) Also worth noting that it seems to be fixed by moving to the 1.91.1 toolchain.

### DIFF
--- a/metrique-macro/src/lib.rs
+++ b/metrique-macro/src/lib.rs
@@ -111,7 +111,7 @@ impl MacroError {
 /// | `exact_prefix` | String | Adds a prefix to flattened entries without inflection | `#[metrics(flatten, exact_prefix="API_")]` |
 /// | `flatten` | Flag | Flattens nested `CloseEntry` metric structs | `#[metrics(flatten)]` |
 /// | `flatten_entry` | Flag | Flattens nested `CloseValue<Closed: Entry>` metric structs, with no prefix or inflection | `#[metrics(flatten_entry)]` |
-/// | `no_close` | Flag | Use the entry directly instead of closing it | `#[metrics(no_close)]` |
+/// | `no_close` | Flag | Use a value that is already in closed form directly, instead of calling `CloseValue::close`. Needed for foreign or hand-written types that implement `Entry` (via `flatten_entry`) or a value formatter's input (via `format`) but not `CloseValue`. Generated `#[metrics]` entries implement an identity `CloseValue`, so they do not need this. | `#[metrics(no_close)]` |
 /// | `ignore` | Flag | Excludes the field from metrics | `#[metrics(ignore)]` |
 /// | `flags` | Path(s) | Applies a flag to this field for format and sink usage. Use `skip(T)` to explicitly exclude. | `#[metrics(flags(my_crate::flags::Export, skip(OtherFlag)))]` |
 /// | `default_flags` | Path(s) | On a flatten field, applies flags to all child fields. Child field-level skips take precedence. | `#[metrics(flatten, default_flags(my_crate::flags::Export))]` |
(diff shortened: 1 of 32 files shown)


popematt:
I regenerated the snapshots with 1.91.1 but it should have been 1.91.0.

metrique-macro/src/lib.rs
```diff
@@ -111,7 +111,7 @@ impl MacroError {
 /// | `exact_prefix` | String | Adds a prefix to flattened entries without inflection | `#[metrics(flatten, exact_prefix="API_")]` |
 /// | `flatten` | Flag | Flattens nested `CloseEntry` metric structs | `#[metrics(flatten)]` |
 /// | `flatten_entry` | Flag | Flattens nested `CloseValue<Closed: Entry>` metric structs, with no prefix or inflection | `#[metrics(flatten_entry)]` |
-/// | `no_close` | Flag | Use the entry directly instead of closing it | `#[metrics(no_close)]` |
+/// | `no_close` | Flag | Use a value that is already in closed form directly, instead of calling `CloseValue::close`. Needed for foreign or hand-written types that implement `Entry` (via `flatten_entry`) or a value formatter's input (via `format`) but not `CloseValue`. Generated `#[metrics]` entries implement an identity `CloseValue`, so they do not need this. | `#[metrics(no_close)]` |
```
rcoh: might be worth writing slightly more to clarify here:

```
The `#[metrics]` macro generates a `<TypeName>Entry` struct. This macro also auto-generates an identity close value. Note that typically, you want to utilize the struct annotated with `#[metrics]`, _not_ the `Entry` struct.
```

that's not perfect either; but making it clear that there is another struct being generated is probably useful

metrique-macro/src/lib.rs
```diff
@@ -2099,13 +2099,26 @@ fn generate_close_value_impls(
         }
     };
 
+    // Identity `CloseValue` on the generated closed entry type. An already-closed
+    // entry is a valid closing field of another `#[metrics]` struct without needing
+    // `#[metrics(no_close)]`. See awslabs/metrique#382.
+    let identity_impl = quote!(
```
rcoh: this is actually technically a breaking change (and I found some actual code in the wild that will fail to compile with this change).

What do you think about making this off-by-default with a flag in the `#[metrics(...)]` macro to turn it on?

Once we fix the existing code in the wild, we could enable it by default

rcoh (changes requested):

metrique-macro/src/lib.rs
popematt: Sure, we can do that. Or we could just keep it off-by-default until the next major version release, since it is a breaking change.

rcoh: yeah also fine; depends on how useful it is

rcoh (approved):

PR #385, opened 2026-08-27
Strongly own case conversion code

Mark-Simulacrum:
✍️ *Description of changes:*

This pulls in the case conversion code into metrique-core rather than using third party libraries for it. The code is short (if dense), and this removes regex + lazy-static dependencies from these crates, hopefully helping speed up builds a little.

Likely best reviewed in separate commits. No functional changes to the output intended (assuming ASCII inputs, at least). As noted in the previous PR (https://github.com/awslabs/metrique/pull/378) it's not obvious that the produced output here is the one we want, but this *should* match inflector/str-inflector behavior and be something we can iterate on if we decide to make breaking changes.


rcoh (commented):
do you think we should add a `prop_test` that tests against the `inflector` as a dev-dep?

Mark-Simulacrum:
I did some local testing on ASCII inputs that seems to show this is equivalent. I can add those commits temporarily and then drop them; I don't think it makes sense to long-term maintain a dependency just for historical comparison.

rcoh:
yeah makes sense to me

rcoh:
I fuzzed this independently. There are some mismatches w/ unicode but that seems incredibly unlikely in practice.

rcoh (approved):

Issue #144, opened 2025-12-24
Add support for rule-based tail sampling


jlizen:
A frequent use case for cost or performance sensitive applications is to avoid publishing the superset of all logs or metrics, and rather publish ones that indicate issues or other "interesting" data points. Or, perhaps publish everything, but publish ones meeting a given condition at a higher sample rate than others.

For instance:
- requests with errors
- requests with latency > X threshold
- requests with some new feature enabled

Today we "sort of" support this via [congressional sampling](https://docs.rs/metrique/latest/metrique/writer/sample/struct.CongressSample.html). That would let you tag a field corresponding to error rates as a "sample group", so that you have a guarantee that you would at least see some entries with errors even if vastly outnumbered by happy requests.

But, that still forces you to treat the "key" as unit of decision making. And that means, you are also publishing sampled requests at the same rate for "not an error" when you want elevated rates for "is an error". And then, for latency thresholds, that is completely out the window.

This is particularly relevant for the new [aggregation support](https://github.com/awslabs/metrique/pull/112), where you only really want entries that you would use to debug issues, since you anyway are ensuring you fairly roll up overall performance in aggregated metrics.

Instead, it would be great to support [tail sampling](https://opentelemetry.io/blog/2022/tail-sampling/). This is essentially one or more rules that can be applied to the metric prior to closing it, that evaluates based on its full contents.

Ideally this could allow multiple policies that cascade, where the first one matching sets the sample rate (and it could fall back to regular or congressional sampling if none match).

A strawman API might be something like:

```
#[derive(Entry)]
struct MyMetrics {
    latency: Duration,
    #[entry(sample_group)]
    operation: &'static str,
    is_error: bool
}

let rule_1 = TailSamplingRule::builder()
                            .with_condition(|entry| entry.is_error)
                            .with_fixed_rate(1.0);

let rule_2 = TailSamplingRule::builder()
                            .with_condition(|entry| entry.latency > Duration::from_secs(1))
                            .with_fixed_entries_per_second(2);

Emf::all_validations("MyApp".into(), vec![vec![]])
        .with_sampling()
        .sample_by_tail_rules(vec![rule_1, rule_2]) // applied in precedence order, first one to match picks the rate
        .sample_by_congress_at_fixed_entries_per_second(1), // fall through if none match
       // you could also just turn off all emission for non-matching entries
        ....
```

PR #394, opened 2026-09-16
feat(aggregation): add TopNSink

quinnwerks:
This bounds memory while retaining the highest-scoring entries from each flush window for detailed inspection.

📬 *Issue #, if available:* 
#144 

✍️ *Description of changes:*
The simplest thing we can do to support granular metrics for analyzing tail issues is to add a new sink that 
aggregates the top N requests where "Top" pertains to a certain raw metric and "N" is the number highest scoring 
raw events that the user want to retain per flush. Added `TopNSink` with accompanying tests and docs.

This can be combined with an aggregate sink via `TeeSink` to both get service level aggregates while retaining granularity of metrics at the tail.

There is definitely more work to be done in this domain including:
1. An ability to exclude metrics with certain properties from the sample (e.g. exclude errors from latency aggregates)
2. TopN% based sampling

- [ x ] I confirm that I've made a best effort attempt to update all relevant documentation.
- [ x ] I confirm that my contribution is made under the terms of the Apache 2.0 license.


metrique-aggregation/examples/rolling_threshold.rs
```diff
@@ -0,0 +1,58 @@
...
+    for latency_ms in scores {
+        sink.merge(Request { latency_ms });
+    }
+    sink.flush();
+}
+
+fn main() {
+    let selected = CollectingSink::default();
+    let mut sink =
+        RollingThresholdSink::new(10, |request: &Request| request.latency_ms, selected.clone());
```
rcoh: recommend a builder instead of `new` here

Issue #366, opened 2026-08-13
Add example for emitting metrics from periodic background tasks (non-request unit-of-work)


jasgin:

rcoh:
This should follow the same pattern as the metrique-util reporters for tokio-metrics and sysinfo: https://docs.rs/metrique-util/latest/metrique_util/trait.AttachGlobalEntrySinkSysinfoExt.html#method.subscribe_sysinfo_metrics

PR #395, opened 2026-09-17
docs: add periodic background task metrics example

fmzbl:
📬 *Issue #, if available:*
#366 
✍️ *Description of changes:*
adds an example of a periodic background task as its own unit of work, a cache-refresh loop emitting one entry per tick, wired like the metrique-util reporters


metrique/examples/unit-of-work-background-task.rs
```diff
@@ -0,0 +1,227 @@
...
+        let (cancel_tx, mut cancel_rx) = oneshot::channel::<()>();
+        tokio::spawn(async move {
+            let mut ticks = tokio::time::interval(config.interval);
+            // A slow refresh should delay the next tick, not make the loop
+            // fire back-to-back trying to catch up.
+            ticks.set_missed_tick_behavior(MissedTickBehavior::Delay);
+            loop {
+                // Shutdown is awaited here and nowhere else: a refresh that has
+                // already started always finishes and emits its entry.
+                tokio::select! {
```
jasgin, line 157: Do we need to join the refresh task before dropping the handle, otherwise the in-flight refresh metric is not guaranteed to emit if the sink gets dropped, right?

Also could this create an unwarranted race if refresh takes longer than the tick? Maybe we should use biased with cancel at the top.

rcoh (approved):

PR #397, opened 2026-09-17
feat(metrique-util): request-scoped metrics pools

jasgin:
📬 *Issue #, if available:*
Implements https://github.com/awslabs/metrique/pull/357

- implement MetricsPool and cloneable MetricsPoolHandle for heterogeneous child metrics flattened into a parent entry
add explicit-handle and poll-scoped current-pool APIs with runtime casing, prefixes, close semantics, and rate-limited last-write-wins collision handling
- add an AWS SDK interceptor example with scoped collection, explicit injection, and global fallback
- add focused tests and Divan benchmarks

✍️ *Description of changes:*


metrique-util/examples/sdk-interceptor-metrics-pool.rs
```diff
@@ -0,0 +1,558 @@
...
+    // A per-call interceptor can instead receive the pool explicitly.
+    let _per_call_interceptor = SdkMetricsInterceptor::builder("CloudWatchLogs")
+        .qualifier("Logging")
+        .metrics_pool(request_metrics.metrics_pool.handle())
+        .build();
+
+    // In a service, the SDK client's `.send().await` runs inside this future.
+    let pool = request_metrics.metrics_pool.handle();
+    let request_metrics = with_metrics_pool(pool, async move {
+        // client.put_log_events().send().await?;
```
rcoh: you can have this depend on a real SDK then use a fake make http server so you actually emit "real" data

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+//! incompatible configuration can invalidate the parent record. Child sample
+//! groups are always ignored and cannot change the parent's sampling policy.
+//!
+//! # Collision policy
+//!
+//! Collisions are compared using fully inflected field names among children in
+//! the same pool. If two children collide, the later child wins and the earlier
+//! child is dropped in full, preventing fields from separate attempts from being
+//! combined into a record that never existed. If one child writes a name more
+//! than once, only its final write is emitted.
```
rcoh: I wonder if we should allow an auto "indexing" mode where multiple entries become `.0.key`, `.1.key` etc.

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+    RootEntry, SnakeCase, append_and_close,
+    concat::{EmptyConstStr, const_str_value},
+};
+use metrique_core::CloseEntry;
+use metrique_core::case_convert::CaseStyle;
+
+const DEFAULT_METRICS_POOL_CAPACITY: usize = 128;
+
+thread_local! {
+    static CURRENT_POOL: RefCell<Vec<MetricsPoolHandle>> = const { RefCell::new(Vec::new()) };
```
rcoh: a comment explaining why this is in a vec would be nice

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+impl RuntimeStyle {
+    fn for_ns<NS: NameStyle>() -> Self {
+        match NS::DESCRIPTOR_STYLE_INDEX {
+            1 => Self::PascalCase,
+            2 => Self::SnakeCase,
+            3 => Self::KebabCase,
+            _ => Self::Identity,
+        }
+    }
+
```
rcoh: I thought we already had this code somewhere. can you reuse it?

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+    fn with_segments<I>(self, segments: I) -> Self
+    where
+        I: IntoIterator<Item = &'static str>,
+    {
+        let mut identity = self.identity.to_string();
+        let mut pascal = self.pascal.to_string();
+        let mut snake = self.snake.to_string();
+        let mut kebab = self.kebab.to_string();
```
rcoh: this is going to allocate a lot of strings...

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+    ///
+    /// Panics if `capacity` is zero.
+    pub fn capacity(mut self, capacity: usize) -> Self {
+        assert!(
+            capacity > 0,
+            "MetricsPool capacity must be greater than zero"
+        );
+        self.capacity = capacity;
+        self
+    }
```
rcoh: we don't want an infinite capacity setting I guess?

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+impl<F: Future> Future for MetricsPoolScope<F> {
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        // Safety: once `self` is pinned, `future` is never moved again.
+        let this = unsafe { self.get_unchecked_mut() };
+        let _scope = ScopeGuard::install(this.pool.clone());
+        // Safety: `future` stays pinned for the lifetime of `self`.
+        unsafe { Pin::new_unchecked(&mut this.future) }.poll(cx)
+    }
```
rcoh: this unsafe can be trivially removed with `pin_project_lite`

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+
+        #[cfg(debug_assertions)]
+        {
+            scan.child_fields = children;
+        }
+        scan
+    }
+}
+
+/// Records the field names one child writes, and how many times, instead of
```
rcoh: it seems like there is a lot of complexity to capture "last write wins"

I wonder if we should store a log and play it backwards instead (e.g. this currently has the failure mode that setting a field to `None` would overwrite a previous value)

A little surprising if not an outright bug

metrique-util/src/metrics_pool.rs
```diff
@@ -0,0 +1,1228 @@
...
+        MetricsPoolScope {
+            pool: self.clone(),
+            future,
+        }
+    }
+
+    /// Close `metric` and append it to the pool.
+    pub fn append<E>(&self, metric: E)
+    where
+        E: CloseEntry + Send + Sync + 'static,
```
rcoh: E doesn't need to be Send+Sync+'static, only `E::Closed` I think you can drop these bounds

rcoh (commented):
the collision behavior just feels a little overengineered and confusing. I wonder if we can find a default behavior that is "pretty much right"

metrique-util/src/metrics_pool.rs
jasgin: Part of it is because we need two separate collision policies, at the intra- and inter-child levels, since we're doing last-child wins to avoid frankenentries for the latter.

> play it backwards instead

Yeah that's what it currently does at the inter-child handling, but using the field names so there's no notion of None. We could change the algorithm to match in the intra-child handling, but the functional change I think would be to not store it as seen if the value None.

metrique-util/src/metrics_pool.rs
jasgin: Are you suggesting add an upper bound to the input or just commenting on why this change was put in?

metrique-util/src/metrics_pool.rs
jasgin: Yeah maybe a CollisionPolicy API with an Index mode as a follow-up?

metrique-util/src/metrics_pool.rs
jasgin: Thinking about this more, the intra-child handling is probably an over-index since that handling isn't really warranted here. I think that simplifies things a lot.

PR #398, opened 2026-09-21
refactor(writer-core): share JSON scalar encoder

popematt:
📬 *Issue #, if available:*

#355 

✍️ *Description of changes:*

Moves the JSON number/string encoders into a new `metrique-writer-core::json_encode` module and point `format-json` at them, so the native renderer and the upcoming object string-fallback share one implementation instead of maintaining divergent copies.


metrique-writer-core/src/json_encode.rs
```diff
@@ -0,0 +1,230 @@
...
+//!
+//! The shared implementation is not a promise of byte-identical output across every path. Two
+//! deliberate carve-outs (see the object-values plan, pt001):
+//! - A [`Observation::Repeated`]'s `count` depends on the `multiplicity` passed by the caller.
+//!   The native renderer passes the entry's sampling multiplicity; the object string-fallback
+//!   has none and passes `None`, so a sampled observation can differ in `count`.
+//! - The pre-existing scalar-array `StringCapture` path in [`crate::value`] is a separate regime
+//!   and is unchanged; it does not route through these functions.
+//!
+//! `#[doc(hidden)]`: this module is hidden from rustdoc, but its items are still part of
```
rcoh: Let's avoid making this doc(hidden)

It's a public API in practice and making it hidden increases the chances of accidentally breaking

metrique-writer-core/src/json_encode.rs
```diff
@@ -0,0 +1,230 @@
...
+pub fn push_observation_comma(buf: &mut String, obs: Observation, multiplicity: Option<u64>) {
+    buf.push(',');
+    push_observation(buf, obs, multiplicity);
+}
+
+/// Push a scalar observation value into the buffer.
+///
+/// `multiplicity` scales a [`Observation::Repeated`]'s occurrence count for sampling; pass `None`
+/// (equivalent to `1`) outside a sampling context, such as the object string fallback.
+pub fn push_observation(buf: &mut String, obs: Observation, multiplicity: Option<u64>) {
```
rcoh: given that we are making these public APIs, it seems a little odd to have two APIs when the previous API just adds a comma to the beginning of it..

rcoh: might be worth changing these to an API that is "generic binary pushable thing" so if we wanted to switch to Vec<u8> in the future we could

metrique-writer-core/src/json_encode.rs
```diff
@@ -0,0 +1,230 @@
...
+            buf.push_str(",\"count\":");
+            buf.push_str(itoa::Buffer::new().format(occurrences.saturating_mul(mult)));
+            buf.push('}');
+        }
+    }
+}
+
+/// Push a float value, clamping infinities to ±`f64::MAX` and writing `null` for NaN.
+pub(crate) fn push_float(buf: &mut String, v: f64) {
+    let v = v.clamp(-f64::MAX, f64::MAX);
```
rcoh: TIL `-f64::MAX == f64::MIN`. I'd still probably prefer f64::MIN, I think that's clearer and more obviously doesn't overflow

metrique-writer-core/src/json_encode.rs
```diff
@@ -0,0 +1,230 @@
...
+        // (e.g. `f64::MAX` renders as `1.79...e308`).
+        let mut buffer = dtoa::Buffer::new();
+        let s = buffer.format_finite(v);
+        // Strip trailing ".0" for cleaner integer-like output
+        buf.push_str(s.strip_suffix(".0").unwrap_or(s));
+    }
+}
+
+/// Push a JSON-escaped string with surrounding quotes into the buffer.
+pub fn push_json_string(buf: &mut String, s: &str) {
```
rcoh: do we already have a prop test that validates that this function always produces a valid JSON string for all inputs?

metrique-writer-format-json/Cargo.toml
```diff
@@ -10,7 +10,6 @@ readme = "README.md"
 
 [dependencies]
 itoa = { workspace = true }
-dtoa = { workspace = true }
 rand = { workspace = true }
 metrique-writer-core = { workspace = true, features = ["serde"] }
```
rcoh: we need to tighten the bound on t he workspace versions of metrique-writer-core so that future versions of the json format mandate a new enough core

